PR #2, opened 2016-07-27
Checks if user is Otto the bot

mikeger:
- This is needed for tracking
- Includes tests


Source/Model/User/ZMUser.m
```diff
@@ -73,6 +73,7 @@
 static NSString *const ShowingUserRemovedKey = @"showingUserRemoved";
 static NSString *const UserClientsKey = @"clients";
 
+static NSString *const OttoPermanentRemoteIdentifier = @"e1832f71-8cdf-45c3-911d-018926f24c9c";
```
typfel: Where did you get this from? Does it always stay the same?

mikeger: Yes, Otto is having a permanent user id

mikeger:
@typfel works in real-life, can be merged

PR #3, opened 2016-08-01
Better way of detecting bots (Anna and Otto)

mikeger:
- It came out that the user id for Otto would still have to change, so it was decided to go on with email detection


marcoconti83:
Looks good! Please release

marcoconti83:
... once Travis is passed :)

mikeger:
@marcoconti83 passed!

mikeger:
... and failed, but failed looks like the fail in the middle of testing

PR #10, opened 2016-08-09
Delete messages for everyone

daehn:
**What's in this PR?**

* Adds the logic to delete a message for all participants of a conversation

‼️ Depends on new releases after the following PR's have been merged: 
* https://github.com/wireapp/generic-message-proto/pull/5  ✅
* https://github.com/wireapp/backend-api-protobuf/pull/2  ✅
* https://github.com/wireapp/wire-ios-protos/pull/2 ✅


mikeger:
@wire-bot @daehn something wrong with bot

daehn:
Nope, it's actually right, the copyright header is missing, see https://github.com/wireapp/wire-ios-data-model/blob/develop/Source/Model/Conversation/ZMConversation.m#L6

mikeger:
😱

marcoconti83:
@daehn protos merged and released as 6.1.0

daehn:
@marcoconti83 thanks! I bumped the dependencies.

Source/Model/Message/ZMMessage+Internal.h
```diff
@@ -70,7 +70,8 @@ extern NSString * const ZMMessageHiddenInConversationKey;
 - (BOOL)shouldGenerateUnreadCount;
 
 - (void)removeMessage;
-+ (void)removeMessageWithRemotelyDeletedMessage:(ZMMsgDeleted *)deletedMessage fromUser:(ZMUser *)user inManagedObjectContext:(NSManagedObjectContext *)moc;
++ (void)removeMessageWithRemotelyHiddenMessage:(ZMMessageHide *)hiddenMessage fromUser:(ZMUser *)user inManagedObjectContext:(NSManagedObjectContext *)moc;
```
marcoconti83: For someone that is not initiated to the terminology,  the difference between`Hidden` and `Deleted` is not clear. We should add method comments.

Source/Model/Conversation/ZMConversation+Internal.h
```diff
@@ -201,6 +202,7 @@ extern NSString *const ZMConversationLastReadLocalTimestampKey;
 - (ZMClientMessage *)appendOTRSessionResetMessage;
 - (ZMClientMessage *)appendOTRMessageWithText:(NSString *)text nonce:(NSUUID *)nonce;
 - (ZMClientMessage *)appendOTRMessageWithLocationData:(ZMLocationData *)locationData nonce:(NSUUID *)nonce;
+- (ZMClientMessage *)appendNonExpiringGenericMessage:(ZMGenericMessage *)genericMessage hidden:(BOOL)hidden;
```
marcoconti83: The name of the method is clear, but what does the `hidden` argument mean? We should add a comment to the method.

daehn:
I noticed in the spec that we shouldn't insert a system message on the device where the deletion is happening, change incoming.

Source/Model/Conversation/ZMConversation+OTR.h
```diff
@@ -52,6 +52,8 @@
 /// Creates the message that warns user about the fact that decryption of incoming message is failed
 - (void)appendDecryptionFailedSystemMessageAtTime:(NSDate *)timestamp sender:(ZMUser *)sender client:(UserClient *)client errorCode:(NSInteger)errorCode;
 
+- (void)appendNewDeletedMessageForEveryoneSystemMessageWithTimestamp:(NSDate *)timestamp sender:(ZMUser *)sender;
```
mikeger: Name of the method for me sound a little bit too long, I think better name could be `appendDeletedForEveryoneSystemMessageWithTimestamp `

- `new` can be skipped, since any appended message is new
- `message...message` produces a tautology

daehn:
@mikeger @marcoconti83 please re-review

Source/Model/Conversation/ZMConversation.m
```diff
@@ -1477,6 +1487,24 @@ - (void)startFetchingMessages
...
+- (ZMClientMessage *)appendNonExpiringGenericMessage:(ZMGenericMessage *)genericMessage hidden:(BOOL)hidden
+{
+    VerifyReturnNil(genericMessage != nil);
+    
+    ZMClientMessage *message = [ZMClientMessage insertNewObjectInManagedObjectContext:self.managedObjectContext];
+    [message addData:genericMessage.data];
+    message.sender = [ZMUser selfUserInContext:self.managedObjectContext];
+    message.isEncrypted = YES;
+    
+    if(!hidden) {
```
mikeger: I would reverse the logic, e.g. `if(hidden)` otherwise for me looks confusing

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -0,0 +1,67 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+
+extension ZMConversation {
+    static func appendSelfConversationWithDeleteOfMessage(message: ZMMessage) {
```
mikeger: For me looks like a comment is missing about what the method does. Also the name sounds against the English language to me, seems it should be `appendToSelfConversation...`

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -0,0 +1,67 @@
...
+        let genericMessage = ZMGenericMessage(hideMessage: messageNonce.transportString(), inConversation: convID.transportString(), nonce: nonce.transportString())
+        ZMConversation.appendSelfConversationWithGenericMessageData(genericMessage.data().copy() as! NSData, managedObjectContext: message.managedObjectContext)
+    }
+}
+
+extension ZMMessage {
+    
+    // NOTE:- This is a free function meant to be call from Objc because you can't call protocol extension from it
+    public static func hideMessage(message: ZMConversationMessage) {
+        guard let castedMessage = message as? ZMMessage else { return }
```
mikeger: Should it assert if the condition is not met?

Source/Model/Message/ConversationMessage+Deletion.swift
daehn: Good point. Btw this is old code that was copied over from sync-engine as it belongs more to the data model.

daehn: Also the name is wrong, should be Hide in this case now

daehn:
@mikeger updated

PR #11, opened 2016-08-11
Add message to hiddenInConversation relation instead of deleting it

daehn:
**What's in this PR?**

There was a possible race condition in which a message has been deleted when a second part (e.g. link preview or medium asset) arrives.

In that case the message would have been re-added to the conversation if we deleted it in the managed object context. To avoid this issue we now add the message to the `hiddenInConversation` relation on it's conversation and remove it from it's `visibleInConversation` relation.


mikeger:
@daehn the relation itself was added before? There was a migration? Just checking ✔️

daehn:
Yes, the relation was there already. It was used to hide duplicate messages during the e2ee tradition phase.

PR #12, opened 2016-08-12
add support for editing messages

basine:
### What's in this pull request

**For sending edits:**
1) When we edit a message, we set the original message to hidden and insert a new MessageEdit message with the new content but the same serverTimeStamp and index
2 a) If the edit succeeds we delete the data of the original message and make sure to not update the replacing message servertimestamp. Instead we save the servertimestamp of the editing message as a new property - the editing date.
2 b) If the edit fails, we keep the edited test and overwrite the nonce of the editMessage to the original message. We also delete the original message. If we resend, instead of resending, we insert a new editMessage with a new nonce and a ref to the original message nonce.

**For receiving edits:**
When we receive an edit, we delete the content of the original message, but keep the "empty shell" set to hidden
We insert a new message with the same serverTimestamp
If we receive a multipart message part of a message that has a hidden part, we discard the content. This way we avoid race conditions in which a part of a multipart message might be reinserted after editing / deleting the first part.

Ready to merge:

- [x] Needs migration test
- [x] Waiting for new release 6.2.0 of https://github.com/wireapp/wire-ios-protos


marcoconti83:
Wire-ios-protos is merged, please make a new release (you can now use Jenkins to create a new release)

Source/Model/Message/ZMMessage.m
```diff
@@ -374,34 +376,56 @@ + (void)removeMessageWithRemotelyDeletedMessage:(ZMMessageDelete *)deletedMessag
...
+    switch (eventType) {
+            
+        case ZMUpdateEventConversationMessageAdd:
+        case ZMUpdateEventConversationKnock:
+            return [[payload dictionaryForKey:@"data"] uuidForKey:@"nonce"];
+
+        case ZMUpdateEventConversationClientMessageAdd:
+        case ZMUpdateEventConversationOtrMessageAdd:
+        {
+            //if event is otr message than payload should be already decrypted and should contain generic message data
```
marcoconti83: `than` -> `then`?

Source/Model/Message/ZMMessage.m
```diff
@@ -415,27 +439,24 @@ - (void)updateWithPostPayload:(NSDictionary *)payload updatedKeys:(__unused NSSe
         self.serverTimestamp = timestamp;
         updatedTimestamp = YES;
     }
-    [self.conversation updateLastReadServerTimeStampIfNeededWithTimeStamp:timestamp andSync:YES];
+    [self.conversation updateLastReadServerTimeStampIfNeededWithTimeStamp:timestamp andSync:NO];
```
marcoconti83: What's the meaning of `andSync:` here? Why did it change?

marcoconti83:
Looks good! Can we remove the "Do not merge"?

Source/Model/Message/ZMMessage.m
basine: With previous changes, we should not need to send lastRead messages for messages we posted ourselves. Instead other clients of the selfUser should update the lastRead automatically when the sender is the selfUser. I think we forgot to update the Bool here.

Source/Model/Message/ZMMessage.m
basine: Yeah I didn't write this comment, just moved it ;) I can change it

Source/Model/Message/ZMClientMessage.m
```diff
@@ -97,7 +109,9 @@ - (void)addData:(NSData *)data
         self.nonce = [NSUUID uuidWithTransportString:messageData.genericMessage.messageId];
     }
     
-    [self setLocallyModifiedKeys:[NSSet setWithObject:ClientMessageDataSetKey]];
+    if (self.managedObjectContext.zm_isUserInterfaceContext) {
```
basine: This check was previously missing. I guess we don't want to set the modified keys on the syncMOC, do we? @marcoconti83

Source/Model/Message/ZMClientMessage.m
```diff
@@ -455,9 +508,17 @@ @implementation ZMClientMessage (ZMTextMessageData)
...
 {
+    if (self.genericMessage.hasEdited) {
+        return self.genericMessage.edited.text.content;
+    }
     return self.genericMessage.text.content;
 }
 
+- (BOOL)isEdited
+{
+    return self.genericMessage.hasEdited;
```
mikeger: `hasEdited` sounds not correct to my english understanding, would it be may be better to call it `hasEdit` or `wasEdited` ?

mikeger: Ah, I guess it is the part of protobuf auto name generation?

Source/Model/Message/ZMClientMessage.m
marcoconti83: We sometimes set locally modified key on the sync context, to force something to upload as a result of some event that happened on the sync context . E.g. when image processing from the image processing queue is completed, it sets the data on the sync context, and we want to upload that to the BE.
Not sure of what are the implications here, but this change seems risky. Did you have problems that required this check to be added? I guess we are receiving a new message A that tells us to change the content of previously existing message B, and if we set this flag, we will try to upload message B again? This could probably be prevented with a predicate on the transcoder/request strategy.

Source/Model/Message/ZMMessage.m
marcoconti83: Got it

Source/Model/Message/ZMClientMessage.m
basine: Exactly, that's an automatically created flag.

Source/Model/Message/ZMClientMessage.m
basine: Reverting.

PR #16, opened 2016-08-25
Adopt to new cryptobox API

typfel:
Cryptobox has been re-written in swift and the API has changed to be thread safe. This PR adopts the new API.


Cartfile.resolved
```diff
@@ -1,13 +1,13 @@
 github "wireapp/PINCache" "2.2.2"
 github "wireapp/libPhoneNumber-iOS" "0.8.12"
-github "wireapp/ocmock" "v3.2.2"
+github "wireapp/ocmock" "v2.2.4-wire"
```
marcoconti83: This should stay at 3.2

typfel: OK I'll revert

PR #19, opened 2016-08-29
Add confirmation message 

basine:
**In this PR:**

When receiving a message that was not sent by the selfUser, we create and send a `Confirmation` message of type .Delivered which contains the messageID to be confirmed.

When we receive a `Confirmation` message of type .Delivered that references a message that the selfUser sent, we create a `ZMMessageConfirmation` object that holds a reference to the message and the user who sent the confirmation. This object has a type which can be updated when the message is read (Confirmatiom messages of type .Read are not implemented in this PR).

Not implemented yet (there is code that has been commented out): The BE needs to support sending messages only to a subset of users in a conversation. Currently we receive a 412 when we try to send the messages only to the sender's clients.


Source/Model/Message/ZMClientMessage.m
```diff
@@ -337,6 +341,39 @@ - (NSData *)encryptedMessagePayloadData
...
 
++ (ZMNewOtrMessage *)otrMessageForGenericMessage:(ZMGenericMessage *)genericMessage
+                                      selfClient:(UserClient *)selfClient
+                                 conversation:(ZMConversation *)conversation
+                                 externalData:(NSData *)externalData
+                               sessionsDirectory:(EncryptionSessionsDirectory *)sessionsDirectory
+{
+    NSArray <ZMUserEntry *>*recipients;
+    
+//    if (genericMessage.hasConfirmation) {
```
marcoconti83: Commented out code

Source/Model/Message/ZMMessage.m
```diff
@@ -199,11 +201,17 @@ - (void)removeExpirationDate;
...
 
-- (void)markAsDelivered
+- (void)markAsSent
 {
     self.isExpired = NO;
 }
 
+- (void)confirmReception
+{
+    ZMGenericMessage *genericMessage = [ZMGenericMessage messageWithConfirmation:self.nonce.transportString type:ZMConfirmationTypeDELIVERED nonce:[NSUUID UUID].transportString];
```
marcoconti83: Just as a thought, what happens if the user of this class by mistake calls `confirmReception` twice on the same message?

Source/Model/Message/ZMClientMessage.m
basine: As mentioned in the description above I am still waiting for the BE to implement this.

Source/Model/Message/ZMMessage.m
basine: We send out two messages - that shouldn't break anything.

Source/Model/Message/ZMClientMessage.m
marcoconti83: I missed it. Thanks for clarifying.

PR #23, opened 2016-08-30
Do not clear the sender when removing a message caused by an edit

daehn:
### [7166](https://wearezeta.atlassian.net/browse/ZIOS-7166): Remotely edited message removes the sender avatar and name
___

* We previously removed the sender remote identifier from a message when calling `[ZMMessage removeMessage]` on it, but in case of an edit we still need this information to update the new message with the correct sender identifier. This lead to an edited message not displaying the sender name / image.
* The `[ZMMessage removeMessage]` has been renamed to `[ZMMessage removeMessageClearingSender:]` and the parameter indicates weather the sender information should be removed.
* This also fixes some flaky test for asset client messages.

### DIFF
--- a/Source/Model/Message/ZMMessage+Internal.h
+++ b/Source/Model/Message/ZMMessage+Internal.h
@@ -74,7 +74,9 @@ extern NSString * const ZMMessageConfirmationKey;
 - (void)resend;
 - (BOOL)shouldGenerateUnreadCount;
 
-- (void)removeMessage;
+/// Removes the message and deletes associated content
+/// @param clearingSender Wether information about the sender should be removed or not
+- (void)removeMessageClearingSender:(BOOL)clearingSender;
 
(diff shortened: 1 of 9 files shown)


Source/Model/Message/ZMMessage+Internal.h
```diff
@@ -74,7 +74,9 @@ extern NSString * const ZMMessageConfirmationKey;
 - (void)resend;
 - (BOOL)shouldGenerateUnreadCount;
 
-- (void)removeMessage;
+/// Removes the message and deletes associated content
+/// @param clearingSender Wether information about the sender should be removed or not
```
typfel: typo `Wether` > `Wheater`

daehn: I think it's `whether`

daehn:
@typfel please re-review

PR #26, opened 2016-08-31
Reaction observation

Einh06:
**What's in this PR**
In this PR was implemented observation for reactions on message.
There is a subtlety in that every `MessageObserverToken` has to observe every Reaction, because we want to notify when list of user changes for a given reaction.


Source/Model/Message/ConversationMessage+Reaction.swift
```diff
@@ -32,6 +32,14 @@ extension ZMMessage {
         message.addReaction(unicodeValue, forUser: ZMUser.selfUserInContext(context))
     }
     
+    public static func removeReaction(onMessage message:ZMConversationMessage)
```
marcoconti83: Should we check if the reaction was not here in the first place, before removing it?

Source/Notifications/ObjectObserverTokens/MessageObserverToken.swift
```diff
@@ -138,9 +153,17 @@ public final class MessageObserverToken: ObjectObserverTokenContainer, ZMUserObs
         
         changeHandler = {
             [weak self] (_, changeInfo) in
+            if changeInfo.reactionsChanged {
+                if let strongSelf = self {
+                    strongSelf.createTokensForReactions(Array<Reaction>(strongSelf.observedMessage.reactions))
+                }
+                // add a new the new reaction observer
```
marcoconti83: is this a TODO?

Source/Model/Message/ConversationMessage+Reaction.swift
Einh06: If the user is not there, nothing would happen. We would still iterate through all reaction to find it, but i think that's fine.

Source/Notifications/ObjectObserverTokens/MessageObserverToken.swift
Einh06: Yes, this is out of date, removing.

PR #27, opened 2016-08-31
Move the database into a shared container

daehn:
# What's in this PR?

We need access to core data in iOS extensions, which has the requirement of the database being located in a shared container. This PR adds the logic to move the database in to a new location.

# Changes

* The `NSManagedObjectContext` creation methods, as well as the methods used to perform a migration (or movement of the database) now take a `NSURL` as additional parameter which specifies the location where the database persistent store should be created (or opened if already present).
* The persistent store will be created in a subdirectory using the bundle identifier as name.
* Extracted common logic for tests using a store that is not in memory into a common super class.

# Things to consider

~~This now uses the `-migratePersistentStore:toURL:options:withType:` method on `NSPersistentStoreCoordinator` to move the database into the new location (we keep the store type), when we previously moved the database from the caches to the application support directory we did this manually. This API takes a persistent store, which means we need to open, and potentially migrate the store in the old location (application support) and then pass it to this method.  The documentation states the following:~~

~~Used for save as - performance may vary depending on the type of old and new store; the old store is usually removed from the coordinator by the migration operation, and therefore is no longer a useful reference after invoking this method~~

**Updated:** 
The movement is now done using the same way as before when moving the store from the caches to the application support directory.

The system API does not move the external data located in `.{database-name}_SUPPORT` and the store to be moved needs to be opened (and thus migrated before moving it), the documentation also states both stores will be open-end and all objects copied over (which makes sense as this method is also used to change the store's type), which might lead to more memory being used than necessary in our case. As we only want to move the store once on startup we now fall back to the way we moved the store before and also move the auxiliary data blob files.

___

We perform this migration if the a store is present in the applications support directory, which is why we manually check and remove the store after the migration is done to ensure it is removed.

**Please pay extra attention if there are more possible test cases that can be added.**


Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.h
```diff
@@ -36,26 +36,30 @@
...
 
 /// Creates persistent store coordinator and migrates store if needed
 /// @param sync defines if the method should execute sycnhronously or not (ususally it makes sence to execute it
 ///         synchronously when @c +needsToPrepareLocalStore is YES)
 /// @param backupCorruptedDatabase: if true, will copy a corrupted database to another folder for later investigation
 /// @param completionHandler callback to be executed on completion (nullable)
-+ (void)prepareLocalStoreSync:(BOOL)sync backingUpCorruptedDatabase:(BOOL)backupCorrputedDatabase completionHandler:(void(^)())completionHandler;
++ (void)prepareLocalStoreSync:(BOOL)sync
+                  inDirectory:(NSURL *)directory
+   backingUpCorruptedDatabase:(BOOL)backupCorrputedDatabase
```
typfel: typo: `backupCorrputedDatabase`

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -442,66 +446,120 @@ + (NSPersistentStoreCoordinator *)onDiskPersistentStoreCoordinator
...
+                                                                       storeURL:self.applicationSupportDirectoryStoreURL
+                                                        backupCorruptedDatabase:backupCorruptedDatabase];
     
-    NSString *oldModelVersion = [sourceMetadata[NSStoreModelVersionIdentifiersKey] firstObject];
+    // perform the migration to the new store location
+    [psc performBlockAndWait:^{
+        NSError *migrationError = nil;
+        [psc migratePersistentStore:oldStore toURL:self.storeURL options:nil withType:NSSQLiteStoreType error:&migrationError];
+        if (nil != migrationError) {
+            ZMLogError(@"Unable to migrate persistent store form application directory into directory %@: %@", DatabaseDirectoryURL, migrationError);
```
typfel: typo: `form` > `from`

daehn: We also might even want to crash if this happens

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -581,17 +670,26 @@ + (NSURL *)cachesDirectoryStoreURL;
...
     NSFileManager *fm = [NSFileManager defaultManager];
-    NSURL *databaseURL = [self cachesDirectoryStoreURL];
-    BOOL fileExists = [fm fileExistsAtPath:databaseURL.path isDirectory:nil];
-    return fileExists;
+    return [fm fileExistsAtPath:url.path isDirectory:nil];
 }
 
 /// Checks if database is created, but it is still locked with iOS file protection
-+ (BOOL)databaseExistsAndNotReadableDueToEncryption
++ (BOOL)databaseExistsAndNotReadableDueToEncryption;
```
typfel: Why add the `;`?

Tests/Source/ManagedObjectContext/Migrations/DatabaseBaseTest.m
```diff
@@ -0,0 +1,159 @@
...
+{
+    return [self moveDatabaseToSearchPathDirectory:NSApplicationSupportDirectory];
+}
+
+- (NSData *)invalidData
+{
+    return [@"Trollolol" dataUsingEncoding:NSUTF8StringEncoding];
+}
+
+- (BOOL)createdUnredableLocalStore
```
typfel: typo: `Unredable` > `Unreadable`

Tests/Source/ManagedObjectContext/Migrations/DatabaseMovingTests.m
```diff
@@ -20,145 +20,138 @@
...
-    NSString *supportURL = [self.applicationSupportStoreURL.URLByDeletingLastPathComponent URLByAppendingPathComponent:@".store_SUPPORT"].path;
-    XCTAssertTrue([self.fm fileExistsAtPath:supportURL]);
+    XCTAssertFalse([NSManagedObjectContext databaseExistsInApplicationSupportDirectory]);
+    XCTAssertFalse([NSManagedObjectContext databaseExistsInCachesDirectory]);
+    XCTAssertFalse([NSManagedObjectContext needsToPrepareLocalStoreInDirectory:self.sharedContainerDirectoryURL]);
+
+    // TODO: The `.store_SUPPORT` file does not get moved to the new location somehow,
+    // either some validation happening (it's empty in the test) or we need to fallback to manually copying it
+    //
+    // NSString *supportURL = [self.sharedContainerStoreURL.URLByDeletingLastPathComponent URLByAppendingPathComponent:@".store_SUPPORT"].path;
```
typfel: So should we activate this?

Tests/Source/ManagedObjectContext/Migrations/DatabaseMovingTests.m
```diff
@@ -198,76 +191,43 @@ - (void)testThatItReportsDatabaseExistsInCachesDirectoryAfterMovingIt
     
     // then
     XCTAssertTrue([NSManagedObjectContext databaseExistsInCachesDirectory]);
-}
+    XCTAssertFalse([NSManagedObjectContext databaseExistsInApplicationSupportDirectory]);
```
typfel: The name of this test is not clear to me. What are we testing here? That `needsToPrepareLocalStoreInDirectory` returns `YES` when the store is located in the caches directory?

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
daehn: Can be removed, I guess I changed it and then copied it over from the header again.

Tests/Source/ManagedObjectContext/Migrations/DatabaseMovingTests.m
daehn: I added this, I still need to verify moving external data works as expected

daehn: I am working on this now and it looks like this test was flawed, it created some files but not a valid SQLite database and of course it breaks when I copy a database.

daehn:
@typfel please re-review

marcoconti83:
Conflict

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -605,23 +675,19 @@ + (BOOL)databaseExistsAndNotReadableDueToEncryption
     return result;
 }
 
-+ (BOOL)moveDatabaseFromCachesToApplicationSupportIfNeeded
++ (BOOL)moveDatabaseFromURL:(NSURL *)fromURL toURL:(NSURL *)toURL
 {
-    // we need to move if true
-    if ([self databaseExistsInCachesDirectory]) {
-        
+    if ([self databaseExistsAtURL:fromURL]) {
```
marcoconti83: We should handle the case when you crash after moving the first file but you still have more files to copy. We should resume the moving at the next start until all files are copied, then delete the folder, only then we should consider the database removed from the old location.

daehn:

marcoconti83:
Looks good

PR #28, opened 2016-09-01
return object containing message and info if it needs confirmation

basine:
**In this PR**

Previously we were inserting the confirmation message directly when inserting a message from an updateEvent. From the outside it's intransparent if a confirmation message was inserted or not. In order to send confirmation messages in the background, `wire-ios-sync-engine` needs to know this. 
Therefore we are now returning the original message and the information if a confirmation message should be sent and insert the confirmation message afterwards (in the clientTranscoder of `wire-ios-sync-engine`).

Note: Without the changes on sync-engine this change will break confirmation messages
Merge https://github.com/wireapp/wire-ios-sync-engine/pull/39 after

### DIFF
--- a/Source/Model/Message/ZMOTRMessage.h
+++ b/Source/Model/Message/ZMOTRMessage.h
@@ -20,6 +20,7 @@
 #import "ZMMessage+Internal.h"
 
 @class UserClient;
+@class MessageUpdateResult;
 
 extern NSString * const DeliveredKey;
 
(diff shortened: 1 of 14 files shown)


Source/Model/Message/ZMOTRMessage.h
```diff
@@ -41,4 +42,12 @@ extern NSString * const DeliveredKey;
                                  inManagedObjectContext:(NSManagedObjectContext *)moc
                                          prefetchResult:(ZMFetchRequestBatchResult *)prefetchResult;
 
++ (MessageUpdateResult *)messageUpdateResultFromUpdateEvent:(ZMUpdateEvent *)updateEvent
+                                     inManagedObjectContext:(NSManagedObjectContext *)moc
+                                             prefetchResult:(ZMFetchRequestBatchResult *)prefetchResult;
+
++ (instancetype)createOrUpdateMessageFromUpdateEvent:(ZMUpdateEvent *)updateEvent
```
daehn: I think we could also directly remove the old method, as `wire-iso-sync-engine` will show an error an fail to build anyways.

marcoconti83:
Blocked: Waiting for @basine's work on sync engine

Source/Model/Message/ZMOTRMessage.h
basine: This is a super class method which is still widely used in our tests (until we remove all tests with ZMTextMessage and ZMImageMessage).

marcoconti83:
This can now be merged!

PR #29, opened 2016-09-01
Send delivery receipt only to sender instead of everyone in conversation

marcoconti83:
Sending a delivery confirmation to all users of a conversation is a waste, only the sender needs to receive it.

# Changes
- If the message is a confirmation message, only the clients of the sender are included in the payload
- Moved around some test definitions for clearer structure
- Swiftification
- Now returning a `MissingClientsStrategy` to clarify whether the message should be sent only to one users or to all users in the conversation

# Testing
- Added testing for this feature


marcoconti83:
BLOCKER: Waiting for testing this on staging backend, when we give green light it will be ported to production backend, then we can merge

typfel:
@marcoconti83  How long will that take? Maybe could extract the fix for tests crashing into another PR?

marcoconti83:
This PR completely changed, because of https://github.com/wireapp/wire-ios-data-model/pull/29/commits/a723180903dcfcce15732f9fcc63f313219c694f.

Please re-review.

Source/Model/Message/ZMAssetClientMessage+Encryption.swift
```diff
@@ -0,0 +1,81 @@
...
+        
+        var recipients : [ZMUserEntry] = []
+        selfClient.keysStore.encryptionContext.perform { (sessionsDirectory) in
+            recipients = genericMessage.recipientsWithEncryptedData(selfClient, recipients: conversation.activeParticipants.array as! [ZMUser], sessionDirectory: sessionsDirectory)
+        }
+        
+        builder.setRecipientsArray(recipients)
+        return builder.build()
+    }
+}
```
daehn: Missing newline

typfel:
So what changed? You also ported some code into swift...

marcoconti83:
@typfel now the methods returning the encrypted data return a tuple, not just data, which specify what's the policy to adopt for missing clients.

Source/Model/Message/ZMAssetClientMessage+OTR.swift
```diff
@@ -0,0 +1,9 @@
+//
+//  ZMAssetClientMessage+OTR.swift
+//  ZMCDataModel
+//
+//  Created by Marco Conti on 02/09/16.
```
daehn: Missing copyright header

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -0,0 +1,172 @@
...
+/// us that some user clients that were supposed to be there are missing (e.g.
+/// another user added a new client that we don't yet know about). The various
+/// strategies give a hint to the backend of how we want to handle missing clients.
+public enum MissingClientsStrategy {
+    
+    /// Fail the request if there is any missing client
+    case DoNotIgnoreAnyMissingClient
+    /// Fail the request if there is any missing client for the given user,
+    /// but ignore missing clients of any other user
+    case IgnoreAllMissingClientsNotFromUser(user: ZMUser)
```
daehn: 👍

PR #34, opened 2016-09-02
handle unarchiving when creating messages

basine:
**In this PR**

We were previously handling unarchiving in the transcoder by inspecting the event type and timestamps. For messages we should not do this, because some message events don't create a message, but only invisible side effects.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Transport.m
+++ b/Source/Model/Conversation/ZMConversation+Transport.m
@@ -240,7 +240,7 @@ NSString *const ZMConversationInfoOTRArchivedReferenceKey = @"otr_archived_ref";
 {
     if ([event canUnarchiveConversation:self]){
         self.internalIsArchived = NO;
-        [self updateArchivedChangedTimeStampIfNeeded:event.timeStamp andSync:YES];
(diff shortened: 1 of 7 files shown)


Source/Model/Conversation/ZMConversation+Transport.m
```diff
@@ -240,7 +240,7 @@ - (void)unarchiveConversationFromEvent:(ZMUpdateEvent *)event;
 {
     if ([event canUnarchiveConversation:self]){
         self.internalIsArchived = NO;
-        [self updateArchivedChangedTimeStampIfNeeded:event.timeStamp andSync:YES];
+        [self updateArchivedChangedTimeStampIfNeeded:event.timeStamp andSync:NO];
```
marcoconti83: Why the switch to NO? Does it mean no event will ever sync unarchiving the conversation?

marcoconti83:
👍

Source/Model/Conversation/ZMConversation+Transport.m
basine: All other clients should receive the same event and unarchive the conversation. It does not need to be synced between clients then.

marcoconti83: So previously we were syncing something that we should have not?

PR #35, opened 2016-09-05
Clear all reactions when a message is removed / edited

daehn:
# What's in this PR?

* If a message is edited the reactions should be reset.
* We now clear the reactions when calling `removeMessageClearingSender:` to ensure this data is removed in all cases when a message is removed (which is also the case for edits as a new message is inserted)

### DIFF
--- a/Source/Model/Message/ConversationMessage+Reaction.swift
+++ b/Source/Model/Message/ConversationMessage+Reaction.swift
@@ -66,6 +66,8 @@ extension ZMMessage {
 
     @objc public func clearAllReactions() {
         reactions.removeAll()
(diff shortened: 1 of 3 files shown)


typfel:
Is to remove it from db? Because isn't edited messages actually a new message anyway?

marcoconti83:
Maybe @typfel has a point? But it's still good to clear the DB of unused entries.

daehn:
I was wondering how it could have happened that the likes were transferred to the new message as well, as it has a new none and we do not copy them over manually... Nevertheless this definitely removes them now, I added a call to `deleteObject` to actually delete them from the db in 8b1a8bb.

PR #38, opened 2016-09-05
Fix observing

Einh06:
**What's in this PR**
The `Reaction` observer type had an index superior than the `MessageWindowObserver`, which lead the former being processed after the later. This means that the UI would be notifies of a change from the second event onward when receiving event. By reordering the reaction observer type index, we now are correctly notified when something is changing.

### DIFF
--- a/Source/Notifications/ObjectObserverTokens/ManagedObjectContextObserver.swift
+++ b/Source/Notifications/ObjectObserverTokens/ManagedObjectContextObserver.swift
@@ -26,6 +26,9 @@ import ZMCSystem
 private let zmLog = ZMSLog(tag: "Observer")
 
 public enum ObjectObserverType: Int {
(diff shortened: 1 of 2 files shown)


Source/Notifications/ObjectObserverTokens/ManagedObjectContextObserver.swift
```diff
@@ -26,6 +26,9 @@ import ZMCSystem
 private let zmLog = ZMSLog(tag: "Observer")
 
 public enum ObjectObserverType: Int {
+    
+    // The order of this enum is important, because some event create additional observer to fire (reaction fires message notification), 
+    // and therefore needs to happen before the message observer handler to propagate properly to the UI
```
daehn: Should be `message window`

PR #39, opened 2016-09-05
never expire confirmation messages

basine:
**In this PR**

We should not expire confirmation messages because they can't be resent and would create an unsent message list indicator upon fail.


daehn:
Should they also be excluded from the unsent indicator calculation?

marcoconti83:
👍  
Merged

PR #40, opened 2016-09-06
Validate emoji before inserting

Einh06:
**What's in this PR**
If this PR is merged, we had the ability of validating the incoming emoji, so that if we receive an emoji that we are not expecting, nothing is inserted.
So far, the only emoji we take into account is ❤️

**NOTE**: This PR breaks the UI facing API.

### DIFF
--- a/Tests/Source/Model/Messages/ZMClientMessagesTests+Reaction.swift
+++ b/Tests/Source/Model/Messages/ZMClientMessagesTests+Reaction.swift
@@ -1,0 +1,75 @@
+//
+//  ZMClientMessagesTests+Reaction.swift
+//  ZMCDataModel
+//
+//  Created by Florian Morel on 9/6/16.
+//  Copyright © 2016 Wire Swiss GmbH. All rights reserved.
+//
+
+import ZMTesting
+
+class ZMClientMessageTests_Reaction: BaseZMClientMessageTests {
+    
+    
+}
+
+extension ZMClientMessageTests_Reaction {
+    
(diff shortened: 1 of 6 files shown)


Tests/Source/Model/Messages/ZMClientMessagesTests+Reaction.swift
```diff
@@ -0,0 +1,75 @@
...
+    func testThatItDoesNOTAppendsAReactionWhenReceivingUpdateEventWithValidReaction() {
+        
+        let conversation = ZMConversation.insertNewObjectInManagedObjectContext(uiMOC)
+        conversation.remoteIdentifier = .createUUID()
+        conversation.conversationType = .OneOnOne
+        
+        let sender = ZMUser.insertNewObjectInManagedObjectContext(uiMOC)
+        sender.remoteIdentifier = .createUUID()
+        
+        let message = conversation.appendMessageWithText("JCVD, full split please") as! ZMMessage
```
marcoconti83: 😆

Tests/Source/Model/Messages/ZMClientMessagesTests+Reaction.swift
```diff
@@ -0,0 +1,75 @@
+//
+//  ZMClientMessagesTests+Reaction.swift
+//  ZMCDataModel
+//
+//  Created by Florian Morel on 9/6/16.
+//  Copyright © 2016 Wire Swiss GmbH. All rights reserved.
```
daehn: Missing Copyright headers

PR #43, opened 2016-09-08
Allow storing database in application support for travis / external users

mikeger:
Relates to https://github.com/wireapp/wire-ios-sync-engine/pull/51

# Issue

On Travis there is no signing identity, and Xcode forbids using the shared container for our app during the tests, this makes tests crash. This is also the issue for external people who have no signing identity.

# Solution

Solution is to allow storing database in Application Data for simulator and builds that are non enterprise and non-appstore.

# Addition
I added the CoreData concurrency debugging flag and set it to be ON in tests.

### DIFF
--- a/Source/Model/User/ZMSearchUser.m
+++ b/Source/Model/User/ZMSearchUser.m
@@ -336,7 +336,14 @@ NSString *const ZMSearchUserTotalMutualFriendsKey = @"total_mutual_friends";
             user.needsToBeUpdatedFromBackend = YES;
             ZMConnection *connection = [ZMConnection insertNewSentConnectionToUser:user];
             connection.message = text;
-            self.user = user;
+            
+            [self.syncMOC saveOrRollback];
+            
+            [self.uiMOC performGroupedBlockAndWait:^{
+                NSError *uiObjectError = nil;
+                self.user = [self.uiMOC existingObjectWithID:user.objectID error:&uiObjectError];
+                Require(uiObjectError == nil);
+            }];
             
             // Do a delayed save and run the handler on the main queue, once it's done:
             ZMSDispatchGroup * g = [ZMSDispatchGroup groupWithLabel:@"ZMSearchUser"];
--- a/Source/Model/User/ZMSearchUser.m
+++ b/Source/Model/User/ZMSearchUser.m
@@ -339,20 +339,20 @@ NSString *const ZMSearchUserTotalMutualFriendsKey = @"total_mutual_friends";
             
             [self.syncMOC saveOrRollback];
             
-            [self.uiMOC performGroupedBlockAndWait:^{
-                NSError *uiObjectError = nil;
-                self.user = [self.uiMOC existingObjectWithID:user.objectID error:&uiObjectError];
-                Require(uiObjectError == nil);
-            }];
-            
             // Do a delayed save and run the handler on the main queue, once it's done:
             ZMSDispatchGroup * g = [ZMSDispatchGroup groupWithLabel:@"ZMSearchUser"];
             [self.syncMOC enqueueDelayedSaveWithGroup:g];
             
+            [self.uiMOC.dispatchGroup enter];
             ZM_WEAK(self);
             [g notifyOnQueue:dispatch_get_main_queue() block:^{
+                NSError *uiObjectError = nil;
+                self.user = [self.uiMOC existingObjectWithID:user.objectID error:&uiObjectError];
+                Require(uiObjectError == nil);
+                
                 ZM_STRONG(self);
                 completionHandler();
+                [self.uiMOC.dispatchGroup leave];
             }];
         }];
     }
--- a/Tests/Source/ManagedObjectContext/Migrations/DatabaseBaseTest.m
+++ b/Tests/Source/ManagedObjectContext/Migrations/DatabaseBaseTest.m
@@ -165,4 +165,14 @@
     }];
 }
 
+- (void)useApplicationSupportDirectoryAsDefault
+{
+    NSFileManager *fm = [NSFileManager defaultManager];
+    NSError *error = nil;
+    NSURL * const directory = [fm URLForDirectory:NSApplicationSupportDirectory inDomain:NSUserDomainMask appropriateForURL:nil create:YES error:&error];
+    Require(error == nil);
+    self.sharedContainerDirectoryURL = directory;
+    [NSManagedObjectContext setDatabaseDirectoryURL:self.sharedContainerDirectoryURL];
+}
+
 @end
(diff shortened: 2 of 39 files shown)


mikeger:
While it was not merged I found an issue, please stand by

Source/Model/User/ZMSearchUser.m
```diff
@@ -336,7 +336,14 @@ - (void)connectWithMessageText:(NSString *)text completionHandler:(dispatch_bloc
             user.needsToBeUpdatedFromBackend = YES;
             ZMConnection *connection = [ZMConnection insertNewSentConnectionToUser:user];
             connection.message = text;
-            self.user = user;
+            
+            [self.syncMOC saveOrRollback];
+            
+            [self.uiMOC performGroupedBlockAndWait:^{
```
marcoconti83: Why this `performGroupedBlockAndWait`? We should never do this between UI and Sync threads (or vice versa). Can this be rewritten without the wait?

marcoconti83: Maybe we can drop this in the "notifyOnQueue" below?

marcoconti83: Actually that `notifyOnQueue` is peformed without waiting on any group so it will escape test teardown :-( We should enter the UI group before the `notify` and leave it inside the block.

mikeger: Will do. This is actually one of the bugs I found with thread debugger: the self.user was app-facing outlet created on sync context.

Tests/Source/ManagedObjectContext/Migrations/DatabaseBaseTest.m
```diff
@@ -42,6 +42,20 @@ - (void)setUp
 {
     [super setUp];
     [self cleanUp];
+    
+    NSError *error = nil;
+    for (NSString *path in [self.fm contentsOfDirectoryAtPath:self.sharedContainerDirectoryURL.path error:&error]) {
```
typfel: Couldn't this be moved into the `cleanUp` method where we clean the other folders?

Tests/Source/ManagedObjectContext/Migrations/DatabaseMovingTests.m
```diff
@@ -207,7 +246,7 @@ - (void)testThatItWipesTheLocalStoreWhenItIsUnreadable
     // given
     [self performIgnoringZMLogError:^{
         XCTAssertTrue([self createdUnreadableLocalStore]);
-        XCTAssertTrue([NSManagedObjectContext needsToPrepareLocalStoreInDirectory:self.sharedContainerDirectoryURL]);
+        XCTAssertFalse([NSManagedObjectContext needsToPrepareLocalStoreInDirectory:self.sharedContainerDirectoryURL]);
```
typfel: Why is this now returning `false`?

Tests/Source/ManagedObjectContext/Migrations/DatabaseBaseTest.m
mikeger: Unfortunately no, then CoreData begins to crash.

Tests/Source/ManagedObjectContext/Migrations/DatabaseMovingTests.m
mikeger: There is no need to prepare local store when there is no local store, see changes in `needsToPrepareLocalStoreInDirectory` implementation

Tests/Source/Model/Conversation/ZMConversationTests+OTR.m
```diff
@@ -59,31 +59,36 @@ - (void)testThatConversationInitialSecurityLevelIsNotSecured
...
-        UserClient *selfClient = [self createSelfClient];
+        UserClient *selfClient = [self createSelfClientOnMOC:self.syncMOC];
         
         // when
         [selfClient trustClients:[NSSet setWithObjects:[[users.firstObject clients] anyObject], [[users.lastObject clients] anyObject], nil]];
-        
-        // then
-        XCTAssertEqual(conversation.securityLevel, ZMConversationSecurityLevelSecure);
+
+        [self.syncMOC saveOrRollback];
```
typfel: Why save and check on UI context only in this test?  for example the `testThatItDoesNotIncreaseTheSecurityLevelIfAConversationContainsUsersWithoutAConnection`test does all checks on the Sync context.

mikeger: Good catch! I guess it was not working for other reason I fixed, now it works on sync thread correctly. Let me commit the change.

Tests/Source/Model/Conversation/ZMConversationTests+OTR.m
```diff
@@ -252,14 +262,14 @@ - (void)testThatIncreasesSecurityLevelOfCreatedGroupConversationWithAllParticipa
...
+        NSArray <ZMUser *> *users = [self createUsersWithClientsOnSyncMOC];
         NSSet *clients = [users.firstObject.clients setByAddingObjectsFromSet:users.lastObject.clients];
-        UserClient *selfClient = [self createSelfClient];
+        UserClient *selfClient = [self createSelfClientOnMOC:self.syncMOC];
         [selfClient trustClients:clients];
         
         // when
         ZMConversation *conversation = [ZMConversation insertGroupConversationIntoManagedObjectContext:self.syncMOC withParticipants:users];
-        
+        [self.syncMOC saveOrRollback];
```
typfel: Don't see why a save should be necessary here. The trust is checked in `insertGroupConversationIntoManagedObjectContext:`

mikeger: Updated

Tests/Source/ManagedObjectContext/Migrations/DatabaseBaseTest.m
mikeger: I will retry, may be after other changes it would stop crashing.

mikeger: Ah no, still crashing. If breaks when it deallocates the MOC, it tries to do something with store (fault/unfault) and hits the missing file.

typfel: You could put it after the `resetSharedPersistentStoreCoordinator` in `cleanUp`? that should be equivalen, no?

mikeger: Worked, I was sceptical first because the crash happens on background thread and `resetSharedPersistentStoreCoordinator` does not really wait for them, but seems to work, so thanks for the tip!

marcoconti83:
Merged, I will release with old scripts

marcoconti83:
Released as 17.0.4

mikeger:
@marcoconti83 old-school scripts

PR #44, opened 2016-09-13
Fix/crash zios 7271

marcoconti83:
The `reactions` key on `ZMMessage` was not marked as ignored, causing sync engine to try to upload it

# Changes
Ignore the `reactions` key (it was ignored only in `ZMSystemMessage`, it should have been ignored in the `ZMMessage` superclass instead)

# Known issues
Removing a key from the list of keys to synchronize (in this case, adding it to the ignore list) will cause the bitfield holding the list of keys that need to be synchronized upstream to go out of sync, potentially causing one other random key to be marked as needing to upload instead. This is undesired and might cause erratic behaviour for that one message, but it is better than the crash loop that we have if we without this PR.


marcoconti83:
Wrong branch!

PR #46, opened 2016-09-14
Port to Swift 3

basine:


Source/Model/Message/ConversationMessage+Reaction.swift
```diff
@@ -19,33 +19,37 @@
 import Foundation
 
 
+public extension UUID {
```
typfel: Should we not move this extension to transport?

Source/Model/UserClient/UserClient.swift
```diff
@@ -91,24 +107,24 @@ public class UserClient: ZMManagedObject, UserClientType {
         // Fetch fingerprint if not there yet (could remain nil after fetch)
         if let managedObjectContext = self.managedObjectContext,
             let _ = self.remoteIdentifier
-            where managedObjectContext.zm_isSyncContext && self.fingerprint == .None
+            , managedObjectContext.zm_isSyncContext && self.fingerprint == .none
         {
             self.fingerprint = self.fetchFingerprint()
```
daehn: Bad bot

Source/Model/Message/ConversationMessage+Reaction.swift
basine: Yeah we should

Source/Model/Confirmation/ZMMessageConfirmation.swift
```diff
@@ -21,43 +21,43 @@ import CoreData
...
         case .READ:
-            return .Read
+            return .read
         }
     }
 }
 
 @objc(ZMMessageConfirmation)
-public class ZMMessageConfirmation: NSManagedObject {
+open class ZMMessageConfirmation: NSManagedObject {
```
daehn: I think this could stay `public` as we never subclass it

daehn:
The `.travis.yml` should be updated to use the `xcode8` image and an iOS 10 Simulator.

Source/Model/Message/AssetCache.swift
```diff
@@ -72,39 +72,39 @@ struct PINAssetCache : Cache {
...
     
-    func hasDataForKey(key: String) -> Bool {
+    func hasDataForKey(_ key: String) -> Bool {
         return self.assetData(key) != nil
     }
 }
 
 // MARK: - Image assets
-public class ImageAssetCache : NSObject {
+open class ImageAssetCache : NSObject {
```
typfel: This can stay public I think. Same thing goes for the methods in this class.

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -46,48 +44,50 @@ extension ZMMessage {
...
-        guard !isZombieObject, let sender = sender where sender.isSelfUser else { return }
+        guard !isZombieObject, let sender = sender , sender.isSelfUser else { return }
         guard let conversation = conversation else { return }
         
         // We insert a message of type `ZMMessageDelete` containing the nonce of the message that should be deleted
         let deletedMessage = ZMGenericMessage(deleteMessage: nonce.transportString(), nonce: NSUUID().transportString())
         
-        conversation.appendGenericMessage(deletedMessage, expires:false, hidden: true)
-        removeMessageClearingSender(true)
+        conversation.append(deletedMessage!, expires:false, hidden: true)
```
typfel: Why is `deletedMessage` optional here. Maybe we can the `ZMGenericMessage` constructor.

Tests/Source/Model/Observer/ObjectObserverToken/GlobalConversationObserverTests.swift
```diff
@@ -381,28 +381,29 @@ class GlobalConversationObserverTests : ZMBaseManagedObjectTest {
...
         self.uiMOC.saveOrRollback()
         
         XCTAssertEqual(pendingList.count, 1)
         XCTAssertEqual(normalList.count, 0)
 
         // when
-        conversation.connection.status = .Accepted
-        conversation.conversationType = .OneOnOne
+        conversation.connection!
+            .status = .accepted
```
daehn: The newline could be removed

Tests/Source/Model/Observer/ObjectObserverToken/MessageObserverTokenTests.swift
```diff
@@ -154,58 +154,58 @@ class MessageObserverTokenTests : ZMBaseManagedObjectTest {
...
+        sender.remoteIdentifier = UUID.create()
+        let message = ZMMessage.insertNewObject(in: self.uiMOC)
         message.sender = sender
-        sender.smallProfileRemoteIdentifier = NSUUID.createUUID()
+        sender.smallProfileRemoteIdentifier = UUID.create()
         
         // when
         self.checkThatItNotifiesTheObserverOfAChange(message,
-            modifier: { $0.sender!.imageSmallProfileData = self.verySmallJPEGData()},
+            modifier: { $0.sender!.imageSmallProfileData = self.verySmallJPEGData()!},
```
daehn: Maybe `verySmallJPEGData` could return an implicitly unwrapped optional or a non optional and assert internally if something goes wrong.

Source/Model/Message/ZMFileMetadata.swift
```diff
@@ -127,16 +147,16 @@ extension ZMFileMetadata {
     
     public var filename : String {
         get {
-            return  fileURL.lastPathComponent ?? "unnamed"
+            return  fileURL.lastPathComponent
```
typfel: `lastPathComponent` is an empty string if it's missing. We should keep the behaviour of returning `"unnamed"` in that case.

Source/Model/User/ZMUser.m
```diff
@@ -956,6 +956,12 @@ - (void)setImageData:(NSData *)imageData forFormat:(ZMImageFormat)format propert
 - (void)setImageData:(NSData *)imageData forKey:(NSString *)key format:(ZMImageFormat)format
 {
     [self willChangeValueForKey:key];
+    
+    if (imageData == nil) {
```
typfel: This is new behaviour. Does it fix a bug?

Source/Model/Message/ZMFileMetadata.swift
basine: Hmm I didn't make that change. Can this be automatic?

basine: Ah lastPathComponent is not optional anymore!

Source/Model/User/ZMUser.m
basine: It was broken after porting because the methods called from ObjC expected non-optional types. I checked in the PinCache project what happens when you call setValueForKey with nil and decided it was safer to use the removeValueForKey instead.

Source/Model/Confirmation/ZMMessageConfirmation.swift
basine: Yeah I did not review the public vs open changes. I assumed we can fix this later.

Source/Model/Message/AssetCache.swift
basine: I did not review public vs open (since there are too many). I assumed we can do this later since we need to port zmessaging ASAP.

Source/Model/UserClient/UserClient.swift
```diff
@@ -39,30 +39,46 @@ public let ZMUserClientRemoteIdentifierKey = "remoteIdentifier"
...
     @NSManaged public var messagesMissingRecipient: Set<ZMMessage>?
     @NSManaged public var numberOfKeysRemaining: Int32
     @NSManaged public var activationAddress: String?
-    @NSManaged public var activationDate: NSDate?
+    @NSManaged public var activationDate: Date?
     @NSManaged public var model: String?
     @NSManaged public var deviceClass: String?
-    @NSManaged public var activationLocationLatitude: Double
-    @NSManaged public var activationLocationLongitude: Double
+    @NSManaged public var activationLocationLatitude: NSNumber?
```
basine: Because UserClient subclasses from ZMManagedObject and it was complaining that those non-optional values are not initialized when calling super.init. So I had to use a type that can be optional while still being visible from ObjC (Double? is not). I made some subsequent changes to the model though - so maybe it works now.

Tests/Source/Model/Observer/ObjectObserverToken/MessageObserverTokenTests.swift
basine: It's defined in ObjC in ZMTesting.

Cartfile.resolved
```diff
@@ -3,11 +3,11 @@ github "wireapp/libPhoneNumber-iOS" "0.8.12"
...
 github "wireapp/ono" "1.2.3"
 github "wireapp/protobuf-objc" "1.9.11"
-github "wireapp/wire-ios-cryptobox" "3.1.4"
-github "wireapp/wire-ios-system" "14.40.4"
-github "wireapp/wire-ios-link-preview" "3.7.0"
-github "wireapp/wire-ios-utilities" "10.30.3"
-github "wireapp/wire-ios-images" "4.18.5"
-github "wireapp/wire-ios-testing" "6.31.6"
-github "wireapp/wire-ios-transport" "9.0.1"
+github "wireapp/wire-ios-cryptobox" "4.0.0"
```
daehn: These should also be `~>` pinned in the `Cartfile` to the new major version

Source/Model/Message/ZMFileMetadata.swift
```diff
@@ -20,18 +20,38 @@
 import Foundation
 import MobileCoreServices
 import ZMCSystem
+fileprivate func < <T : Comparable>(lhs: T?, rhs: T?) -> Bool {
```
daehn: I can add it to `wire-ios-utilities` and make a new release

Source/Model/UserClient/UserClient.swift
basine: Nope, still does not work without these changes

typfel: Can't we just remove the optional and initialise them to some default value?

Source/Model/Message/ZMFileMetadata.swift
daehn: Done in https://github.com/wireapp/wire-ios-utilities/pull/8

Source/Model/UserClient/UserClient.swift
basine: No, if it's not optional it will complain - even if there is a default value defined in the data model. You can try it yourself ;)

Tests/Source/Model/ConversationList/ZMConversationListTests.m
```diff
@@ -526,7 +526,7 @@ - (void)testThatAddingMessageToClearedConversationMovesItToActiveConversationsLi
     ZMConversation *c1 = [ZMConversation insertNewObjectInManagedObjectContext:self.uiMOC];
     c1.conversationType = ZMConversationTypeOneOnOne;
     c1.lastModifiedDate = [NSDate date];
-    ZMMessage *message = [c1 appendMessageWithText:@"message"];
+    ZMMessage *message = (id)[c1 appendMessageWithText:@"message"];
```
typfel: Which should change this into `ZMConversationMessage`. But I guess also at some point later.

Source/Model/Message/LinkPreview+ProtocolBuffer.swift
```diff
@@ -26,8 +26,8 @@ extension LinkPreview {
         self.init(originalURLString: protocolBuffer.url, permamentURLString: protocolBuffer.permanentURLString, offset: Int(protocolBuffer.urlOffset))
     }
     
-    public var protocolBuffer: ZMLinkPreview {
-        let linkPreviewBuilder = ZMLinkPreview.builder()
+    public var protocolBuffer: ZMLinkPreview? {
+        let linkPreviewBuilder = ZMLinkPreview.builder()!
```
marcoconti83: Is there a reason why we changed this var to optional, but then force unwrap the `.builder` method? To me either we do one or the other.
I think it's fair to crash if we can't assemble a protocol buffer anyway, maybe adding the `!` if needed (maybe even in the return line if needed) is enough.

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -77,30 +74,33 @@ extension ZMGenericMessage {
...
     /// Returns a message with recipients and a strategy to handle missing clients
-    private func otrMessage(selfClient: UserClient,
+    fileprivate func otrMessage(_ selfClient: UserClient,
                             conversation: ZMConversation,
-                            externalData: NSData?,
+                            externalData: Data?,
                             sessionDirectory: EncryptionSessionsDirectory) -> (message: ZMNewOtrMessage, strategy: MissingClientsStrategy) {
         
-        let recipientUsers : [ZMUser]
+        var recipientUsers : [ZMUser] = []
```
marcoconti83: Did the behavior of uninitialized `let`s change in Swift 3? Previously I could leave this as a `let`, not initialize it here, and initialize it later in all code branches. Losing the `let` in favor of a `var` might cause unwanted overwrite to go unnoticed.

Source/Model/UserClient/UserClient.swift
```diff
@@ -177,10 +193,10 @@ public class UserClient: ZMManagedObject, UserClientType {
     /// Can be called several times without issues
     public func resetSession() {
         // Delete should happen on sync context since the cryptobox could be accessed only from there
-        UserClient.deleteSession(forClientWithRemoteIdentifier: self.remoteIdentifier, managedObjectContext: (self.managedObjectContext?.zm_syncContext)!)
+        UserClient.deleteSession(forClientWithRemoteIdentifier: self.remoteIdentifier, managedObjectContext: (self.managedObjectContext?.zm_sync)!)
```
marcoconti83: I would have preferred to keep `zm_syncContext` because `zm_sync` doesn't really say much (what is a "sync"?). I know that "context" is redundant as we already have the type to tell that this is a `NSManagedObject`, but `zm_sync` to me still sounds as it's missing something.

Tests/Source/Model/Messages/ZMClientMessageTests+OTR.swift
```diff
@@ -186,25 +182,23 @@ extension ClientMessageTests_OTR {
...
             textMessage.senderClientID = senderID
             let confirmationMessage = textMessage.confirmReception()
             
             //when
-            guard let payloadAndStrategy = confirmationMessage.encryptedMessagePayloadData() else {
-                XCTFail()
-                return
-            }
+            guard let payloadAndStrategy = confirmationMessage?.encryptedMessagePayloadData()
+            else { return XCTFail()}
```
typfel: Does this return & fail or just return?

Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
```diff
@@ -74,14 +74,14 @@ public class GeneralConversationChangeInfo : ObjectChangeInfo {
         var dict : Dictionary<String, NSObject?> = [:]
         
         for key in keys {
-            dict[key] = ""
+            dict[key] = "" as NSObject??
```
marcoconti83: I'm confused, why double optional here?

Tests/Source/Model/Messages/ZMClientMessageTests+OTR.swift
daehn: Returns and fails

Source/Notifications/ObjectObserverTokens/Helpers/SetSnapshot.swift
```diff
@@ -69,34 +69,34 @@ struct SetStateUpdate {
 
  struct SetSnapshot {
     
-    let set : OrderedSet<NSObject>
+    let set : NSOrderedSet
     let moveType : ZMSetChangeMoveType
     
-    init(set: OrderedSet<NSObject>,  moveType : ZMSetChangeMoveType) {
-        self.set = set
+    init(set: NSOrderedSet,  moveType : ZMSetChangeMoveType) {
```
marcoconti83: A very annoying person would point out that there are two spaces here :)

Source/Notifications/ZMNotifications.m
```diff
@@ -167,7 +167,10 @@ @implementation ZMMessageNotification
 
 + (id<ZMMessageObserverOpaqueToken>)addMessageObserver:(id<ZMMessageObserver>)observer forMessage:(id<ZMConversationMessage>)conversationMessage;
 {
-    return (id) [[MessageObserverToken alloc] initWithObserver: observer object:conversationMessage];
+    if ([conversationMessage isKindOfClass:[ZMMessage class]]) {
+    return (id) [[MessageObserverToken alloc] initWithObserver: observer object:(ZMMessage *)conversationMessage];
```
marcoconti83: Indentation

Tests/Source/Model/Observer/ManagedObjectContextObserverTests.swift
```diff
@@ -298,36 +298,36 @@ class ManagedObjectContextObserverTests : ZMBaseManagedObjectTest {
...
         self.uiMOC.globalManagedObjectContextObserver.isTesting = true
-        let filename = "foo.mp4"
-        let documents = NSSearchPathForDirectoriesInDomains(.DocumentDirectory, .UserDomainMask, true).first!
-        let documentsURL = NSURL(fileURLWithPath: documents)
-        let fileURL =  documentsURL.URLByAppendingPathComponent(filename)
-        verySmallJPEGData().writeToURL(fileURL, atomically: true)
-        defer { try! NSFileManager.defaultManager().removeItemAtURL(fileURL) }
-        
-        let conversation = ZMConversation.insertNewObjectInManagedObjectContext(self.uiMOC)
+        let filename = "video.mp4"
```
marcoconti83: "Foo" is not cool anymore? :) The change is OK, I was wondering if there is any reason why we changed it (or if it was just casual).

Tests/Source/Model/UserClient/UserClientKeyStoreTests.swift
```diff
@@ -122,15 +120,15 @@ class UserClientKeysStoreTests: OtrBaseTest {
...
+        try! FileManager.default.createDirectory(atPath: UserClientKeysStore.legacyOtrDirectory.path, withIntermediateDirectories: true, attributes: [:])
     }
     
     func testThatTheNonEmptyLegacyOTRFolderIsDetected() {
         
         // given
-        self.dynamicType.createFakeOTRFolder()
-        "foo".dataUsingEncoding(NSUTF8StringEncoding)!.writeToURL(UserClientKeysStore.legacyOtrDirectory.URLByAppendingPathComponent("dummy.txt"), atomically: true)
+        type(of: self).createFakeOTRFolder()
+        try? "foo".data(using: String.Encoding.utf8)!.write(to: UserClientKeysStore.legacyOtrDirectory.appendingPathComponent("dummy.txt"), options: Data.WritingOptions.atomic)
```
marcoconti83: Should we use `try!` instead? This way it will not silently fail.

Tests/Source/Model/UserClient/UserClientKeyStoreTests.swift
```diff
@@ -170,17 +168,17 @@ class UserClientKeysStoreTests: OtrBaseTest {
     func testThatItMovesTheLegacyCryptobox() {
         
         // given
-        self.dynamicType.cleanOTRFolder()
+        type(of: self).cleanOTRFolder()
 
-        self.dynamicType.createFakeOTRFolder()
-        "foo".dataUsingEncoding(NSUTF8StringEncoding)!.writeToURL(UserClientKeysStore.legacyOtrDirectory.URLByAppendingPathComponent("dummy.txt"), atomically: true)
+        type(of: self).createFakeOTRFolder()
+        try? "foo".data(using: String.Encoding.utf8)!.write(to: UserClientKeysStore.legacyOtrDirectory.appendingPathComponent("dummy.txt"), options: Data.WritingOptions.atomic)
```
marcoconti83: Should we use `try!` instead? This way it will not silently fail.

Tests/Source/Model/UserClient/UserClientTests.swift
```diff
@@ -164,21 +158,20 @@ class UserClientTests: ZMBaseManagedObjectTest {
...
     func testThatItDeletesASessionWhenDeletingAClient() {
         self.syncMOC.performGroupedBlockAndWait{
             // given
-            let selfClient = self.createSelfClientOnMOC(self.syncMOC)
+            let selfClient = self.createSelfClient(onMOC: self.syncMOC)
             var preKeys : [(id: UInt16, prekey: String)] = []
             selfClient.keysStore.encryptionContext.perform({ (sessionsDirectory) in
-                preKeys = try! sessionsDirectory.generatePrekeys(Range(0..<2)).flatMap{$0}
-            })
+                preKeys = try! sessionsDirectory.generatePrekeys(CountableRange<UInt16>(0..<2))            })
```
marcoconti83: Weird indent

Tests/Source/Model/UserClient/UserClientTests.swift
```diff
@@ -272,34 +265,37 @@ class UserClientTests: ZMBaseManagedObjectTest {
...
             fetchRequest.fetchLimit = 1
             // when
-            let fetchedClient = self.syncMOC.executeFetchRequestOrAssert(fetchRequest).first as? UserClient
-            
-            // then
-            XCTAssertNotNil(fetchedClient)
-            XCTAssertNotNil(fetchedClient!.fingerprint)
+            var fetchedClient: UserClient?
+            do {
+                fetchedClient = try self.syncMOC.fetch(fetchRequest).first
```
marcoconti83: The definition of `fetchedClient` can be moved inside the do and turned into a `let`

Source/Model/Message/ConversationMessage+Reaction.swift
daehn: I am on it

Source/Model/Message/ConversationMessage+Deletion.swift
basine: https://github.com/wireapp/wire-ios-protos/pull/10

Tests/Source/Model/Observer/ObjectObserverToken/MessageObserverTokenTests.swift
basine: https://github.com/wireapp/wire-ios-testing/pull/5

Source/Model/Message/ZMFileMetadata.swift
basine: Fixed

Cartfile.resolved
basine: Fixed

Source/Model/Message/ZMClientMessage+Encryption.swift
basine: I guess that was an automatic change.

Source/Model/UserClient/UserClient.swift
basine: Bridging from ObjC unfortunately drops the `Context` in Swift

Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
basine: That was an automatic change

Source/Notifications/ObjectObserverTokens/Helpers/SetSnapshot.swift
basine: Which were there before ;)

Tests/Source/Model/Observer/ManagedObjectContextObserverTests.swift
basine: I thought the test broke because of the missing file. So I added a file with that name - turned out something else was broken. Didn't revert the change.

Tests/Source/Model/UserClient/UserClientKeyStoreTests.swift
basine: maybe a if (try? ...) instead

Source/Model/Message/LinkPreview+ProtocolBuffer.swift
basine: My guess is that it was an automatic change.

Source/Model/Message/ConversationMessage+Reaction.swift
daehn: Done in https://github.com/wireapp/wire-ios-transport/pull/7

Tests/Source/Model/Observer/ObjectObserverToken/GlobalConversationObserverTests.swift
basine: Fixed

Source/Model/Message/ZMClientMessage+Encryption.swift
basine: I guess that's how you do it now - no more declaring a let that is set afterwards.

daehn: If all branches initialize it it should still be possible

daehn: and here there is one branch where it would not be set now (if the `fetch` call returns `nil`)

marcoconti83: It should work if you add:
```
} else {
    recipientUsers = nil
}
```

after
```
if let message = ZMMessage.fetch(withNonce: messageID, for: conversation, in: conversation.managedObjectContext) {
     recipientUsers = [message.sender].flatMap { $0 }
}
```

basine:
I updated the dependencies ZMUtilities, ZMTransport, ZMProtos and ZMTesting and made the changes in ZMDataModel

daehn:
🙌

PR #52, opened 2016-09-23
Enable extension API only flag

typfel:
Enabling this flag requires some changes like not accessing the shared UIApplication.

- ManagedObjectContextObserver no longer checks the application state to decide when to
  process changes. Instead the new property `propagateChanges` should be set by the
  owner of the context.

- We no longer create an UIAlertView when we can't open the database. Instead we send
  a notification which the UI can listen to.


Tests/Source/Model/Observer/ManagedObjectContextObserverTests.swift
```diff
@@ -26,6 +26,7 @@ class ManagedObjectContextObserverTests : ZMBaseManagedObjectTest {
         super.setUp()
         self.setUpCaches()
         
+//        self.uiMOC.globalManagedObjectContextObserver.propagateChanges = true
```
daehn: Can be removed

mikeger: Can be removed / should be uncommented?

PR #53, opened 2016-09-26
Return message nonce as imageDataIdentifier in case there is an original image

daehn:
# What's in this PR?

* Link previews did not show the image on the sender side, during the migration to Swift 3 we needed to optionally bind the `imageDataIdentifier` which is only present after the image has been uploaded. As we do not receive an additional notification when the image has been uploaded and also want to dimply the (original) image before it has been uploaded on the sender side the `imageDataIdentifier` property now returns the message nonce if there is an original image but no assetId for the link preview.
* Convert an enum to the new Swift 3.0 naming guidelines.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ branches:
 install:
   - bundle install
   - curl -O https://raw.githubusercontent.com/wireapp/wire-ios-shared-resources/master/Dangerfile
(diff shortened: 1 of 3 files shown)


mikeger:
@daehn `dimply` should be understood as `simply`?

daehn:
@mikeger fixed, should have been `display` ;)

PR #58, opened 2016-10-05
Disable sending native pushes for delivery receipts

daehn:
# What's in this PR?

Some users reported "New message" notifications without having received any new visible content inside the application. These notifications are received when we fail to pingback to the backend after receiving a `VoIP` notification, the most probable scenario is that this happens when we receive a lot of delivery receipt pushes and fail to pingback either because of bad network conditions or because we are running in the background for too long.

As there is no actual need to send native pushes for delivery receipts (as they are only visible in the UI and should not generate a local notification) we now no longer are sending them.


typfel:
Can we add test for this?

daehn:
@typfel I was thinking about this as well, I guess we could directly test this function if we make it `internal` instead of `file private`, otherwise it won't be visible even with `@testable`

daehn:
@mikeger @typfel I added a test, please re-review.

daehn:
Released as `19.0.6`

PR #59, opened 2016-10-06
ZIOS-7352 Fix race condition crash

Einh06:
**What's in this PR**
This PR fixes a crash that would occurs when initializing `NSManagedObjectContext`. The `ZMSession` and SelfUser would be initialize asynchronously for both the UI context and the Sync context, sometime resulting in a race condition, triggering and Primary Key exception in Core Data because object where created in multiple thread.
This is fixed by enforcing initialization to be done serially.

**Also**
 Remove switching to main queue to call completion handler
The responsibility of moving to the correct queue was moved to the caller. This removes hidden async calls inside the function that can not be monitored (and waited on) for the tests tear down.


Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -100,12 +100,10 @@ + (void)prepareLocalStoreInternalBackingUpCorruptedDatabase:(BOOL)backupCorrupte
     [self setDatabaseDirectoryURL:directory];
 
     dispatch_block_t finally = ^() {
-        dispatch_async(dispatch_get_main_queue(), ^{
```
typfel: Is removing this async call the main fix?

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -176,27 +174,23 @@ + (NSPersistentStoreCoordinator *)requirePersistentStoreCoordinatorInDirectory:(
...
-        }
-    });
-    return result;
+    NSPersistentStoreCoordinator *psc = [self requirePersistentStoreCoordinatorInDirectory:storeDirectory];
+    RequireString(psc != nil, "No persistent store coordinator, call -prepareLocalStore: first.");
+    
+    SharedUserInterfaceContext = [[NSManagedObjectContext alloc] initWithConcurrencyType:NSMainQueueConcurrencyType];
+    [SharedUserInterfaceContext markAsUIContext];
+    [SharedUserInterfaceContext configureWithPersistentStoreCoordinator:psc];
+    [SharedUserInterfaceContext initialiseSessionAndSelfUser];
```
typfel: So this is now the point where the session and self user is created so we now always create the UI context before any other context.

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
marcoconti83: No, it's to have less flaky tests. Every `dispatch_async(dispatch_get_main_queue()` in our code adds to the flakiness of the tests, they should all be removed.

PR #60, opened 2016-10-11
ZIOS-7400 crash nil user

Einh06:
**What's in this PR**
This PR tries to find from multiple sources the sender of the original message before encrypting the confirmation. The message won't be uploaded if no sender is infered, so we will have a sender at this point.


Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -96,12 +96,32 @@ extension ZMGenericMessage {
...
+                    if sender != nil { return sender }
+                    if conversation.connectedUser != nil { return conversation.connectedUser }
+                    if conversation.otherActiveParticipants.count > 0 { return conversation.otherActiveParticipants.firstObject! as? ZMUser }
+                    return nil
+                }()
+    
+                if let recipient = recipient {
+                    recipientUsers = [recipient]
+                } else {
+                    fatal("confirmation cannot have not recipient")
```
marcoconti83: Let's add extra logging here. You can use string interpolation to add conversation ID, type, etc. so when it crashes we get all the info we need to debug why there's no sender, no connection and no participants!

typfel:
Can you explain the scenario when connectedUser would be nil. It seems a bit complicated to look for the recipient in three places.

Einh06:
@typfel In the case we receive a message from a conversation we dont have locally yet, we create a conversation object without the connection element yet as we will need to fetch it later, we still want to send back delivery receipt in that case.

Source/Model/Message/ZMClientMessage+Encryption.swift
Einh06: Indeed, we should try to know as much as possible in that case. Will do

PR #61, opened 2016-10-13
Ephemeral messages

basine:
**What's in this PR**

### Sending ephemeral messages
The following messages will have ephemeral behaviour: Text, Knock, Location, Asset, ImageAsset

Ephemeral on the sending side means:
1) The sender updates the `messageDestructionTimeout` to one of the preset time intervals by calling `conversation.updateMessageDestructionTimeout(:)`
2) While the conversation has a messageDestructionTimeout > 0, we set an expiration timeout on the mentioned message types when they are created.
3) The message makes a roundtrip to the backend. When we receive the response success, we start an obfuscationTimer (`ZMMessageDestructionTimer`) and set the `destructionDate` on the message
4) When the timer fires, we call a timerCompletionHandler which replaces the genericMessage content with obfuscated content via `message.obfuscate()`

### Receiving ephemeral messages
1) Starting the deletion on the receiving side is triggered by the UI by calling `message.startMessageDestructionTimerIfNeeded()`
Internally we check if the message matches the following conditions:
a) ZMClientMessages: As soon as we have the first part
b) ZMAssetClientMessages: For assets we wait until we have the uploaded or uploadedCancelled. For images we wait until we have the mediumGenericMessage
2) We start a deletionTimer (`ZMMessageDestructionTimer`) for the timeout specified in the genericMessage (rounded to the closest predefined ZMConversationMessageDestructionTimeout) and set the `destructionDate` on the message
3) When the timer fires, we delete the genericMessage content and send a delete for all to all clients.

When the app starts we are checking for messages that had timers running when the app got killed and either obfuscate / delete them or restart their timer if the destructionDate is in the future.

- [x] Needs ZMProtos release


Source/Model/Conversation/ZMConversation+Ephemeral.swift
```diff
@@ -0,0 +1,79 @@
...
+            start += 1
+            if currentTimeout == timeout {
+                return timeout
+            }
+            if currentTimeout < timeout {
+                lastTimeout = currentTimeout
+            } else {
+                if (currentTimeout - timeout) < (timeout - lastTimeout) {
+                    return currentTimeout
+                }
```
mikeger: I think you can add
```
else {
    return lastTimeout
}
```
to prevent extra looping to the end.

Source/Model/Conversation/ZMConversation+Ephemeral.swift
```diff
@@ -0,0 +1,79 @@
...
+        return lastTimeout
+    }
+}
+
+public extension ZMConversation {
+
+    /// Sets messageDestructionTimeout
+    /// @param timeout The timeout after which an appended message should "self-destruct"
+    public func updateMessageDestructionTimeout(timeout : ZMConversationMessageDestructionTimeout) {
+        if (conversationType != .oneOnOne) {
```
mikeger: I think it would be more swifty to do the same with `guard`.

Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
```diff
@@ -0,0 +1,123 @@
...
+
+enum MessageDestructionType : String {
+    static let UserInfoKey = "destructionType"
+    
+    case obfuscation, deletion
+}
+
+public class ZMMessageDestructionTimer : ZMMessageTimer {
+
+    internal var isTesting : Bool = false
```
mikeger: I don't like having the vars like that, but it's ok if there is no other way around.

Source/Utilis/Protos/ZMGenericMessage+Utils.h
```diff
@@ -0,0 +1,105 @@
...
+
+
++ (ZMGenericMessage *)messageWithEmojiString:(NSString *)emojiString
+                                   messageID:(NSString *)messageID
+                                       nonce:(NSString *)nonce;
+- (BOOL)knownMessage;
+
+@end
+
+NS_ASSUME_NONNULL_END
```
mikeger: Usually this goes to the end of the file; typo or as designed to keep here?

Tests/Source/Model/Messages/BaseClientMessageTests.swift
```diff
@@ -20,6 +20,40 @@
 import Foundation
 @testable import ZMCDataModel
 
+/*
```
mikeger: Needed for the future? Or can be removed / uncommented?

ZMCDataModel.xcodeproj/project.pbxproj
```diff
@@ -2316,7 +2394,7 @@
 				F9331C901CB42FCF00139ECC /* zmessaging1.28.xcdatamodel */,
 				F9331C911CB42FCF00139ECC /* zmessaging2.3.xcdatamodel */,
 			);
-			currentVersion = F97E28BC1D8AA27700A77567 /* zmessaging2.17.0.xcdatamodel */;
+			currentVersion = F963E95C1D9AB80B00098AD3 /* zmessaging2.18.0.xcdatamodel */;
```
mikeger: Would we need a migration test?

Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
basine: I don't like it particularly either. If you have a different solution for using 0.5 s timeouts for testing when it's supposed to round up or down to the closest predefined timeout, then I am gladly implementing it.

Source/Utilis/Protos/ZMGenericMessage+Utils.h
basine: This file was copied from ZMProtos.

Source/Model/Conversation/ZMConversation+Ephemeral.swift
basine: true, thanks

Tests/Source/Model/Messages/BaseClientMessageTests.swift
basine: forgot to remove it

Source/Model/Conversation/ZMConversation+Ephemeral.swift
```diff
@@ -0,0 +1,79 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+@objc public enum ZMConversationMessageDestructionTimeout : Int {
+    case none, fiveSeconds, fifteenSeconds, oneMinute, fiveMinutes, fifteenMinutes
```
basine: Interesting fact: looping through an enum also only works when the enum is defined in Swift and not in ObjC. In ObjC it will continue iterating past the last case and then overflow the iterator.

Source/Model/Message/ZMMessage.m
```diff
@@ -1063,3 +1072,84 @@ - (BOOL)shouldGenerateUnreadCount;
...
+    if (isSelfUser && self.managedObjectContext.zm_isSyncContext) {
+        self.destructionDate = [NSDate dateWithTimeIntervalSinceNow:self.deletionTimeout];
+        ZMMessageDestructionTimer *timer = self.managedObjectContext.zm_messageObfuscationTimer;
+        [timer startObfuscationTimerWithMessage:self timeout:self.deletionTimeout];
+        return YES;
+    }
+    else if (!isSelfUser && self.managedObjectContext.zm_isUserInterfaceContext){
+        ZMMessageDestructionTimer *timer = self.managedObjectContext.zm_messageDeletionTimer;
+        NSTimeInterval matchedTimeInterval = [timer startDeletionTimerWithMessage:self timeout:self.deletionTimeout];
+        self.destructionDate = [NSDate dateWithTimeIntervalSinceNow:matchedTimeInterval];
```
marcoconti83: Shouldn't we set the destruction date before starting the timer, in case the (async) timer fires before the instruction that sets the destruction date fires? Or will the timer switch back to this queue anyway when it fires?

Source/Model/Message/ZMMessage.m
```diff
@@ -1063,3 +1072,84 @@ - (BOOL)shouldGenerateUnreadCount;
...
+    fetchRequest.predicate = [NSPredicate predicateWithFormat:@"%K != nil AND %K != nil AND %K == FALSE",
+                              ZMMessageDestructionDateKey,          // If it has a destructionDate, the timer did not fire in time
+                              ZMMessageSenderKey,                   // As soon as the message is deleted, we would delete the sender
+                              ZMMessageIsObfuscatedKey];            // If the message is obfuscated, we don't need to obfuscate it again
+    return fetchRequest;
+}
+
++ (void)deleteOldEphemeralMessages:(NSManagedObjectContext *)context
+{
+    NSFetchRequest *request = [self fetchRequestForEphemeralMessagesThatNeedToBeDeleted];
```
marcoconti83: This is called at startup, right? Is there any concern about performances here, as it has to go through ALL messages in history? Is there any index we can set on the DB to speed up the query?

Source/Model/Conversation/ZMConversation+Ephemeral.swift
basine: We do this is multiple places, so we are def f$&#ed when this happens. But I can add this observation as a comment.

Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
```diff
@@ -0,0 +1,123 @@
...
+            strongSelf.messageTimerDidFire(message: message, userInfo:userInfo)
+        }
+    }
+    
+    func messageTimerDidFire(message: ZMMessage, userInfo: [AnyHashable: Any]?) {
+        guard let userInfo = userInfo as? [String : Any],
+              let type = userInfo[MessageDestructionType.UserInfoKey] as? String
+        else { return }
+        
+        switch MessageDestructionType(rawValue:type) {
```
basine: Almost. The obfuscation timer starts as soon as we receive the BE response that the message was received by the BE - not by the other user. So it would be deliveryState sent (not delivered).

Source/Model/Message/ZMGenericMessageImageOwner.swift
```diff
@@ -41,18 +41,18 @@ import ZMProtos
     }
     
     fileprivate var encrypted : Bool {
-        return self.genericMessage.hasImage() && self.genericMessage.image.otrKey.count > 0
+        return self.genericMessage.imageAssetData != nil && self.genericMessage.imageAssetData?.otrKey.count > 0
```
basine: hasImage() is a property on ZMGenericMessage and returns true when the message has an ImageAsset, however for ephemeral messages hasImage() returns false. You have to check for hasEphemeral && ephemeral.hasImage()
The imageAssetData property is a computed property which is doing this for us. It either returns image or ephemeral.image

Source/Model/Message/ZMMessage.m
basine: These calls are synchronous, so it should not make a difference. For incoming messages we have to match the timeout we receive against our predefined values. It was easiest for testing purposes to put this matching into the timer and return the matched value instead of putting the matching here. I am open for suggestions of solving this differently. 
In theory - neither the UI nor the sync-engine is limited by the predefined values. But we agreed on matching unknown values to those that were predefined. So I am sticking with the agreement.

Source/Model/Message/ZMMessage.m
basine: It is only fetching messages with destructionDate set. If we can set an index for that, then we should probably do that. 
Independent from that, I think we should limit the results also to messages that are visible in the conversation or set the deletionDate to nil when we "delete" the message.

basine: I will set the deletionDate to nil - then we should have fewer messages to fetch.

Source/Model/Conversation/ZMConversation+Ephemeral.swift
Einh06: Or we can actually change it when we see it to, you know, to prevent a potential catastrophe.

daehn:
@mikeger I changed the enum base type, please re-review

marcoconti83:
I finished looking at the tests. That's a lot of test. Good job folks! 👍

Source/Model/Message/ConversationMessage.swift
```diff
@@ -93,6 +93,26 @@ public protocol ZMConversationMessage : NSObjectProtocol {
...
+    /// Returns YES if a timer was started by the message call
+    func startSelfDestructionIfNeeded() -> Bool
+    
+    /// Returns true if the message is ephemeral
+    var isEphemeral : Bool { get }
+    
+    /// If the message is ephemeral, it returns a fixed timeout
+    /// Otherwise it returns -1
+    /// Override this method in subclasses if needed
+    var deletionTimeout : TimeInterval { get }
```
basine: A bit too late for such a suggestion. Remember, 4000 lines of code ;)
But thanks for the impulse for next time.

PR #63, opened 2016-10-19
CallKit support

mikeger:
- Support for observation in background. It is needed in order to observe voice channel state when the app is in the background.
- Support for fetching users by email or by the phone number. Needed to respond to handoff from phone app.


marcoconti83:
> Support for observation in background

Uh-oh. This is gonna make iOS unhappy. If we create all the observers when we are launched in the BG because of APNS, iOS will penalize us. 
Is there a way to initialize it only if we are being started from CallKit, but not when we have been started from APNS?

mikeger:
@marcoconti83 it does not start all the observers on the background, only voice channel observers. Also the purpose of having the observers is to tell CallKit that we have a call when appropriate, so we always need to observe the voice channel states.

marcoconti83:
@mikeger thanks for clarifying!

Source/Model/User/ZMUser+Internal.h
```diff
@@ -51,6 +50,8 @@ extern NSString * __nonnull const SessionObjectIDKey;
 - (void)updateWithTransportData:(nonnull NSDictionary *)transportData authoritative:(BOOL)authoritative;
 
 + (nullable instancetype)userWithRemoteID:(nonnull NSUUID *)UUID createIfNeeded:(BOOL)create inContext:(nonnull NSManagedObjectContext *)moc;
++ (nullable instancetype)userWithEmailAddress:(nonnull NSString *)emailAddress inContext:(nonnull NSManagedObjectContext *)context;
++ (nullable instancetype)userWithPhoneNumber:(nonnull NSString *)phoneNumber inContext:(nonnull NSManagedObjectContext *)context;
```
marcoconti83: FYI there is a plan to remove phone numbers and emails of contacts, we will need to use address book contact matching instead. But I guess we will worry about it when we get there.

Source/Notifications/ObjectObserverTokens/GlobalConversationObserver.swift
```diff
@@ -129,7 +135,7 @@ final class GlobalConversationObserver : NSObject, ObjectsDidChangeDelegate, ZMG
                 _ = try? conv.managedObjectContext?.obtainPermanentIDs(for: [conv])
             }
             if self.conversationTokens[conv.objectID] == nil {
-                self.conversationTokens[conv.objectID] = GeneralConversationObserverToken<GlobalConversationObserver>(observer: self, conversation: conv)
+                self.conversationTokens[conv.objectID] = GeneralConversationObserverToken<GlobalConversationObserver>(observer: self, conversation: conv, managedObjectContextObserver: self.managedObjectContextObserver!)
```
marcoconti83: I'm not sure, but I think that if your code goes down this path, it will initiate the full shebang: observing all users in the conversation for each conversation, etc. eventually observing a lot of stuff even if you only needed to observe the voice channel. This will cause slowdown on start that is a risk for APNS.
@basine @silvan do you remember if this is the case?

mikeger:
# iPhone 6s

| --- _No CallKit_  |  |         |
| -----------------| ------ | ------- |
| Foreground 1     | 45 MB  | 2.8 sec |
| Foreground 2     | 45 MB  | 2 sec |
| Foreground 3     | 45.3 MB| 2.3 sec |
| Foreground 4     | 44.4 MB| 2 sec |
| Foreground 5     | 45.4 MB| 2.2 sec |
| Foreground 6     | 45.6 MB| 2.5 sec | 
| *Foreground - AVG* | 45 MB| 2.3 sec | 
| Background 1     | 9.8 MB | 1 sec |
| Background 2     | 9.7 MB | 0.75 sec |
| Background 3     | 9.8 MB | 0.85 sec |
| Background 4     | 9.8 MB | 0.9 sec |
| Background 5     | 9.8 MB | 0.9 sec |
| Background 6     | 9.9 MB | 0.8 sec | 
| *Background - AVG* | 9.8 MB| 0.86 sec |

| --- _CallKit_ |      |         | 
| -----------------| ------ | ------- |
| Foreground 1     | 46 MB | 2.13 sec |
| Foreground 2     | 46 MB | 2.30 sec |
| Foreground 3     | 46.2 MB | 2.34 sec |
| Foreground 4     | 46 MB | 2.03 sec |
| Foreground 5     | 44.9 MB | 1.99 sec |
| Foreground 6     | 46 MB | 2.3 sec | 
| *Foreground - AVG* | 46 MB|  sec |
| Background 1     | 12.9 MB | 1.96 sec |
| Background 2     | 12.9 MB | 2.08 sec   |
| Background 3     | 12.9 MB | 2.19 sec |
| Background 4     | 12.9 MB | 2.12 sec   |
| Background 5     | 12.9 MB |  1.35 sec |
| Background 6     | 12.9 MB  | 2.09 sec |
| *Background  - AVG*     | 12.9 MB   | 1.96 sec   |

Verdict: CallKit observation doubles the startup time in the background

mikeger:
@marcoconti83 please re-review, now the observers part is running via `managedObjectContext` (on `wire-ios-sync-engine`)

Source/Model/User/ZMUser.m
```diff
@@ -424,9 +424,50 @@ + (instancetype)userWithRemoteID:(NSUUID *)UUID createIfNeeded:(BOOL)create inCo
     }
 }
 
-+ (NSOrderedSet <ZMUser *> *)usersWithRemoteIDs:(NSOrderedSet <NSUUID *>*)UUIDs inContext:(NSManagedObjectContext *)moc;
++ (nullable instancetype)userWithEmailAddress:(NSString *)emailAddress inContext:(NSManagedObjectContext *)context
+{
+    RequireString(0 != emailAddress.length, "emailAddress required");
```
basine: is this email normalized? should we normalize it the same way we do during registration?

Source/Model/User/ZMUser.m
```diff
@@ -424,9 +424,50 @@ + (instancetype)userWithRemoteID:(NSUUID *)UUID createIfNeeded:(BOOL)create inCo
...
+        return users.firstObject;
+    }
+    else {
+        return nil;
+    }
+}
+
++ (nullable instancetype)userWithPhoneNumber:(NSString *)phoneNumber inContext:(NSManagedObjectContext *)context
 {
+    RequireString(0 != phoneNumber.length, "phoneNumber required");
```
basine: is this phone number normalized? should we normalize it the same way we do during registration?

mikeger: I was not thinking about normalisation, but I don't think it's necessary since we are not normalising the user credentials as far as I know, and the email that lands here is the email we pass to CallKit so it should exactly match to the stuff in the database. If needed, the user of the API can extra normalise data before calling this method.

mikeger:
@marcoconti83 @basine can this be merged now? Any open questions?

marcoconti83:
Is this an API breaker?

mikeger:
@marcoconti83 no, should not break the API.

PR #68, opened 2016-10-26
Move cryptobox folder

Einh06:
**What's in this PR**
This PR moves the Cryptobox directory in whatever directory it is given and changes how Cryptobox is initialize.
Formally, the `UserClientKeysStore` was lazily initialized with a hardcoded path to the OTR directory,
with a simple check of whether or not we are accessing it from the sync context.
This was changed since we need to have the OTR directory in the shared container, which path cannot be inferred. The initialization of the box is done only on the syncMOC only.


Source/Utilis/CryptoBox.swift
```diff
@@ -26,6 +26,20 @@ extension NSManagedObjectContext {
...
     
+    @objc(setupUserKeyStoreForDirectory:)
+    public func setupUserKeyStore(for directory: URL) -> Void
+    {
+        if !self.zm_isSyncContext {
+            fatal("Can't initiliazie crypto box on non-sync context")
+        }
+        
+        let sharedDirectory = directory.appendingPathComponent(UserClientKeysStore.otrFolderPrefix)
+        // Moving directories needed here
```
marcoconti83: Is this a TODO?

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -205,6 +205,7 @@ + (instancetype)createSyncContextWithStoreDirectory:(NSURL *)storeDirectory;
         [moc markAsSyncContext];
         [moc configureWithPersistentStoreCoordinator:psc];
         [moc setupLocalCachedSessionAndSelfUser];
+        [moc setupUserKeyStoreForDirectory:storeDirectory]; //TODO: implement this
```
marcoconti83: The bot has a point! :)

Source/Utilis/CryptoBox.swift
```diff
@@ -34,10 +48,10 @@ extension NSManagedObjectContext {
         let keyStore = self.userInfo.object(forKey: NSManagedObjectContext.ZMUserClientKeysStoreKey)
         if let keyStore = keyStore as? UserClientKeysStore {
             return keyStore
+        } else {
+            fatal("What?! where is your keystore?!")
```
mikeger: I think would be better to stick to more formal error messages, also would be cool to include some details on error.

PR #69, opened 2016-10-27
remove event id

basine:
- [x] needs new release of ZMTransport with https://github.com/wireapp/wire-ios-transport/pull/16


Cartfile.resolved
```diff
@@ -9,5 +9,5 @@ github "wireapp/wire-ios-link-preview" "4.0.2"
 github "wireapp/wire-ios-utilities" "13.0.0"
 github "wireapp/wire-ios-images" "7.0.0"
 github "wireapp/wire-ios-testing" "8.0.1"
-github "wireapp/wire-ios-transport" "14.0.0"
+github "wireapp/wire-ios-transport" "b574d05ce462dbe19d9e62f925edec3a73cadff7"
```
marcoconti83: This can now be bumped to 15

marcoconti83: This can now be bumped to 15 (also in the `Cartfile`)

Source/Model/Conversation/ZMConversation.m
```diff
@@ -121,28 +106,31 @@
 static NSString *const MessageDestructionTimeoutKey = @"messageDestructionTimeout";
 
 
-NSTimeInterval ZMConversationDefaultLastReadEventIDSaveDelay = 3.0;
+static NSString *const DownloadedMessageIDsDataKey = @"downloadedMessageIDs_data";
+static NSString *const LastEventIDDataKey = @"lastEventID_data";
+static NSString *const ClearedEventIDDataKey = @"clearedEventID_data";
+static NSString *const ArchivedEventIDDataKey = @"archivedEventID_data";
+static NSString *const LastReadEventIDDataKey = @"lastReadEventID_data";
```
marcoconti83: I think these lines should go

Source/Model/Conversation/ZMConversation.m
```diff
@@ -397,6 +378,11 @@ - (NSSet *)ignoredKeys;
             ZMConversationIsArchivedKey,
             ZMConversationIsSilencedKey,
             MessageDestructionTimeoutKey,
+            DownloadedMessageIDsDataKey,
+            LastEventIDDataKey,
+            ClearedEventIDDataKey,
+            ArchivedEventIDDataKey,
+            LastReadEventIDDataKey,
```
marcoconti83: I think these lines should go

ZMCDataModel.xcodeproj/project.pbxproj
```diff
@@ -2145,8 +2145,6 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
 				TARGETED_DEVICE_FAMILY = "1,2";
-				VERSIONING_SYSTEM = "apple-generic";
```
marcoconti83: Was this deletion on purpose (e.g. it was the default so no need to keep it in project-level too)?

Source/Model/Conversation/ZMConversation.m
basine: What if we have a message that still has modifications for those? If I remove these lines, those keys will still be in the keys tracked for local modifications.

Source/Model/Conversation/ZMConversation.m
basine: see comment below

Source/Model/Conversation/ZMConversation.m
marcoconti83: That's a good point, I have to think about it.

marcoconti83: As discussed, keeping the list of changed fields as a bitmask is not a viable solution anymore. Dropping a field in a later model version would mean that a bitmask that was non-null would not match the correct fields anymore. In order to update it, we would need to keep old, deprecated fields around, which we do not want.
We could migrate the bitmask from the old model to the new model, as we have access to both the old and the new. However, just knowing which field are in the model is not enough to parse the bitmask, we also need to know which keys were ignored at the moment when the bitmask was created. So we would need to keep a (potentially) ever-growing list of ignored past keys, which is also not a clean solution.
The most flexible solution would be to replace the bitmask with a concatenation of strings with the name of the changed keys. String parsing and comparison is extremely slow compared to the bitmask, however the number of entry that would have a non-nil list of string is very low (basically only 2-3 objects per user-initiated actions, and it will clear as soon as the network request is completed) so the performance impact should not be dramatic.
We will explore this in a separate PR.

PR #70, opened 2016-10-28
Use actual keys to track local modifications

basine:
**In this PR**
We are currently tracking local modifications through a bitfield that is updated whenever we save: We loop through all the entity's attributes which are sorted alphabetically and store a 0 or 1 indicating modification. However, this leads to problems when we remove properties or exclude keys that were previously tracked. Since the removed key is then missing from the entity's attributes or included in the ignored keys, the bitfield would now be shifted and keys that were not modified were assigned the wrong value which would essentially lead to a crash when syncing.

To circumvent this, we decided to store the modified keys as strings instead, so we can assign the stored value to a specific changed key with certainty. In this case if we remove keys from the database or exclude them from the keys to be tracked we can easily remove them from the modifiedKeys as well. 

Changes made:
* Remove modifiedDataFields property and replace it by modifiedKeys which stores a Set of Strings
* When saving we add the changed field names to the modifiedKeys
* In awakeFromFetch we update modifiedKeys subtracting keys that are no longer tracked or have been removed from the entity

ToDo before merging:
- [x] needs to be tested with zmessaging

### DIFF
--- a/Resources/zmessaging.xcdatamodeld/.xccurrentversion
+++ b/Resources/zmessaging.xcdatamodeld/.xccurrentversion
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>zmessaging2.21.0.xcdatamodel</string>
+	<string>zmessaging2.21.1.xcdatamodel</string>
 </dict>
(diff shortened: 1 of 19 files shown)


marcoconti83:
👍

PR #73, opened 2016-11-01
Add logic to receive /v3 assets

daehn:
# What's in this PR?

This adds the initial logic to support receiving assets send using the `/assets/v3` endpoint, this also refactors `ZMAssetClientMessage` internally to declutter the code from constant `self.version < 3` checks.

To do so the `imageMessageData` accessor now returns different implementations of this protocol depending on the version of the underlying asset. The two new added classes, `V2Asset` and `V3Asset` manage which cache and which cache key should be used to store the asset data and where to look for the image / file data etc.

There are still a couple of missing test for the receiving part that I am working on right now.

### DIFF
--- a/Source/Model/Message/ZMAssetClientMessage.m
+++ b/Source/Model/Message/ZMAssetClientMessage.m
@@ -324,6 +324,8 @@ static NSString * const AssociatedTaskIdentifierDataKey = @"associatedTaskIdenti
 - (BOOL)hasDownloadedImage
 {
     if(self.imageMessageData != nil || self.fileMessageData != nil) {
+        // TODO: Check if it's v3 and then use file cache
+
         return [self.managedObjectContext.zm_imageAssetCache assetData:self.nonce format:ZMImageFormatMedium encrypted:NO] != nil // processed or downloaded
         || [self.managedObjectContext.zm_imageAssetCache assetData:self.nonce format:ZMImageFormatOriginal encrypted:NO] != nil; // original
     }
@@ -332,6 +334,7 @@ static NSString * const AssociatedTaskIdentifierDataKey = @"associatedTaskIdenti
 
 - (BOOL)hasDownloadedFile
 {
+    // TODO: V3 If we receive an image as asset we might not have a filename
     return self.fileMessageData != nil && self.filename != nil &&
         [self.managedObjectContext.zm_fileAssetCache hasDataOnDisk:self.nonce fileName:self.filename encrypted:NO];
 }
--- a/Source/Model/Message/ZMAssetClientMessage.m
+++ b/Source/Model/Message/ZMAssetClientMessage.m
@@ -63,6 +63,9 @@ static NSString * const AssociatedTaskIdentifierDataKey = @"associatedTaskIdenti
 @property (nonatomic) ZMGenericMessage *cachedGenericAssetMessage;
 @property (nonatomic) int16_t version;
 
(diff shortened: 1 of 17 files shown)


Source/Model/Message/ZMAssetClientMessage.m
```diff
@@ -830,6 +871,13 @@ - (void)processingDidFinish
 
 - (NSData *)imageDataForFormat:(ZMImageFormat)format encrypted:(BOOL)encrypted
 {
+    // V3
```
daehn: This can be moved into the `V2Asset` / `V3Asset` objects as well.

Source/Model/Message/ZMAssetClientMessage.m
```diff
@@ -169,12 +191,29 @@ - (NSSet *)ignoredKeys
...
-    } else {
-        return nil;
+- (NSURL *)fileURL
+{
+    return self.asset.fileURL;
+}
+
+// Unfotunately we can not create the assets in -awakeFromInsert, as
+// the version is not yet set at that point.
+- (V2ImageAsset *)v2ImageAsset
```
marcoconti83: I am very weary of caches. Did you use a cache here because if it's V2, it's going to be V2 forever?

Source/Model/Message/ZMAssetClientMessage.m
```diff
@@ -286,9 +325,17 @@ + (NSSet *)keyPathsForValuesAffectingAssociatedTaskIdentifier
 
 - (id<ZMImageMessageData>)imageMessageData
 {
-    BOOL isImageMessage = self.mediumGenericMessage.imageAssetData != nil ||
-                          self.previewGenericMessage.imageAssetData != nil;
-    return isImageMessage ? self : nil;
+    // V2
```
marcoconti83: Could we also extract this to vXImageAsset?

Source/Model/Message/ZMAssetClientMessage.m
marcoconti83: Could we also extract this to vXImageAsset?

Source/Model/Message/ZMAssetClientMessage.m
daehn: This is more meant as lazy init of the underlying instance variable than as caching, If a message is v2 then it should always be v2. The version should only be set once, either when receiving updating with a generic message that contains the necessary information or when we insert the message ourselves.

The issue here was that we can only create the vXAssets after the version has been set, otherwise we might end up with the wrong asset, which is not guaranteed in awakeFromInsert for example. I'm also not happy with this approach and was thinking of either settings them in the version setter or wrapping this in a dispatch_once, as a v3 asset will always try to generate a v2 asset when accessed now.

Source/Model/Message/ZMAssetClientMessage.m
daehn: Yep, currently working on it.

Source/Model/Message/ZMAssetClientMessage.m
daehn: Yes, we should

daehn:
@marcoconti83 please re-review, I moved more logic into the v2 / v3 container, which also means they are no longer only conforming to `ZMImageMessageData` but also to a new protocol describing common methods (conforming them to `ZMImageAssetStorage` and `ZMFileMessageData` seemed excessive as a lot of the logic there is similar).

PR #74, opened 2016-11-03
Stop deletion timer of ephemeral message on receiving delete

basine:
**In this PR**
What happened: When the receiver of an ephemeral message was looking at the message on two devices at the same time, it is possible that we delete a message locally and receive the delete from the second device before creating the payload for the local delete. In this case we will end up in a crash loop.

To avoid this: When receiving a delete, we stop any timers for ephemeral messages.

### DIFF
--- a/Source/Model/Message/ZMMessage.m
+++ b/Source/Model/Message/ZMMessage.m
@@ -383,8 +383,28 @@ NSString * const ZMMessageIsObfuscatedKey = @"isObfuscated";
     if (nil != message && ![senderID isEqual:selfUser.remoteIdentifier] && !message.isEphemeral) {
         [conversation appendDeletedForEveryoneSystemMessageWithTimestamp:message.serverTimestamp sender:message.sender];
     }
-
+    // If we receive a delete for an ephemeral message that was not originally sent by the selfUser, we need to stop the deletion timer
+    if (nil != message && message.isEphemeral && ![message.sender.remoteIdentifier isEqual:selfUser.remoteIdentifier]) {
+        [self stopDeletionTimerForMessage:message];
+    }
     [message removeMessageClearingSender:YES];
+}
+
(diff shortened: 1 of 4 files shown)


Source/Model/Message/ZMMessage.m
```diff
@@ -383,10 +383,30 @@ + (void)removeMessageWithRemotelyDeletedMessage:(ZMMessageDelete *)deletedMessag
     if (nil != message && ![senderID isEqual:selfUser.remoteIdentifier] && !message.isEphemeral) {
         [conversation appendDeletedForEveryoneSystemMessageWithTimestamp:message.serverTimestamp sender:message.sender];
     }
-
+    // If we receive a delete for an ephemeral message that was not originally sent by the selfUser, we need to stop the deletion timer
+    if (nil != message && message.isEphemeral && ![message.sender.remoteIdentifier isEqual:selfUser.remoteIdentifier]) {
+        [self stopDeletionTimerForMessage:message];
+    }
     [message removeMessageClearingSender:YES];
```
marcoconti83: This (`message removeMessageClearingSender:`) will be executed synchronously after calling `stopDeletionTimerForMessage:`. However the `stopDeletionTimerForMessage:` will be executed async (because it does a `uiMOC performGroupedBlock`). Is it possible that we will have a race condition, where the timer still expires before we have a chance to cancel it, even if the message is already deleted? What happens in that case?

basine: Yes, in that case the second part of the PR jumps in - if the original sender is gone, we send a delete only to the selfuser's client. Alternatively we could iterate through the hidden messages to get the currently pending delete message, but I am not sure if this collection is ordered correctly so we can iterate e.g. through the last 10 messages.

marcoconti83: 👍

PR #78, opened 2016-11-08
Avoid crash when setting sender on updated message

basine:
**What's happening?**

We are currently experiencing a crash in `[ZMMessage updateWithTimestamp:senderUUID:eventID:forConversation:isUpdatingExistingMessage:]` 

> Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'Illegal attempt to establish a relationship 'sender' between objects in different contexts

It's impossible that the objects are really from different contexts since the sender is created  or fetched using the managedObjectContext of the message. It is very possible that the message is currently being deleted since this crash was introduced with deleting message. However, the flag isDeleted seems not to be returning `true` since we check for zombie objects in the method that calls this one and return early. 

To prevent further crashes, I added another check for zombie objects and for equality of contexts. This should prevent the crash.


marcoconti83:
When an object is deleted, all its getters will return `0`, `false`, `nil`. So `isDeleted` will actually return false. The `isDeleted` flag is supposed to be checked only in a specific situation (I think when merging contexes). In any other moment, it should not be used as it does not do what its name suggest :(

PR #79, opened 2016-11-09
Add crash info for crash when accessing MOC's userInfo

basine:
We have several crashes when accessing a value in the managedObjectContext's userInfo. Instead of returning the proper value, it returns an object of a different class. I added a helper method to log what exactly is currently contained in the MOC's userInfo.

### DIFF
--- a/Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
+++ b/Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
@@ -242,7 +242,18 @@ static BOOL storeIsReady = NO;
     if (value == nil) {
         return nil;
     }
-    RequireString([value isKindOfClass:class], "Value for key %s is not of class %s - userInfo contains: %s", key.UTF8String, NSStringFromClass(class).UTF8String, self.userInfo.debugDescription.UTF8String);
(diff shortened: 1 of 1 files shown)


Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -236,14 +236,24 @@ - (void)configureWithPersistentStoreCoordinator:(NSPersistentStoreCoordinator *)
     self.persistentStoreCoordinator = psc;
 }
 
+- (id)validUserInfoValueOfClass:(Class)class forKey:(NSString *)key;
+{
+    id value = self.userInfo[key];
+    if (value == nil) {
+        return nil;
+    }
+    RequireString([value isKindOfClass:class], "Value for key %s is not of class %s - userInfo contains: %s", key.UTF8String, NSStringFromClass(class).UTF8String, self.userInfo.debugDescription.UTF8String);
```
marcoconti83: We saw from the crash logs that this often returns a dictionary. 
If it is a dictionary, where does it come from? Is it from our code, or some internal CoreData stuff? I would add an additional check: if it's a dictionary, print the keys, this will help us identify where it comes from.

PR #80, opened 2016-11-10
Add support for receiving assets sent using the /assets/v3 endpoint

daehn:
# What's in this PR?

* This PR adds a new model version which introduces a new `version` field on `ZMAssetClientMessage`. This integer is set to 3 when we receive an asset message that was sent using the `/assets/v3` endpoint, which we determine by inspecting the protobuf, if we have an asset ID in the `Asset.Uploaded` or in the `Asset.Preview` (before modifying it ourselves).

* To make the reasoning about `v2` / `v3` differences easier the different implementations have been extracted in the new classes `V2Asset` and `V3Asset`, the `imageMessageData` accessor now returns different implementations of this protocol depending on the version of the underlying asset.
The two  classes manage which cache and which cache key should be used to store the asset data and where to look for the image / file data etc.

### Still to be done:

- [x] Migration test for the new data model version.


Source/Model/Message/ZMAssetClientMessage.m
```diff
@@ -82,7 +88,10 @@ @implementation ZMAssetClientMessage
 @dynamic transferState;
 @dynamic progress;
 @dynamic associatedTaskIdentifier_data;
+
 @synthesize cachedGenericAssetMessage;
+@synthesize v2Asset = _v2Asset;
```
marcoconti83: Didn't we say we were gonna get rid of the cache?

Source/Model/Message/ZMAssetClientMessage.m
```diff
@@ -126,6 +144,11 @@ + (instancetype)assetClientMessageWithFileMetadata:(ZMFileMetadata *)metadata no
     return message;
 }
 
+- (id <AssetProxyType>)asset
+{
+    return self.v2Asset ?: self.v3Asset;
```
marcoconti83: 👍

Source/Model/Message/ZMAssetClientMessage.m
```diff
@@ -1045,6 +976,11 @@ - (BOOL)isAudio
     return [self.mimeType isAudioMimeType];
 }
 
+- (BOOL)isImage
+{
+    return self.genericAssetMessage.v3_isImage;
```
marcoconti83: Why do we check only the v3 field? What if this `ZMAssetClientMessage` is a v2? `isImage` is public so someone WILL call it at some point on a v2

Source/Model/Message/ZMAssetClientMessage.m
daehn: Yes, forgot to remove this.

Source/Model/Message/ZMAssetClientMessage.m
daehn: Good point, I guess it should not be public, I will check if it is still needed at all as we now return a valid `ZMImageMessageData` when we have a v3 message.

marcoconti83: As an alternative call it `isV3Image`?

daehn: I was able to completely remove this flag as we can infer it from other properties already

daehn:
I added some changes that were made on the `feature/sending-v3-images` branch before, mainly storing images and file preview images in the image cache to avoid invalidation caused by large files.

daehn:

Source/Model/Message/ZMAssetClientMessage+Download.swift
```diff
@@ -27,18 +27,7 @@ extension ZMAssetClientMessage {
     public static var ImageDownloadNotificationName : String {
         return "ZMAssetClientMessageImageDownloadNotification"
     }
-    
-    public override func requestImageDownload() {
```
marcoconti83: 👍

marcoconti83:
So we are waiting for the migration tests on this?

daehn:
@marcoconti83 depends, I am working on it now, but I can also make a separate PR for it.

daehn:
@marcoconti83 migration test has been added ✅

PR #83, opened 2016-11-14
Share extension support

typfel:
## Purpose of this PR
In order to support running inside a share extension we need to do the following changes:
- The database is currently created in a folder named after the bundle identifier. This won't work inside an extension since it has its own bundle identifier. 
<b>Solution:</b> add a parameter for the name of the database.

- Caches which store images/files which have been downloaded or are about to be uploaded are currently stored in their own container's cache. In order for a request to transition from the an extension to host app the caches must live in the shared container. 
<b>Solution:</b> Add a parameter for specifying the location of a cache on disk.

## Refactoring

As pointed in the code review, the static variables in `NSManagedObjectContext` were getting out of hand. There the code related to moving the location of the database has been extracted out to the `PersistentStoreRelocator`.


Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.h
```diff
@@ -34,31 +34,34 @@
...
 
 /// Checks if migration is needed or the database has to be moved
-+ (BOOL)needsToPrepareLocalStoreInDirectory:(NSURL *)databaseDirectory;
++ (BOOL)needsToPrepareLocalStoreInDirectory:(NSURL *)databaseDirectory identifier:(NSString *)databaseIdentifier;
 
 /// Creates persistent store coordinator and migrates store if needed
 /// @param sync defines if the method should execute sycnhronously or not (ususally it makes sence to execute it
 ///         synchronously when @c +needsToPrepareLocalStore is YES)
+/// @param databaseDirectory where database should we created / moved to
+/// @param databaseIdentifier identifier of the database. Has historally been the bundle identifier.
```
marcoconti83: `s/historally/historically`

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -47,6 +47,7 @@
 
 static BOOL UsesInMemoryStore;
 static NSURL * DatabaseDirectoryURL;
+static NSString * DatabaseIdentifier;
```
marcoconti83: I think we should move away from global static variables to store paths.
 It seems like we pass the identifier to all `create...contextWithStoreDirectory` and `needsToPrepareLocalStoreInDirectory`, so probably we don't need to keep the folder/identifier in a global. Or is there another reason why we keep them?

typfel: No there's no reason to keep them afaik. Changing it would be a fairly big re-factoring of the database creation/migration code. Need to add these parameters to a ton of methods or re-factor them.

Source/ManagedObjectContext/NSManagedObjectContext+zmessaging.m
```diff
@@ -81,28 +78,33 @@ @implementation NSManagedObjectContext (zmessaging)
 
 static BOOL storeIsReady = NO;
 
-+ (BOOL)needsToPrepareLocalStoreInDirectory:(NSURL *)databaseDirectory identifier:(NSString *)databaseIdentifier
+//+ (NSURL *)storeURLInDirectory:(NSSearchPathDirectory)directory;
```
daehn: Can be removed?

typfel: yes

daehn:
@typfel is this one ready to be merged?

typfel:
Well if no one needs to touch the data model for the next release.

marcoconti83:
FYI I created data model 2.24 already

mikeger:
@marcoconti83 do you approve this one now? Should it be merged?

marcoconti83:
But we did touch the data model. What does it mean, Jacob?
I bumped data model version in latest version.

Source/Utilis/UserImageLocalCache.swift
```diff
@@ -73,10 +73,23 @@ extension NSManagedObjectContext
     /// Cache for small user profile image
     fileprivate let smallUserImageCache : PINCache
     
-    public override init() {
-        largeUserImageCache = PINCache(name: "largeUserImages")
+    
+    /// Create UserImageLocalCache
+    /// - parameter location: where cache is persisted on disk. Defaults to caches directory if nil.
```
Einh06: Can we specify where does that store the caches if the location is nil? I think that'd be helpful to know the default directory.

typfel: It says it stores it in the caches directory if nil. We could say `NSCachesDirectory` if that makes it more clear.

typfel:
@marcoconti83 what are you concerned about? this PR doesn't modify the model version.

Source/Utilis/UserImageLocalCache.swift
Einh06: My bad, i misread.

PR #89, opened 2016-11-25
Support V2 and V3 calling simultaneously

typfel:
First attempt att supporting calling V2 & V3 simultaneously. I plan to re-factor the calling later. This is just to get it running.

The old `ZMVoiceChannel` has been put inside a `VoiceChannelRouter` together with the new `VoiceChannelV3`. The router then decide which voice channel it should forward any and requests `join`, `leave` depending on which is active.

The `WireCallCenter` is moved here since we need it for the VoiceChannel observers.


Source/VoiceChannel/ZMVoiceChannel.m
```diff
@@ -116,19 +116,24 @@ - (ZMConversation *)firstOtherConversationWithActiveCall
...
-    const BOOL selfJoined = [self.callParticipants containsObject:selfUser];
-    const BOOL otherJoined = otherUser != nil;
-    const BOOL isDeviceActive = self.callDeviceIsActive;
-    const BOOL flowActive = self.isFlowActive;
+//    VerifyReturnValue(self.managedObjectContext != nil, ZMVoiceChannelStateInvalid);
+    
+//    ZMUser *selfUser = [ZMUser selfUserInContext:self.managedObjectContext];
+//    ZMUser *otherUser = [self.callParticipants.array firstObjectMatchingWithBlock:^BOOL(ZMUser *user) {
+//        return user != selfUser;
+//    }];
```
mikeger: Commented code. Todo?

Source/VoiceChannel/WireCallCenter.swift
```diff
@@ -0,0 +1,255 @@
...
+        super.init()
+        
+        let observer = Unmanaged.passUnretained(self).toOpaque()
+        
+        let resultValue = wcall_init(
+            (userId as NSString).utf8String,
+            (clientId as NSString).utf8String,
+            { (version, context) in
+                if let context = context {
+                    _ = Unmanaged<WireCallCenter>.fromOpaque(context).takeUnretainedValue()
```
typfel: Right, Florian had some issues with `takeRetainedValue` but I should take another look at this.

Source/VoiceChannel/ZMVoiceChannel.m
typfel: No I should delete this

Source/VoiceChannel/WireCallCenter.swift
typfel: Question is, we are basically setting ourself as a `delegate` to the C api callbacks. In order to avoid retain cycles this should be `weak`/`unretained`. What's missing I guess is a way to nil out the "delegate" when `WireCallCenter`gets deallocated.

mikeger: I think C API does not know about the ownership, so what we pass in is an unowned reference. We have a contract with this API that this reference would be valid during the lifetime of the C object. When `WireCallCenter` is deallocated we should destroy the C stuff referencing it.

typfel: Exactly. We currently setup up the observing in `wcall_init` but we are missing a corresponding `wcall_deinit` to call in `deinit` to stop the callbacks from being called on the deallocated object.

typfel: Ah so turn out there's a `wcall_close()` function. Calling it from `deinit` now.

Resources/zmessaging.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>zmessaging2.21.2.xcdatamodel</string>
+	<string>zmessaging2.24.1.xcdatamodel</string>
```
daehn: Looks like this comes from the merged in changes, maybe you can rebase you changes onto of the current develop

typfel: I've merged develop into this branch. Which makes the diff useless when compared to `feature/calling-v3`

typfel: I now compare against `feature/calling-v3` which is branched directly from develop

daehn:
Can we kick of a new release for this one?

typfel:
What do you mean? This an intermediate PR

PR #91, opened 2016-11-29
Create entity to store address book reference

marcoconti83:

We need to map users to a local address book entry, to be able to display how a user is named in your address book. Searching for contacts in your AB with the same email as the user, like we do now, won't scale as we would need to iterate the entire AB and soon we won't have access to the email of connected users anymore.

However if the store the local address book entry reference ID when the BE informs us of a match during the upload, the look up should be fast.


Source/Model/AddressBookEntry.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public class AddressBookEntry : ZMManagedObject {
+    
+    enum Fields : String {
```
daehn: If this is only used from Swift it might be possible to used the `#keyPath` syntax instead

Source/Model/AddressBookEntry.swift
```diff
@@ -0,0 +1,46 @@
...
+    public override func keysTrackedForLocalModifications() -> Set<String> {
+        return []
+    }
+    
+    public override static func entityName() -> String {
+        return "AddressBookEntry"
+    }
+    
+    public override static func sortKey() -> String? {
+        return Fields.idIOS10.rawValue
```
mikeger: Should it fork here, i.e. if iOS9 then return idIOS9?

Source/Model/AddressBookEntry.swift
basine: :-o

daehn: Just checked and the properties have to be marked as `@objc` unfortunately to do this:

```swift
class Person {
    @objc let age: Int = 0
}

#keyPath(Person.age) // "age"
```

Source/Model/AddressBookEntry.swift
marcoconti83: I had to write something to make the default implementation happy, but we will actually never fetch them so it doesn't need to sort.

marcoconti83:
Actually the Contacts framework is available starting from iOS 9, not 10. I'm updating the PR.

mikeger:
@marcoconti83 then we can potentially ignore iOS 8 users in the context of that feature, i.e. on iOS 8 we would not show the address book matches to simplify the code?

marcoconti83:
Numbers of iOS 8 users are relatively low, but it's not my call.

jenskasemets:
> Numbers of iOS 8 users are relatively low, but it's not my call.

Currently there are ca 2% active users on iOS 8, declining all the time. Let's skip implementation of only-iOS8 code as it adds complexity (development, testing etc) to non-released feature. Once this feature is released and we see that there are users missing it (feature which currently don’t have anyway), we can always reconsider.

marcoconti83:
@jenskasemets  actually users have this feature now (who's who in your address book) but it's limited to the first 1000 contacts due to poor performances. 
With this PR, we are working towards removing the limit. But if we don't do the iOS 8 part, user with iOS 8 will lose the current feature.

jenskasemets:
@marcoconti83 ok, still fine to drop iOS 8 support for this now

PR #96, opened 2016-12-14
Add mock of WireCallCenter

typfel:
We need a mock of the `WireCallCenter` since we can't make calls to AVS in our tests.

### DIFF
--- a/Source/VoiceChannel/WireCallCenter.swift
+++ b/Source/VoiceChannel/WireCallCenter.swift
@@ -182,7 +182,7 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
         wcall_close()
     }
     
-    public required init(userId: UUID, clientId: String, dontRegisterObservers : Bool = false) {
+    public required init(userId: UUID, clientId: String, registerObservers : Bool = true) {
         self.userId = userId
         
         super.init()
@@ -191,7 +191,7 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
             fatal("Only one WireCallCenter can be instantiated")
         }
         
-        if (dontRegisterObservers) {
+        if (registerObservers) {
             
             let observer = Unmanaged.passUnretained(self).toOpaque()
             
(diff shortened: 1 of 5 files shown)


Source/VoiceChannel/WireCallCenter.swift
```diff
@@ -191,113 +191,116 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
...
+                { (version, context) in
+                    if let context = context {
+                        _ = Unmanaged<WireCallCenter>.fromOpaque(context).takeUnretainedValue()
+                        
+                        
+                    }
+                },
+                { (token, conversationId, userId, clientId, data, dataLength, context) in
+                    guard let token = token, let context = context, let conversationId = conversationId, let userId = userId, let clientId = clientId, let data = data else {
+                        print("BAD callback")
```
mikeger: Should it be `ZMLogError` ? (this comment applies to some more print statements in this file)

Source/VoiceChannel/WireCallCenter.swift
```diff
@@ -191,113 +191,116 @@ private typealias WireCallMessageToken = UnsafeMutableRawPointer
...
+                },
+                { (conversationId, userId, isVideoCall, context) -> Void in
+                    guard let context = context, let conversationId = conversationId, let userId = userId else {
+                        print("BAD callback")
+                        return
+                    }
+                    
+                    let selfReference = Unmanaged<WireCallCenter>.fromOpaque(context).takeUnretainedValue()
+                    
+                    print("entering incoming")
```
mikeger: I am not sure if `print` is still necessary here? (this comment applies to some more print statements in this file)

Source/VoiceChannel/WireCallCenter.swift
typfel: This will be deleted later and I can't use any references to the outside the block inside a C-block

Source/VoiceChannel/WireCallCenter.swift
typfel: Will also be deleted

typfel:
@mikeger deleted the print statements

PR #97, opened 2016-12-15
Feature/categorization in core data

marcoconti83:
We want to cache the category of a message so that we don't have to access the protocol buffer encoded properties every time we want to know the category of a message

# Changes
Created a new Core Data version


marcoconti83:
The PR has been modified to accept a list of categories. 

Since creating an array of OptionSets is ambiguous (it automatically translate to one single OptionSet that includes all flags from individual OptionSets), the API accepts a Set instead of an array.

marcoconti83:
@basine

PR #102, opened 2016-12-16
Collections

basine:
**In this PR**
I added two classes with the same functionality - one of which we will remove again after testing performance on the device
`AssetCollection` - fetches the messages in batches (with multiple requests) and analyzes them
`AssetCollectionBatched` - fetches the messages with one request with `fetchBatchSize` set which theoretically evaluates all objects, but only brings the specified batchSize into memory. When iterating through all objects, it refaults the ones that are not used anymore and brings new ones into memory.

Both are initialized with a conversation, a set of categories to group by and a set of categories to exclude (e.g. one may want all files, but no videos) as well as a delegate.

The delegate is called for every batch with the new elements fetched:
    `func assetCollectionDidFetch(messages: [MessageCategory: [ZMMessage]], hasMore: Bool)`
as well as when all elements have been fetched & analyzed.
    `func assetCollectionDidFinishFetching(result : AssetFetchResult)`

To get all assets of a category call `assets(for category: MessageCategory)`
To cancel the process, call `tearDown`.


Source/Model/Conversation/AssetCollection.swift
```diff
@@ -0,0 +1,241 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum AssetFetchResult : Int {
+    case success, failed, noAssetsToFetch
+}
+
+public protocol ZMCollection : NSObjectProtocol {
```
mikeger: I think this one should include `var delegate : AssetCollectionDelegate` ?

Source/Model/Conversation/AssetCollection.swift
```diff
@@ -0,0 +1,241 @@
...
+    func tearDown()
+    func assets(for category: MessageCategory) -> [ZMMessage]
+}
+
+public protocol AssetCollectionDelegate : NSObjectProtocol {
+    /// The AssetCollection calls this when the fetching completes
+    /// To get all messages for any category defined in `including`, call `assets(for category: MessageCategory)`
+    func assetCollectionDidFetch(messages: [MessageCategory: [ZMMessage]], hasMore: Bool)
```
mikeger: So calling `assets(for category: MessageCategory)` starts fetching the assets, and later the delegate would get calls to `assetCollectionDidFetch`? Why it returns array of `ZMMessage` synchronously? Not sure.

Source/Model/Conversation/AssetCollection.swift
```diff
@@ -0,0 +1,241 @@
...
+
+
+public class AssetCollection : NSObject, ZMCollection {
+
+    private unowned var delegate : AssetCollectionDelegate
+    private var assets : CategorizedFetchResult?
+    private var lastAssetMessage : ZMAssetClientMessage?
+    private var lastClientMessage : ZMClientMessage?
+    private let conversation: ZMConversation
+    private let including : [MessageCategory]
```
mikeger: I thought `MessageCategory` is OptionSet, so it's not necessary to have an array of those?

Source/Model/Conversation/AssetCollection.swift
mikeger: I see, `init` starts fetching messages immediately

basine: Yes, the init starts the fetch. `assets(for category: MessageCategory)` is just a method to retrieve the ones that were fetched in the meantime (e.g. when you are switching detail screens from images to link previews)

Source/Model/Conversation/AssetCollection.swift
basine: Yeah initially I wanted to include the object in the delegate callback but that made all my tests crash ... because Swift

Source/Model/Conversation/AssetCollection.swift
```diff
@@ -0,0 +1,251 @@
...
+public enum AssetFetchResult : Int {
+    case success, failed, noAssetsToFetch
+}
+
+public protocol ZMCollection : NSObjectProtocol {
+    func tearDown()
+    func assets(for category: MessageCategory) -> [ZMMessage]
+}
+
+public protocol AssetCollectionDelegate : NSObjectProtocol {
```
marcoconti83: When you say `AssetCollection...`, we mean any collection, not just those with assets, right? (e.g. text links, ...)

Source/Model/Conversation/AssetCollection.swift
basine: It is an OptionSet, but you want to get sorted results, don't you? Not sure what your question is.

mikeger:
Ok, looks good to me, @marcoconti83 please review

Source/Model/Conversation/AssetCollection.swift
```diff
@@ -0,0 +1,251 @@
...
+    /// The AssetCollection calls this when the fetching completes
+    /// To get all messages for any category defined in `including`, call `assets(for category: MessageCategory)`
+    func assetCollectionDidFetch(messages: [MessageCategory: [ZMMessage]], hasMore: Bool)
+    
+    /// This method is called when all assets in the conversation have been fetched & analyzed / categorized
+    func assetCollectionDidFinishFetching(result : AssetFetchResult)
+}
+
+
+public class AssetCollection : NSObject, ZMCollection {
```
marcoconti83: Could you please add a comment to the class about what it responsibilities are? Something like `A collection of assets (?) for a given conversation, that are fetched asynchronously, using multiple fetches`

Source/Model/Conversation/AssetCollectionBatched.swift
```diff
@@ -0,0 +1,275 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+public class AssetCollectionBatched : NSObject, ZMCollection {
```
marcoconti83: Could you please add a comment to the class about what it responsibilities are? Something like `A collection of assets (?) for a given conversation, that are fetched asynchronously, using a single batch fetch`

Source/Model/Conversation/AssetCollection.swift
basine: Yes.

PR #104, opened 2016-12-19
Set excluding categoy more granularly

basine:
Any idea which hashValue to use? maybe just the including category?


Source/Model/Conversation/AssetCollectionBatched.swift
```diff
@@ -17,16 +17,32 @@
 //
 
 
+public struct CategoryMatch : Hashable {
+    public let including: MessageCategory
+    public let excluding: MessageCategory
+    
+    public var hashValue: Int {
+        // TODO Sabine: this is a baaaad hashValue
```
marcoconti83: usually bitwise XOR is a good choice for this: `excluding.hashValue ^ including.hashValue`.

PR #113, opened 2016-12-21
Fix core data concurrency issue

marcoconti83:


Source/Model/Conversation/AssetCollection.swift
```diff
@@ -84,9 +84,10 @@ public class AssetCollection : NSObject, ZMCollection {
         self.matchingCategories = matchingCategories
         super.init()
         
+        let syncMoc = self.syncMOC
         syncMOC?.performGroupedBlock {
             guard !self.tornDown else { return }
-            guard let syncConversation = (try? self.syncMOC?.existingObject(with: self.conversation.objectID)) as? ZMConversation else {
+            guard let syncConversation = (try? syncMoc!.existingObject(with: self.conversation.objectID)) as? ZMConversation else {
```
typfel: Why does this lead to a concurrency issue?

marcoconti83: `self.syncMOC` is a computed property, defined like:
```
...
self.conversation.managedObjectContext?.zm_syncManagedObjectContext
...
```
because `self.conversation` is a managed object on the UI, and the `zm_syncManagedObjectContext` computed property requires accessing the user info, this is a threading violation according to Core Data. Sadly this does not fix any specific crash.

PR #120, opened 2016-12-27
Explicit types to be store in metadata

marcoconti83:
With the current API, it is possible to store anything in the persistent store metadata, even instances of objects that can't be serialized and will therefore cause the managed object context to fail to save


Source/ManagedObjectContext/NSPersistentStore+Metadata.swift
```diff
@@ -0,0 +1,146 @@
...
+    /// Non-persisted store metadata
+    fileprivate var nonCommittedMetadata : [String: Any] {
+        get {
+            return (self.userInfo[metadataKey] as? NSDictionary) as? [String: Any] ?? [:]
+        }
+    }
+    
+    /// Non-persisted deleted metadata (need to keep around to know what to remove
+    /// from the store when persisting)
+    fileprivate var nonCommitedDeletedMetadataKeys : Set<String> {
```
typfel: Typo: comited > comitted

Source/ManagedObjectContext/NSPersistentStore+Metadata.swift
```diff
@@ -0,0 +1,146 @@
...
+        self.nonCommittedMetadata.forEach { (key: String, value: Any) in
+            storedMetadata[key] = value
+        }
+        
+        self.persistentStoreCoordinator?.setMetadata(storedMetadata, for: store)
+        self.discardNonCommitedMetadata()
+    }
+    
+    /// Remove key from keys that are deleted
+    fileprivate func removeFromNonCommittedDeteledMetadataKeys(key: String) {
```
typfel: I would change this function name to be consistent with the 'add' function. Maybe 'removeNonComittedMetadataKey'

Source/ManagedObjectContext/PersistableInMetadata.swift
```diff
@@ -0,0 +1,20 @@
+//
```
typfel: 😱

Source/ManagedObjectContext/NSPersistentStore+Metadata.swift
marcoconti83: Semantically, `removeNonComittedMetadataKey` is wrong. The key might already be in the metadata on disk, I just deleted it in memory, and I want to revert the fact that it is deleted in memory, so that when I persist everything, I don't actually delete it from disk too.

marcoconti83: I'm updating the comment to explain this better

PR #125, opened 2017-01-11
Add option to append a message without fetching link preview

daehn:
# What's in this PR?

* At the moment we do not want to fetch link previews when sending messages through the iOS share extension (this might change at some point). 
* As the share extension does not fetch the previews itself this currently leads to the previews being fetched once the main app is opened, which is undesired. 
* This PR adds a parameter to specify if a link preview should be fetched when appending a text message.


marcoconti83:
@daehn Please make a release

PR #129, opened 2017-01-13
Add modified conversation list to store touched conversations from an extension context

daehn:
# What's in this PR?

In the iOS Share Extension, we set up a separate Core Data stack that uses the same persistent store as the main app, but has it's own two contexts. When sending a message the main app will not receive a change notification (as it uses different contexts). The change will be in the persistent store on disk but will not be propagated, which is why we now keep a list of touched conversations, manually turn them into faults and fire a notification for their last message with observer system in order to update the message window and notify the UI.

### DIFF
--- a/Source/Model/Conversation/SharedModifiedConversationsList.swift
+++ b/Source/Model/Conversation/SharedModifiedConversationsList.swift
@@ -65,6 +65,8 @@ public extension NSManagedObjectContext {
 
         if zm_isUserInterfaceContext {
             conversations?.forEach {
+                // When notifying the last message changed, the message window will be
+                // recalculated and a notification about a potentially added message will be fired.
                 guard let message = $0.messages.lastObject as? ZMMessage else { return }
                 globalManagedObjectContextObserver.notifyNonCoreDataChangeInManagedObject(message)
             }
--- a/Source/Model/Conversation/SharedModifiedConversationsList.swift
+++ b/Source/Model/Conversation/SharedModifiedConversationsList.swift
@@ -1,0 +1,52 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


Source/Model/Conversation/SharedModifiedConversationsList.swift
```diff
@@ -0,0 +1,74 @@
...
+        let conversations = ZMConversation.fetchObjects(withRemoteIdentifiers: NSOrderedSet(set: identifiers), in: self)?.array as? [ZMConversation]
+
+        conversations?.forEach { conversation in
+            refresh(conversation, mergeChanges: true)
+        }
+
+        if zm_isUserInterfaceContext {
+            conversations?.forEach {
+                guard let message = $0.messages.lastObject as? ZMMessage else { return }
+                globalManagedObjectContextObserver.notifyNonCoreDataChangeInManagedObject(message)
```
marcoconti83: I would add a comment here, saying that we only notify about the last message as this is enough to make the message windows to recalculate

Source/Model/Conversation/SharedModifiedConversationsList.swift
```diff
@@ -0,0 +1,74 @@
...
+
+
+private let zmLog = ZMSLog(tag: "modified conversations")
+
+
+/// This class is used to mark conversations as modified in an extension 
+/// context in order to refetch them in the main application.
+@objc public class SharedModifiedConversationsList: NSObject {
+
+    private let defaults = UserDefaults.shared()
```
marcoconti83: Q: why not store in the persistent store metadata though? we normally use that for this kind of data

daehn: Good idea!

daehn: @marcoconti83 I was unable to get the added identifiers when in the main app immediately after adding them in the extension. I am unsure why and would suggest to change and investigate this at a later point.

daehn:

PR #131, opened 2017-01-17
Collections: do not show gifs in the list

vytis:
## What does it fix?

In the collections we shouldn't show GIFs to save download data, but it was showing up for the sender.

## What's in this PR?

Message category is cached, but was getting out of sync and didn't have the information that we can get after processing the image. In this case the image message category was not updated to reflect that it's a GIF.

### DIFF
--- a/Tests/Source/Model/Messages/ZMMessageCategorizationTests.swift
+++ b/Tests/Source/Model/Messages/ZMMessageCategorizationTests.swift
@@ -21,6 +21,14 @@ import ZMUtilities
 import ZMCDataModel
 import ZMCLinkPreview
 
+extension ZMConversationMessage {
+    var categorization : MessageCategory {
+        guard let message = self as? ZMMessage else {
+            return .none
+        }
+        return message.category
+    }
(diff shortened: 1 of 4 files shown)


Tests/Source/Model/Messages/ZMMessageCategorizationTests.swift
```diff
@@ -110,7 +118,21 @@ class ZMMessageCategorizationTests : ZMBaseManagedObjectTest {
...
     }
-    
+
+    func testThatItUpdatesCachedCategoryAfterSettingFullImageData() {
+        // GIVEN
+        let conversation = ZMConversation(remoteID: .create(), createIfNeeded: true, in: syncMOC)!
+        let data = self.data(forResource: "animated", extension: "gif")!
+        conversation.appendMessage(withImageData: data)
+
+        let message = conversation.messages.lastObject as! ZMAssetClientMessage
```
marcoconti83: Missing a `// WHEN`, and I would add a sanity check

PR #132, opened 2017-01-17
Make sure security warning when a conversation degraded is displayed once only

marcoconti83:
When a conversation degrades from verified to not verified with a message, if the app is killed and restarted, the takeover warning of the degradation is displayed again at every restart.

# Changes
- Do not display the takeover if the conversation snapshot had no security level (e.g. there was no snapshot). It means that we just restarted the app. The message will still not be sent anyway (it is already marked as "expired") in any case.
- Remove magic string

### DIFF
--- a/Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
+++ b/Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
@@ -27,7 +27,7 @@ public protocol ZMGeneralConversationObserver {
 extension ZMConversation : ObjectInSnapshot {
     
     public var observableKeys : [String] {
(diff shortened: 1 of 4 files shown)


Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
```diff
@@ -265,7 +265,8 @@ extension ConversationChangeInfo {
     /// might have added a participant or renamed the conversation (causing a
     /// system message to be inserted)
     fileprivate var recentNewClientsSystemMessageWithExpiredMessages : ZMSystemMessage? {
-        if(!self.securityLevelChanged || self.conversation.securityLevel != .secureWithIgnored) {
+        let previousSecurityLevel = (self.previousValueForKey(SecurityLevelKey) as? NSNumber).flatMap { ZMConversationSecurityLevel(rawValue: $0.int16Value) }
```
typfel: Didn't you say you would re-write the searching backwards code?

marcoconti83: Yes, I will eventually, if I don't manage, this at least fixed one bug.

PR #135, opened 2017-01-18
Convert code to handle degradation of conversations to swift

marcoconti83:
I need to make a few changes here to fix bugs in share extension. This is the first PR that just ports the class to Swift, the actual fix will be in the next PRs, but if I create a single PR it will be impossible to see what changed, hence this one is submitted first.


Source/Model/Conversation/ZMConversation+OTR.swift
```diff
@@ -0,0 +1,330 @@
...
+        let clients = client.flatMap { Set(arrayLiteral: $0) } ?? Set<UserClient>()
+        let serverTimestamp = date ?? self.timestamp(after: self.messages.lastObject as? ZMMessage)
+        _ = self.appendSystemMessage(type: type,
+                                     sender: sender,
+                                     users: nil,
+                                     clients: clients,
+                                     timestamp: serverTimestamp)
+    }
+
+    public func appendDeletedForEveryoneSystemMessage(at date: Date, sender: ZMUser) {
```
mikeger: Seems not OTR-related, can be moved to another category? (also `expireAllPendingMessages`)

Source/Model/Conversation/ZMConversation+OTR.swift
```diff
@@ -0,0 +1,330 @@
...
+        let index = self.sortedAppendMessage(systemMessage)
+        systemMessage.visibleInConversation = self
+        return (message: systemMessage, insertionIndex: index)
+    }
+    
+
+    
+    /// Returns a timestamp that is shortly (as short as possible) before the given message,
+    /// or the last modified date if the message is nil
+    fileprivate func timestamp(before: ZMMessage?) -> Date? {
```
mikeger: I think it should be an extension on `Date`.

Source/Model/Conversation/ZMConversation+OTR.swift
```diff
@@ -0,0 +1,330 @@
...
+        return timestamp.addingTimeInterval(0.001)
+    }
+
+}
+
+// MARK: - Conversation participants status
+extension ZMConversation {
+    
+    /// Returns true if all participants are connected to the self user and all participants are trusted
+    public var trusted : Bool {
```
mikeger: Should it return false for empty conversations?

Source/Model/Conversation/ZMConversation+OTR.swift
```diff
@@ -0,0 +1,330 @@
...
+    /// If true the conversation might still be trusted / ignored
+    public var hasUntrustedClients : Bool {
+        return (self.activeParticipants.array as! [ZMUser]).first { $0.untrusted() } != nil
+    }
+}
+
+// MARK: - Notifications
+extension ZMConversation {
+    
+    func notifyOnUI(notification: String) {
```
mikeger: Does not seem to belong here, looks like a general utility method.

mikeger:
@marcoconti83 up to you if you want to change the comments here or if you planned to change the structure in the upcoming PR.

marcoconti83:
I will fix in the next PR

Source/Model/Conversation/ZMConversation+OTR.swift
marcoconti83: Will move out

Source/Model/Conversation/ZMConversation+OTR.swift
marcoconti83: Will move out. 
`expireAllPendingMessage` though belongs here, as it's about blocking all further messages when sending them would be a security concern (as a result of a conversation degradation).

Source/Model/Conversation/ZMConversation+OTR.swift
marcoconti83: I need the `self.lastModifiedDate` of the conversation in this function, it could be an extension of `Date` but then I would need to pass in the conversation anyway. 
This is about returning a timestamp to use to insert in the conversation if I want to insert a new message after the given message one, or if none is given, at the end of the conversation. It's tightly linked to the conversation itself.

Source/Model/Conversation/ZMConversation+OTR.swift
marcoconti83: Good point

PR #136, opened 2017-01-19
Better message expiration/resend in case of security degradation

marcoconti83:
The detection of which messages caused the conversation to degrade the security level was a bit hacky, and it would case a few issues (takeover displayed at every restart, resending of all expired messages even if they were expired for other reasons).
This PR make it possible to resend/cancel only those messages that caused the conversation to degrade.

# Changes
- Also fixed test subclassing (it was causing the same tests to be run many times)
- Also extracted methods not related to conversation degradation, as mentioned in https://github.com/wireapp/wire-ios-data-model/pull/135

### DIFF
--- a/Source/Model/Conversation/ZMConversation+OTR.swift
+++ b/Source/Model/Conversation/ZMConversation+OTR.swift
@@ -51,9 +51,7 @@ extension ZMConversation {
         guard self.decreaseSecurityLevelIfNeeded() else { return }
         self.appendNewAddedClientSystemMessage(added: clients, before: message)
         if let message = message {
-            if message.deliveryState != .sent && message.deliveryState != .delivered {
-                self.expireAllPendingMessages(staringFrom: message)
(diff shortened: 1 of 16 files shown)


Source/Model/Conversation/ZMConversation+OTR.swift
```diff
@@ -121,51 +119,78 @@ extension ZMConversation {
...
+            }
+            
+            if message.isConversationNotVerifiedSystemMessage {
+                stop.initialize(to: true)
+            }
+        }
+    }
+    
+    /// Reset those that caused degradation
+    public func doNotResendMessagesThatCausedDegradation() {
```
daehn: `doNotResendMessagesThatCausedDegradation` and  `resendMessagesThatCausedConversationSecurityDegradation` look the same expect the additional `resend` call, maybe the common logic can be extracted.

PR #145, opened 2017-01-23
Add appendMissedCallMessage method

typfel:


Source/Model/Conversation/ZMConversation+Calling.swift
```diff
@@ -0,0 +1,27 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public extension ZMConversation {
+    
+    public func appendMissedCallMessage(fromUser user: ZMUser, at timestamp: Date) {
+        _ = appendSystemMessage(type: .missedCall, sender: user, users: Set<ZMUser>([user]), clients: nil, timestamp: timestamp)
```
daehn: We could mark `appendSystemMessage` as `@discardableResult`

typfel: Good idea

PR #146, opened 2017-01-24
Longer client session identifiers

marcoconti83:
Adopting the new Cryptobox API for client sessions identifiers

# Testing
I'm not testing session migration (old identifiers to new identifiers) here. I tested it at the wire-ios-cryptobox level, and I plan to make an integration test at the wire-ios-sync-engine level when adopting the new API.


Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -229,9 +231,10 @@ extension ZMGenericMessage {
                         }
                     }
                     
-                    guard let encryptedData = try? sessionDirectory.encrypt(self.data(), recipientClientId: clientRemoteIdentifier) else {
+                    guard let encryptedData = try? sessionDirectory.encrypt(self.data(), for: clientRemoteIdentifier) else {
                         return nil
                     }
+                    print(clientRemoteIdentifier)
```
typfel: Delete?

marcoconti83: Yup

mikeger:
Just a clarification, would there be a migration step from the existing identifiers to the new ones?

marcoconti83:
@mikeger yes, there need to be one (done in wire-ios-sync-engine) or you will lose all sessions

marcoconti83:
@typfel fixed

PR #147, opened 2017-01-24
Only set the uploaded state to done when the placeholder was not yet uploaded

daehn:
# What's in this PR?

* Only set the uploaded state to done when the placeholder was not yet uploaded
* When a conversation degrades we do not want to send a `Asset.NotUploaded` message, but in case we already sent the placeholder (which should not happen when the conversation did degrade) we want to send it.


typfel:
Could this be covered by a test?

daehn:
@typfel @marcoconti83 updated, but I will add additional testing here as well as in `wire-ios-sync-engine` and will leave the `DO NOT MERGE` until then.

daehn:
@typfel @marcoconti83 added tests and fixed a logical error, please re-review.

PR #149, opened 2017-01-25
`persistentStoreMetadataForKey:` optimization

mikeger:
# Issue

During profiling I noticed that all the transcoders (and strategies) are taking a significant time to check if there is a new request available. The check is happening since the save happens on the database and it is possible that there where changes that could lead to the request creation.

During the further profiling it was found that one particular method subcall is taking more time than the others:

screen shot 2017-01-25 at 18 11 30

Method `persistentStoreMetadata(key:)` used to query `self.persistentStoreCoordinator?.metadata(for: store)[key]` that caused a significant performance impact (18% of all CPU used). 

# Investigation

It comes out that the method is returning `NSDictionary` in Obj-C interface and `[String: Any]` dictionary in Swift. Apparently the implementation of the method is done in Obj-C, because the call tree shows that the dictionary is converted from ObjC to swift by enumerating all the key-value pairs. This renders the method extremely slow, since the persistent store metadata in test used to contain ca. 40 elements. I did not clarified yet if the structure is apparently converted recursively.

# Solution

Unfortunately casting the return of `metadata(for: store)` to `NSDictionary` does not seem to help, because the implementation is still first converting the originally returned NSDictionary to swift dictionary first. 

As of my understanding, there is a certain logic in the conversion layer that is used to convert the ObjC APIs to Swift that is adding a non-optimal conversion.

Unfortunately there is no other way than rewriting the method in ObjC. Rewrite gives ca. 12x performance boost to the original code, reducing the execution time from around 2 seconds to 0.168 seconds:

screen shot 2017-01-25 at 18 17 19

# Conclusion

As of now, the way the implicit conversion to Swift type in `unconditionallyBridge...` is a not documented part of the app that is really hard to spot. For the future the best strategy to avoid the similar errors would be to note the behavior for ObjC APIs returning `NSDictionary` and converted to return swift dictionaries, where the returned object has a significant amount of entries. 

I checked the profiler output and it does not seem to be the significant issue in any other place.

[SwiftMetadata.trace.zip](https://github.com/wireapp/wire-ios-data-model/files/731926/SwiftMetadata.trace.zip)


ZMCDataModel.xcodeproj/project.pbxproj
```diff
@@ -2198,7 +2198,7 @@
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				COPY_PHASE_STRIP = NO;
 				CURRENT_PROJECT_VERSION = 4;
-				DEBUG_INFORMATION_FORMAT = dwarf;
+				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
```
vytis: Not sure what's the impact would be in this project, but dSYM generation takes a bit of time and might increase build time.

mikeger: The setting changed is for dSYMs to be created for the debug build. If it's off the profiler cannot pick any symbol info, very annoying if you ask me.

vytis: Yeah, but building a debug build happen way more often than profiling - maybe there is no problem there, really depend on project. A (random) blog post about this - http://holko.pl/2016/10/18/dsym-debug/

mikeger: I wonder if the regular debugging is possible without dSYM info?
Also if this is relevant for the UI project, since there we also always create the dSYM.

vytis: I don't think it has any use in regular debugging - as I understand when you debug there is some magic happening in the cpu or actually changing the binary itself to inject extra code. The dSYM contain mapping between between function addresses and actual readable method names and info, so profiling and crash symbolication needs it to be able to show anything useful.
Maybe @marcoconti83 knows a bit more about it, I will try to see how long it takes on UI project.

typfel:
Could we write a performance test for this?

mikeger:
@typfel not sure, do we have the performance tests now?

ZMCDataModel.xcodeproj/project.pbxproj
mikeger: Ok if it's not useful for the regular debugging I would remove it.

daehn:
I assume we could write a test on a separate branch that artificially hits this method **a lot**, run it, merge this one in and see what it tells us. This way we also will know as soon as there is some regression.

mikeger:
@daehn I can also upload the trace I recorded, hopefully this would be enough?

daehn:
I am more thinking about tracking regressions in this regard, but I am happy to see the trace as well 😊

daehn:
But I guess as long as no-one rewrites this in part in Swift again we should be fine.™

ZMCDataModel.xcodeproj/project.pbxproj
marcoconti83: @vytis I'm not sure how it works exactly either. I just know that the dSYM is used to symbolicate profiling and crash logs, as you mentioned.

vytis: So I tried some simple benchmarking with Xcode showing build time (using ` defaults write com.apple.dt.Xcode ShowBuildOperationDuration YES`) for the UI project. It takes constant time, so mostly noticeable in small incremental compilations. 

I was commenting in/out one line of Swift code and building - after several tries it seems the version with sDYM takes around 24s to build and the one without is around 16s. So in absolute numbers not a big difference, but not negligible.

Tried the same for Data model project, but the difference is too small to notice using this way to measure build time.

marcoconti83:
Then maybe we should add some code documentation, that says that this is explicitly written in Obj-c to avoid paying the Swift conversion cost. Otherwise it will be swiftified at some point!

mikeger:
```
------------------------------------------------------------------------------------------------
// // // // // // // // // // No Swifting beyond this point // // // // // // // // // //
------------------------------------------------------------------------------------------------
```

mikeger:
I added the comment to the objc implementation.

ZMCDataModel.xcodeproj/project.pbxproj
mikeger: @vytis ok let's remove dSYMs for debug on UI project too then. One thing we need to take care about is building for simulator for QA automation, I believe they still need dSYMs even on debug.

PR #150, opened 2017-01-27
Calling 3

typfel:
## Why this PR
For implementing the new calling protocol version 3 we needed to re-factor the calling data models so that they can support two different calling protocols co-existing. 

## What changed
- Calling related data model has been moved to `wire-ios-sync-engine`.
- Calling observers has been moved to `wire-ios-sync-engine`.
- Calling conversation lists have been deleted since they are not necessary for the UI anymore.


typfel:
*test pass*

daehn:
@typfel What about the copyright headers, If the files have been there before it should be fine.

typfel:
Only `ZMConversation+Predicates.swift` but I might have created it last year :)

Source/ZMCDataModel.h
```diff
@@ -70,11 +70,11 @@
 
 #import <ZMCDataModel/ZMNotifications.h>
 #import <ZMCDataModel/ZMNotifications+Internal.h>
-
-#import <ZMCDataModel/ZMVoiceChannel.h>
-#import <ZMCDataModel/ZMVoiceChannel+Internal.h>
-#import <ZMCDataModel/ZMVoiceChannelNotifications.h>
-#import <ZMCDataModel/ZMVoiceChannelNotifications+Internal.h>
+//
```
basine: you probably wanted to remove those

Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
```diff
@@ -27,7 +27,7 @@ public protocol ZMGeneralConversationObserver {
 extension ZMConversation : ObjectInSnapshot {
     
     public var observableKeys : [String] {
-        var keys = ["messages", "lastModifiedDate", "isArchived", "conversationListIndicator", "voiceChannelState", "activeFlowParticipants", "callParticipants", "isSilenced", SecurityLevelKey, "otherActiveVideoCallParticipants", "displayName", "estimatedUnreadCount", "clearedTimeStamp"]
+        var keys = ["messages", "lastModifiedDate", "isArchived", "conversationListIndicator", "isSilenced", SecurityLevelKey, "displayName", "estimatedUnreadCount", "clearedTimeStamp"]
```
basine: how do we notice that the conversationListIndicator changed due to an incoming call?

Tests/Source/Model/Conversation/ZMConversationTests.m
```diff
@@ -2731,12 +2729,21 @@ - (void)setConversationAsHavingMissedCall:(ZMConversation *)conversation
...
-
 - (void)setConversationAsHavingActiveCall:(ZMConversation *)conversation
 {
     conversation.callDeviceIsActive = YES;
 }
 
+- (void)setConversationAsHavingIgnoredCall:(ZMConversation *)conversation
+{
+    conversation.isIgnoringCall = YES;
+//    conversation.callDeviceIsActive = YES;
```
basine: forgot to remove the comment

Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
typfel: isn't `conversationListIndicator` part of the snapshot  and will then detect a change on a incoming call?

basine: `conversationListIndicator` is a computed property in this case evaluating `callDeviceIsActive` which in turn is only stored in memory.  Therefore a change `callDeviceIsActive` will not result in the conversation being included in the ObjectDidChange Notification. We will still need to forward it as a non-core-data change. On the other hand - callDeviceIsActive only changes when you join or leave a call in which case you will most likely not be looking at the conversationlist. So.... I guess we don't really care about that change.

Source/Notifications/ObjectObserverTokens/ManagedObjectContextObserver.swift
```diff
@@ -272,7 +269,6 @@ public final class ManagedObjectContextObserver: NSObject {
     fileprivate var globalUserObserver : GlobalUserObserver!
     
     fileprivate var accumulatedChanges = ManagedObjectChanges()
-    fileprivate var changedCallStateConversations = ManagedObjectChanges()
     fileprivate var isSyncDone = false
     
     public let callCenter = CTCallCenter()
```
basine: Are you going to move this into your own calling thingy?

typfel: I have not moved it but I suppose I could since the voice channel is no longer here.

Source/Notifications/ObjectObserverTokens/ConversationObserverToken.swift
typfel: :)

PR #152, opened 2017-01-31
Fix marking system message as read

typfel:
It wasn't possible to mark a system message as read since we didn't update
the lastServerTimestamp when appending them to a conversation.


marcoconti83:
This is a tricky one, in theory, adding a system message that is generated locally (e.g. "new device") should not bump the conversation up in the conversation list (which is what happens if you update the server timestamp), otherwise your other devices will see a different order for the conversation list. 

There might be a different way to mark those as "read". @basine is there a way?

If not, we should rely (but it's a bit fragile) on the fact that in order to degrade a conversation (this PR is about system messages degrading conversations) you must have just sent or received a message, hence the conversation should already have a `lastServerTimestamp` that is really, really close to the one we are setting (like, max a few seconds away).

typfel:
I also use this method when adding missed calls system messages and in that case the `lastServerTimestamp` is not close to the timestamp of the added message. So maybe I should add a parameter for controlling if the message should generate an unread dot.

But the issue still remains for degradation messages.

typfel:
The Issue comes from the fact that when we mark messages as read and the message is the last message in a conversation we take the `lastServerTimestamp` instead of the `serverTimestamp` of the message.

Don't why we do this but one solution would be to take the most recent of `lastServerTimestamp` and `serverTimestamp.

typfel:
@marcoconti83  @basine updated the PR with the solution we discussed/

PR #153, opened 2017-02-01
Add patch to persistent store data when loading/creating DB

marcoconti83:
We need to make sure that client session identifier are migrated. If we don't guarantee this step before Core Data is used, we will break all stored session.
Enforcing this at this level guarantees that both share extension and main app will perform migration, no matter which one is opened first.

# Changes
I added a migration right after the NSManagedObjectContext initialization stage (sync), so we are sure that this will be applied

# For the future
Unify this code (at this level!) with the `ZMHotFix` which is present in `wire-ios-sync-engine`.
See discussion on: https://github.com/wireapp/wire-ios-sync-engine/pull/239


ZMCDataModel.xcodeproj/project.pbxproj
```diff
@@ -2385,6 +2413,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = F9C9A67C1CAD7A790039E10C /* ios-test-host.xcconfig */;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
typfel: So now we are adding this back again?

marcoconti83: It needs to be added to the test target or it won't run. It was removed from the test target.
On the framework target, it should be off (or apple will reject).

PR #154, opened 2017-02-01
Remove abbreviated displayname 

basine:
The UI won't display the abbreviated displayName anymore. I took this chance to rewrite the displayNameGenerator and related classes in Swift. This however leads to a performance drop due to Swift - ObjectiveC bridging of String / NSString :-/


Source/Model/User/DisplayNameGenerator.swift
```diff
@@ -0,0 +1,171 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+public class DisplayNameGenerator : NSObject {
+    public var idToPersonNameMap : [NSManagedObjectID: PersonName] = [:]
```
daehn: It looks like this property could be declared as `private`.

Source/Model/User/DisplayNameGenerator.swift
```diff
@@ -0,0 +1,171 @@
...
+        mapUsersToNames(returnChanges: false)
+    }
+    
+}
+
+let DisplayNameGeneratorKey = "DisplayNameGenerator"
+
+extension NSManagedObjectContext {
+    
+    var nameGenerator : DisplayNameGenerator? {
```
daehn: Does this object exist in both contexts? If not an assert could be added to ensure it is accessed from the correct thread.

Tests/Source/Model/User/ZMPersonNameTests.m
```diff
@@ -18,15 +18,15 @@
 
 
 #import "ZMBaseManagedObjectTest.h"
-#import "ZMPersonName.h"
-#import "NSString+ZMPersonName.h"
+//#import "PersonName.h"
```
daehn: Can be removed?

Source/Model/User/DisplayNameGenerator.swift
basine: The displayNameGenerator is set when setting up the uiMOC. Adding an assert makes a lot of tests crash that use `performPretendingUiMocIsSyncMoc` :-/

daehn:
💣 Conflicting

PR #157, opened 2017-02-03
Move from shared UserDefaults to writing to file in shared container

daehn:
# What's in this PR?

* In order to keep the main application up to date with changes made to the database in the share extension we were serializing `NSManagedObjectContextDidSave` notifications and storing them in the shared `UserDefaults`, this can be quite a lot of notifications and as we now want to store additional tracking data for the main app to be picked up it makes sense to move away from `UserDefaults` and storing the data on disk in the shared container.
* To achieve this in a reusable way a generic class `SharedObjectStore<T>` has been added, were `T` is the kind of object that can be stored and retrieved. An instance is instantiated by providing the shared container URL and a name for the file in which the data should be stored in.
* The existing `ContextDidSaveNotificationPersistence`, as well as the newly added `ShareExtensionAnalyticsPersistence` now use this class internally to store and retrieve the data.

#### Considerations

* At the moment we only write to the shared container from the share extension and then read from the main application. We can also ensure that the share extension is not writing while the main application is in the foreground (which is the only point at which the main app reads). However as we might add more extension this could change in the future. In order to provide safe access from multiple processes we could make use of a `FileCoordinator` to synchronize the file access between those processes.


Source/Model/Conversation/SharedObjectStore.swift
```diff
@@ -0,0 +1,183 @@
...
+
+    public func load() -> [T] {
+        if !fileManager.fileExists(atPath: url.path) {
+            zmLog.debug("Skipping loading shared file as it does not exist")
+            return []
+        }
+
+        do {
+            let data = try Data(contentsOf: url)
+            let stored = NSKeyedUnarchiver.unarchiveObject(with: data) as? [T]
```
mikeger: So potentially T can be constrained to `NSCoding` ?

mikeger: I guess it does not matter since Objective-C duck typing interface is used in NSCoding

daehn: I tried it, but for example `[AnyHashable: AnyObject]` is not visibly compliant to `NSCoding`, but is achievable using `archiveRootObject` as it takes an `Any`. I tried constraining the `[AnyHashable: AnyObject]` type but did not have any luck so far (`AnyHashable` is a type and thus can't be used in protocol composition).

PR #163, opened 2017-02-08
Added CircleCI integration

vytis:
## Why we might want CircleCI?

It would be good to have tests running on pull requests. CircleCI looks like the most popular tool for this so why not to try it.

## What's in this PR?

Added a config that could be reused in all other repos, because it doesn't have any `wire-ios-data-model` code. The resulting config is slightly convoluted, because it re-uses exactly the same build scripts as we currently use on Jenkins.

Let's see how it all works out and maybe we can set up other repos in the same fashion.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source 'https://rubygems.org'
 
 gem 'danger', '~>3.2'
+gem 'slather'
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,40 @@
+machine:
+  xcode:
+    version: "8.2"
+  environment:
+    # Used for compatibility with scripts made for Jenkins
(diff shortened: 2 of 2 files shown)


mikeger:
@vytis can we see a demo run from PR, or this one should be merged first?

vytis:
Well it's actually turned on for all PRs in this repository, so it will fire on any PR, but will fail pretty quickly because it will unsuccessfully try to build the project with default settings.
Another option would be to make a new PR to this branch - it will have the config and should complete. The actual results of the run can be seen on [CircleCI report](https://circleci.com/gh/wireapp/wire-ios-data-model/63?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

PR #165, opened 2017-02-13
Add normalizedText property and add TextSearchQuery class to perform FTS

daehn:
# What's in this PR?

* Add a new data-model version `2.26.0` which adds one new property on the `Message` entity, `normalizedText` which will be used to store a normalized version of the message text in order to be searched.
* Add the new class `TextSearchQuery` which will be used to perform a search for messages in a given conversation using a query string.
* The search will first look for already indexed messages which have a non-nil `normalizedText` field and perform a optimised query using the `[n]` keyword to let SQLite know that the data is already normalized. After this search non-normalized messages, if still present in the conversation, will be fetched, normalized and checked for matches.

**🚫 This is an intermediate PR and there are still some test cases and TODO's missing.**


Source/Model/Message/ZMMessage.m
```diff
@@ -639,7 +641,7 @@ + (NSPredicate *)predicateForMessageInConversation:(ZMConversation *)conversatio
     NSSet *noncesData = [nonces mapWithBlock:^NSData*(NSUUID *uuid) {
         return uuid.data;
     }];
-    NSPredicate *noncePredicate = [NSPredicate predicateWithFormat:@"%K IN %@", noncesData];
+    NSPredicate *noncePredicate = [NSPredicate predicateWithFormat:@"%K IN %@", noncesData]; // FIXME? How can this work at all?
```
daehn: Not related to this PR, but I came across this and am unsure how this can work, but will check if we actually call this method `predicateForMessageInConversation:withNonces:` from somewhere.

vytis:
Will this pose a problem for highlighting search results in UI?

daehn:
@vytis this only returns the messages that the normalized search query is contained in and does not return any ranges of matches. Checking the range of a query inside of the matched message texts can be done on the UI, where the original (non-normalized) query and the message are present. So we can just take the original query and get the ranges from the non-normalized message text there.

circle.yml
```diff
@@ -0,0 +1,40 @@
+machine:
```
mikeger: Hmm comes with free circle integration?

Resources/Configurations/version.xcconfig
```diff
@@ -1,2 +1,2 @@
 MAJOR_VERSION = 42
-CURRENT_PROJECT_VERSION = 42.0.3
+CURRENT_PROJECT_VERSION = 42.0.5
```
mikeger: Hmm bumps project version?

circle.yml
daehn: @mikeger  I merged develop into `feature/add-normalized-text-field` but not into the base `feature/text-search`, should be fixed now.

Resources/Configurations/version.xcconfig
daehn: @mikeger same here, should be caused by the merge mentioned above.

daehn:
I assume we can try something like this on the UI to get the ranges which performs a diacritic insensitive search, but I still assume there will be edge cases in which we might not be able to show the correct highlights:

```swift
let text = "håkon"
let query = "ā"
let range = test.range(of: query, options: [.diacriticInsensitive, .caseInsensitive], range: text.startIndex..<text.endIndex, locale: nil)
let substring = range.map(text.substring)
```

Source/Model/Message/TextSearchQuery.swift
```diff
@@ -0,0 +1,304 @@
...
+    private var numberOfIndexedMessages = 0
+    private var numberOfNonIndexedMessages = 0
+
+    /// Creates a new `TextSearchQuery` object.
+    /// - param conversation The conversation in which the search should be performed. Needs to belong to the UI context.
+    /// - param query The query string which will be searched for in the messages of the conversation.
+    /// - param delegate The delegate which will be notified with the results.
+    /// - param configuration An optional configuration specifying the fetch batch size (useful in tests).
+    public init?(
+        conversation: ZMConversation,
```
mikeger: We would need to implement the global search first, just something to keep in mind. Current architecture looks good for it.

PR #175, opened 2017-02-15
Don't resort hidden messages since it will make them visible again

typfel:
Fix for https://wearezeta.atlassian.net/browse/ZIOS-8063

Also remove duplicated code I discovered while investigating this.

### DIFF
--- a/Source/Model/Conversation/ZMConversation.m
+++ b/Source/Model/Conversation/ZMConversation.m
@@ -981,6 +981,7 @@ const NSUInteger ZMConversationMaxTextMessageLength = ZMConversationMaxEncodedTe
 - (void)resortMessagesWithUpdatedMessage:(ZMMessage *)message
 {
     if (message.visibleInConversation == nil) {
+        ZMLogWarn(@"Attempt to resort message not visible in conversation");
         return;
     }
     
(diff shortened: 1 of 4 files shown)


Source/Model/Message/ZMMessage.m
```diff
@@ -570,13 +570,6 @@ + (instancetype)fetchMessageWithNonce:(NSUUID *)nonce forConversation:(ZMConvers
                 return (id) message;
             }
         }
-        for (ZMMessage *message in conversation.hiddenMessages) {
```
daehn: Where is the duplication?

basine: Why are we removing this?

typfel: on line 584-590

typfel: Because this code is run again on on line 584-590 inside the `if` statement that check if the hiddenMessages is not faulted.

basine: Makes sense, yes. I assume you ran all tests

typfel:
tests are crashing. *investigating*

typfel:
@basine @daehn fixed tests, please re-review.

Source/Model/Conversation/ZMConversation.m
```diff
@@ -980,6 +980,10 @@ - (void)sortMessages
 
 - (void)resortMessagesWithUpdatedMessage:(ZMMessage *)message
 {
+    if (message.visibleInConversation == nil) {
```
daehn: Maybe we should even log an error in that case, as it seems like a programmer error if you try to resort a message that is not in the conversation in the first place.

typfel: Sure why not

PR #176, opened 2017-02-15
Conversation security degradation changes

vytis:
## What's the reason behind this PR?

We want to prevent users from unknowingly sending messages to conversations which became not verified. At the moment we ask the user only once and if she dismisses the screen the conversation is automatically marked as not secure. These changes will avoid this situation and pester the user until she agrees to degrade the conversation to not secure by tapping "send anyway". 

This is first of several PRs that will tweak the current behaviour. 

## What's in this PR?

* Slightly changed the semantics of conversation security levels
* When we resend the messages that caused degradation the conversation is marked as not secure instead of staying at "secure with ignored" state
* Slightly simplified the way messages that caused degradation are stored in `userInfo` of MOC
* Resending messages after degradation possible only explicitly one by one, not all of them at once.
* Updated tests


marcoconti83:
Merged. @vytis please make release and bump in all dependend PRs

typfel:
@marcoconti83 @vytis you know this PR just merged into `feature/degradation-changes`

marcoconti83:
oops :)

Source/Model/Conversation/ZMConversationSecurityLevel.h
```diff
@@ -20,21 +20,22 @@
 /* Secure level can change only the following way:
  
  NotSecure -> Secure
-                ^
-                |
-                v           
-            PartialSecureWithIgnored
+     ^          ^
+      \         |
```
daehn: So the general logic changed now? Before it was not possible to ever get back to the NotSecure state as a user could only ignore clients and still send messages which is when we show the half shield (that's never shown if the conversation was never secure before).

daehn: Ok just read the description again ;)

daehn: But I think this could affect the system messages which are being inserted ATM.

vytis:
Yeah it's merged into a separate branch which is the same as develop. Initially I was thinking that this functionality will be split into several PRs, but it didn't happen. Will make a new PR to develop with this.

Source/Model/Conversation/ZMConversationSecurityLevel.h
vytis: I am not super familiar with the

vytis: So the current behaviour is that we don't let any messages to be send at the `wire-ios-message-strategy` if conversation security status is `PartialSecureWithIgnored` (https://github.com/wireapp/wire-ios-message-strategy/pull/40). Not completely sure what was the previous behaviour with half shield though. Maybe @marcoconti83 can expand on this?

PR #184, opened 2017-02-17
Call the delegate with empty results if there are no messages to be searched

daehn:
# What's in this PR?

* When searching in a conversation without any messages that can be searched the `TextSearchQueryDelegate` was never notified.
* Fixes the failing `wire-ios-sync-engine` test `testThatItDoesNotFindAMessageDeletedRemotely`.


vytis:
Maybe it's worth adding a test here as well? Would have been good to catch it in this framework instead

daehn:
@vytis good point, updated!

PR #187, opened 2017-02-21
Do not re-insert client message with uploaded link preview

daehn:
# What's in this PR?

* This logic is moved into a new strategy which utilises a modified object sync instead.
* This modifies the internal behaviour heavily, so I am unsure if it shouldn't be a major release.


typfel:
No tests broke because of this change?

daehn:
@typfel nope, all ✅ in this framework.

PR #188, opened 2017-02-21
Fix for flaky test

vytis:
It seems the test introduced in #186 is quite flaky, trying to make it more stable

### DIFF
--- a/Source/Model/Conversation/ZMConversationSecurityLevel.swift
+++ b/Source/Model/Conversation/ZMConversationSecurityLevel.swift
@@ -1,0 +1,32 @@
+//
+// Wire
+// Copyright (C) 2017 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


Source/Model/Conversation/ZMConversationSecurityLevel.swift
```diff
@@ -0,0 +1,32 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ZMConversationSecurityLevel: CustomDebugStringConvertible {
```
mikeger: Additional debug utility? Does not seem to be used.

vytis: It's used when calling `.debugDescription`, which happens when `XCTAssert` fails. Without it we get this sort of failure:

```
XCTAssertEqual failed: ("ZMConversationSecurityLevel") is not equal to ("ZMConversationSecurityLevel") - 
```

mikeger: Ah nice

mikeger: The thing could also just display it's `rawValue` per default. I always thought it's the another annoying apple testing bug.

PR #190, opened 2017-02-22
Fixed failing images tests

mikeger:
Errors are happening mostly because we now preprocess the images before uploading, removing the metadata.

For some reason, removing the metadata twice still produces the data that is different from the one where the metadata is removed once:

screen shot 2017-02-22 at 12 23 25

So I decided that it would be equally good to compare the data length with the expected data.

### DIFF
--- a/Tests/Source/Model/Conversation/ZMConversationTests+messages.m
+++ b/Tests/Source/Model/Conversation/ZMConversationTests+messages.m
@@ -19,6 +19,7 @@
 
 #import "ZMConversationTests.h"
 #import "ZMClientMessage.h"
+@import zimages;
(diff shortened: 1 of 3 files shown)


vytis:
Could it happen because metadata is stored in a dictionary-like structure, so they get reshuffled in memory when resaved? Maybe there is a way to prevent that? Checking for file length doesn't seem too robust, but maybe not so bad as it's only tests...

mikeger:
@vytis potentially we can compare the bitmaps, but it seemed like an overkill for me

vytis:
Yeah that would be too much I guess.

PR #192, opened 2017-02-23
Expose store URL for debugging purposes

marcoconti83:
Expose the store URL to easily access the raw store (e.g. when running on simulator)


mikeger:
@marcoconti83 can this be helpful / be used instead of saving the store URL? https://developer.apple.com/reference/coredata/nspersistentstorecoordinator/1468865-url

marcoconti83:
@mikeger it looks useful! I will change to use that, thanks for the tip

marcoconti83:
@mikeger please re-review

PR #193, opened 2017-02-24
Add new model version and index on Message.expirationDate

daehn:
# What's in this PR?

* Add new model version `2.27.0` which adds an index on the `expirationDate` field on the `Message` entity. This index should speed up the fetching of ephemeral messages on startup where we fetch all message that have a non-nil `expirationDate`.
* As adding an index alone does not trigger a Core Data migration a version hash modifier was set for the changed attribute.
* Add a migration test that checks that Core Data prepares the local store and performs the migration.


mikeger:
May be adding of the index is a trivial change for CoreData, so it's not necessary to have the migration?

mikeger:
@daehn sorry, clicked the wrong button

mikeger:
@daehn ok, I think you are doing it right http://stackoverflow.com/questions/2178781/add-core-data-index-to-certain-attributes-via-migration

PR #199, opened 2017-02-27
Hide session reset message

daehn:
# What's in this PR?

* Add session reset message to hidden messages instead of visible ones.


marcoconti83:
Do we need to add it at all? Can't we just throw it away?

typfel:
@marcoconti83 this is when you send a session reset, so I guess it needs the message needs to be in a conversation to be picked up by the client message transcoder.

daehn:
We need to add it in order for the client message transcoder to pick it up, but generally we could also do it the same way we do it with calling v3 signalling messages which AFAIK do neither get added to a conversation nor persisted (cc @typfel).

typfel:
@daehn yes we could. Look at `GenericMessageStrategy`

daehn:
@marcoconti83 @mikeger @typfel updated, the associated `wire-ios-message-strategy` is https://github.com/wireapp/wire-ios-message-strategy/pull/43.

PR #202, opened 2017-03-07
added cases for special character search

evr3n:


Tests/Source/Model/TextSearchQueryTests.swift
```diff
@@ -322,6 +322,65 @@ class TextSearchQueryTests: BaseZMClientMessageTests {
...
+        verifyThatItFindsMessage(withText: "https://www.link.com/something-to-read?q=12&second#reader", whenSearchingFor: "something to read")
+        verifyThatItFindsMessage(withText: "<8000 x a's>", whenSearchingFor: "<8000 x a's>")
+        verifyThatItFindsMessage(withText: "bb бб bb", whenSearchingFor: "бб")
+        verifyThatItFindsMessage(withText: "bb бб bb", whenSearchingFor: "bb")
+        verifyThatItFindsMessage(withText: "苹果", whenSearchingFor: "Ping")
+
+
+
+
+
```
mikeger: Looks good. 
```










```
The newlines seems unnecessary.

Tests/Source/Model/TextSearchQueryTests.swift
```diff
@@ -322,6 +322,65 @@ class TextSearchQueryTests: BaseZMClientMessageTests {
...
+        verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "bb")
+        verifyThatItFindsMessage(withText: "bb aa", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa bb", whenSearchingFor: "aa bb")
+        verifyThatItFindsMessage(withText: "aa", whenSearchingFor: "bb ")
+        verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "aa\nbb")
+        verifyThatItFindsMessage(withText: "aa aa aa", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa bb aa", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa bb", whenSearchingFor: "aa cc")
+        verifyThatItFindsMessage(withText: "aa aa aa", whenSearchingFor: "aa aa")
+        verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "a")
```
mikeger: I thought we don't have any support for searching one letter. Should there be the cases for it?

daehn: Creating the `TextSearchQuery` object should fail in that case. I will check the test setup if it handles this correctly.

Tests/Source/Model/TextSearchQueryTests.swift
```diff
@@ -322,6 +322,65 @@ class TextSearchQueryTests: BaseZMClientMessageTests {
         verifyThatItFindsMessage(withText: "aa aa aa", whenSearchingFor: "aa")
         verifyThatItFindsMessage(withText: "aa.aa", whenSearchingFor: "aa")
         verifyThatItFindsMessage(withText: "11:45", whenSearchingFor: "11:45")
+        verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "aabb")
+        verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "bb")
+        verifyThatItFindsMessage(withText: "bb aa", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa bb", whenSearchingFor: "aa bb")
+        verifyThatItFindsMessage(withText: "aa", whenSearchingFor: "bb ")
```
daehn: Does this test pass? Searching for `"bb "` should not return a message with text `"aa"`

Tests/Source/Model/TextSearchQueryTests.swift
```diff
@@ -322,6 +322,65 @@ class TextSearchQueryTests: BaseZMClientMessageTests {
...
+        verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "aa\nbb")
+        verifyThatItFindsMessage(withText: "aa aa aa", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa bb aa", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa bb", whenSearchingFor: "aa cc")
```
daehn: This one should not pass as well.

Tests/Source/Model/TextSearchQueryTests.swift
```diff
@@ -322,6 +322,65 @@ class TextSearchQueryTests: BaseZMClientMessageTests {
...
+        verifyThatItFindsMessage(withText: "aa-bb", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa/bb", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa/bb", whenSearchingFor: "bb")
+        verifyThatItFindsMessage(withText: "aa:bb", whenSearchingFor: "aa")
+        verifyThatItFindsMessage(withText: "aa:bb", whenSearchingFor: "bb")
+        verifyThatItFindsMessage(withText: "@peter", whenSearchingFor: "peter")
+        verifyThatItFindsMessage(withText: "rené", whenSearchingFor: "Rene")
+        verifyThatItFindsMessage(withText: "https://www.link.com/something-to-read?q=12&second#reader", whenSearchingFor: "something to read")
+        verifyThatItFindsMessage(withText: "<8000 x a's>", whenSearchingFor: "<8000 x a's>")
+        verifyThatItFindsMessage(withText: "bb бб bb", whenSearchingFor: "бб")
```
mikeger: This test case seems not correct, I guess the aim is to proof that searching for cyrillic `b`, i.e. `б` in the text with latin b is working. It should be "bb" matching search for "бб". Same applies for the next test, but reversed, i.e. searching for Cyrillic B should return the latin one.

mikeger:
@evr3n did you run the tests?

Tests/Source/Model/TextSearchQueryTests.swift
evr3n: Is there a way to verify that it doesn't find? To e.g. test the one letter cases

daehn: Yes there is like this:
```swift 
verifyThatItFindsMessage(withText: "aabb", whenSearchingFor: "a", shouldFind: false)
```

Tests/Source/Model/TextSearchQueryTests.swift
daehn: @evr3n I pushed some changes on this branch to ensure the tests are not crashing anymore but failing instead if the creation of a `TextSearchQuery` fails (which happens when the search query is shorter than 2 characters long).

mikeger:
@evr3n this works now, tests are ✅ ?

daehn:
@mikeger test were run by CircleCi and are

PR #204, opened 2017-03-08
Add performed call system message

daehn:
# What's in this PR?

**TL;DR: Adds the logic to reference associated system message from a system message, which is used to batch multiple call related system messages.**

* The `.performedCall` case has been added to `ZMSystemMessageType`. A message of this type will be inserted when calling `appendPerformedCallMessage` on a conversation. Both `appendPerformedCallMessage` and the existing `appendMissedCallMessage` check the previous message and update it in case it is a system message of the same type (and the same caller etc.).
* In order to do so a new data model version `2.28.0` has been added. It adds the new fields `duration` (used to store the duration of a performed call) as well as the `childMessages` relation and its inverse, `parentMessage` to the `SystemMessage` entity.
* The added relationship is used to reference other system messages. In case of `.performedCall` and `.missedCall` a message will be hidden in a conversation and instead added to this relation when the previous message is of the same type. This way the UI can show the subsequent calls in a unified way.
* This PR also updates the `MessageObserverTests` by making the `verify` helper method generic (this avoids a lot of casting in the modification blocks). There also was a logical error as some change info keys were not included in the array in the helper method. This lead to some tests passing even though they were setup with wrong keys that will never change. This was fixed by checking that the expected key is actually contained in the change info keys. If a new test with a new flag in the `MessageChangeInfo` is added and the key is not added to the helper methods array this will fail now.
* As this PR adds a new model schema version it also adds a corresponding migration test.


daehn:
**There is one TODO item left for this PR:**

- [ ] Similar to https://github.com/wireapp/wire-ios-data-model/pull/206 a conversation should be unarchived when accepting an incoming call as well.

daehn:
Fixing tests

daehn:
Fixed ✅

mikeger:
@daehn I trust you, but "ci/circleci — Your tests failed on CircleCI" -- something is broken in the integration?

daehn:
@mikeger it failed to run the tests on CircleCI, I restarted them.

PR #209, opened 2017-03-13
Update ZMCommonContactsSearchDelegate protocol

daehn:
# What's in this PR?

* We will stop using the `/search/common` endpoint and the UI only needs to know the number of common connections. This PR updates the `ZMCommonContactsSearchDelegate` protocol accordingly.

### DIFF
--- a/Source/Public/ZMBareUser.h
+++ b/Source/Public/ZMBareUser.h
@@ -70,7 +70,7 @@
 
 @protocol ZMCommonContactsSearchDelegate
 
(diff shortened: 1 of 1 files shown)


marcoconti83:
Why is this protocol defined here but then never used here?

daehn:
Was wondering the same. I guess we can move it into SE. I will close this PR and update the SE one!

daehn:
@marcoconti83 updated!

marcoconti83:
Thanks!

PR #214, opened 2017-03-15
Remove extreme combinig characters from varoius text data

mikeger:
Changed fields:

1. Text messages
2. Conversation names
3. User names
4. File names
5. Link preview and twitter titles / messages
6. Location names
7. Conversation rename system messages


Source/Model/Message/LinkPreview+ExcessiveDiacritics.swift
```diff
@@ -0,0 +1,49 @@
...
+        newStatus.username = self.username?.removingExtremeCombiningCharacters
+        newStatus.author   = self.author?.removingExtremeCombiningCharacters
+        return newStatus
+    }
+}
+
+internal extension Article {
+    @objc(linkPreviewByRemovingExcessiveDiacriticsUsage)
+    override var removingExcessiveDiacriticsUsage: Article {
+        let newStatus = Article(protocolBuffer: self.protocolBuffer)
```
daehn: Doesn't the `init(protocolBuffer:) already remove the excessive diacritics?

mikeger: True!

mikeger:
@daehn updated

mikeger:
@marcoconti83 I updated it in order to clean the UserClient info, but I think the other cases are a stretch, since if user does not have control over his own account there are things much worse to stop him using the app, for example deleting this user's other user clients or account hijack with changing the email credentials. 
In regards of the address book, it is not the case because if such text is there then the user would be unable to use the iOS address book app.

marcoconti83:
@mikeger fair point, it has to be your own device and if this is possible, then you have other things to worry about.  However if it requires only adding `.removingExtremeCombiningCharacters` in two lines of code, I would still do it for email and phone. This code is executed only once you update from another device so I don't think it's a performance concern.

mikeger:
@marcoconti83 updated, please re-verify

PR #215, opened 2017-03-16
Profile Picture V3: Added asset identifiers

vytis:


Source/Model/User/ZMUser.swift
```diff
@@ -0,0 +1,56 @@
...
+let previewProfileAssetIdentifierKey = #keyPath(ZMUser.previewProfileAssetIdentifier)
+let completeProfileAssetIdentifierKey = #keyPath(ZMUser.completeProfileAssetIdentifier)
+
+extension ZMUser {
+    public var previewProfileAssetIdentifier: String {
+        set {
+            modifyValue(newValue, forKey: previewProfileAssetIdentifierKey)
+        }
+        get {
+            return getValue(forKey: previewProfileAssetIdentifierKey)
```
daehn: Is it necessary to overwrite the getter here? The `getValue` implementation looks like it does the same as the default implementation.

daehn: But maybe it would need to marked as `@NSManaged` then.

vytis: It's Swift requirement actually - you can't have custom setter without getter. It seems `@NSManaged` only works without custom accessors, spits out an error in here.

daehn: Ahh, too bad!

Source/Model/User/ZMUser.swift
```diff
@@ -0,0 +1,56 @@
...
+
+import Foundation
+
+let previewProfileAssetIdentifierKey = #keyPath(ZMUser.previewProfileAssetIdentifier)
+let completeProfileAssetIdentifierKey = #keyPath(ZMUser.completeProfileAssetIdentifier)
+
+extension ZMUser {
+    public var previewProfileAssetIdentifier: String? {
+        set {
+            modifyValue(newValue, forKey: previewProfileAssetIdentifierKey)
```
daehn: Shouldn't the modified key only be set in case the user is the `selfUser`?

mikeger:
@vytis would you add the migration test later?

daehn:
@vytis see https://github.com/wireapp/wire-ios-data-model/wiki/How-to-write-a-Core-Data-migration-test

vytis:
@daehn @mikeger can you take a look at it now - added migration test.

PR #216, opened 2017-03-17
Allow having multiple message windows

mikeger:
- Before the `MessageWindowObserverCenter` allowed only one window snapshot.
- Bug happened when we used to display two conversation views at the same time (with force touch in conversation list).
- In this case the latter window was loosing the observer since it's snapshot was removed when first window was deallocated.

Also removed some old and unused code / notifications.

### DIFF
--- a/Source/Notifications/MessageWindowObserverCenter.swift
+++ b/Source/Notifications/MessageWindowObserverCenter.swift
@@ -22,11 +22,7 @@ import Foundation
 private var zmLog = ZMSLog(tag: "MessageWindowObserverCenter")
 
 extension Notification.Name {
-    
-    static let ZMConversationMessageWindowScrolled = Notification.Name("ZMConversationMessageWindowScrolledNotification")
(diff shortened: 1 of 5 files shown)


Source/Notifications/MessageWindowObserverCenter.swift
```diff
@@ -50,68 +46,65 @@ extension NSManagedObjectContext {
 
 @objc final public class MessageWindowObserverCenter : NSObject, ChangeInfoConsumer {
     
-    var windowSnapshot : MessageWindowSnapshot?
+    var windowSnapshots : [MessageWindowSnapshot] = []
```
marcoconti83: What's the average expected number of windows? 1/2 or more? 
If it becomes one per conversation, then I would use a map from window to snapshot for faster lookup. Otherwise if it's just a couple of windows, let's leave the array.

mikeger: @marcoconti83 should not exceed 2 at the moment

marcoconti83: All good then!

PR #218, opened 2017-03-22
Profile picture v3: updated image cache to show v3 image if available

vytis:
This PR enables interoperability between v3 and v2 profile images. It falls back to showing v2 images if v3 is not available.

It also includes changes to predicates used for fetching V2 images and will prevent them to be downloaded if v3 is available.


Source/Model/User/ZMUser.swift
```diff
@@ -18,14 +18,135 @@
...
+    public static let completeAssetFetchNotification = Notification.Name(rawValue:"ZMRequestUserProfileCompleteAssetV3NotificationName")
+
     @NSManaged public var previewProfileAssetIdentifier: String?
     @NSManaged public var completeProfileAssetIdentifier: String?
     
+    
+    @objc(setImageData:size:)
+    public func setImage(data: Data?, size: ProfileImageSize) {
+        let key = size.userKeyPath
+        willChangeValue(forKey: key)
```
daehn: The `willChangeValue` and `didChangeValue` can be moved inside the `if`, as that's the only case where we are actually changing the value for the key.

Source/Model/User/ZMUser.swift
```diff
@@ -18,14 +18,135 @@
...
+            managedObjectContext?.zm_userImageCache.setUserImage(self, imageData: imageData, size: size)
+        }
+        didChangeValue(forKey: key)
+        managedObjectContext?.saveOrRollback()
+    }
+    
+    @objc(imageDataforSize:)
+    public func imageData(for size: ProfileImageSize) -> Data? {
+        if isSelfUser {
+            return primitiveValue(forKey: size.userKeyPath) as? Data
```
daehn: Should this be wrapped in a `willAccesValue` / `didAccessValue`?

Source/Model/User/ZMUser.swift
vytis: I guess it would make sense to move the save in there as well?

Source/Model/User/ZMUser.m
```diff
@@ -1017,63 +1030,31 @@ - (void)setImageData:(NSData *)imageData forFormat:(ZMImageFormat)format propert
...
-        
-        if (self.originalProfileImageData != nil) {
-            [self setLocallyModifiedKeys:[NSSet setWithObject:key]];
-        }
-    }
-    else {
-        if (nil == imageData) {
-            [self.managedObjectContext.zm_userImageCache removeAllUserImages:self];
-            return;
+    switch (format) {
```
basine: So the selfUser profile image is treated the same way as other user's?

daehn: I think the logic moved into the swift extension here: https://github.com/wireapp/wire-ios-data-model/pull/218/files/90682ff10f97bb40d5138589ed18f9ea287c8f42#diff-4c1af6fa3965231c74cda4dd73e4a247R86

Source/Model/User/ZMUser.swift
vytis: Just for the record - I tried moving this to where the actual change happens but this broke some tests.

PR #220, opened 2017-03-27
Make sure sessions with prekeys are always established

marcoconti83:
Always re-create a session when we get a prekey, even if that session exists already

### DIFF
--- a/Source/Model/UserClient/UserClient.swift
+++ b/Source/Model/UserClient/UserClient.swift
@@ -429,11 +429,13 @@ public extension UserClient {
         keysStore.encryptionContext.perform { (sessionsDirectory) in
             
             // Session is already established?
-            guard !sessionsDirectory.hasSession(for: sessionIdentifier) else {
-                zmLog.debug("Session with \(sessionIdentifier) is already established")
-                return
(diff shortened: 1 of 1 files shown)


Source/Model/UserClient/UserClient.swift
```diff
@@ -429,11 +429,13 @@ public extension UserClient {
...
-                zmLog.debug("Session with \(sessionIdentifier) is already established")
-                return
+            if sessionsDirectory.hasSession(for: sessionIdentifier) {
+                zmLog.debug("Session with \(sessionIdentifier) was already established, re-creating")
+                sessionsDirectory.delete(sessionIdentifier)
             }
-            
+        }
+        
+        keysStore.encryptionContext.perform { (sessionsDirectory) in
```
typfel: What's the reason for doing this this in two separate `perform` blocks?

marcoconti83: Because of caching within one block, it commits to disk only at the end of a block. I don't think the cache is smart enough to perform the sum of operations (delete + recreate) if at the end of the block it just sees that the session with that ID is still there. This depends on cbox implementation, but just to be safe, I split it in two. I will add a code comment to make this more evident.

PR #221, opened 2017-03-28
Upload messages regardless of the “encrypted” flag

marcoconti83:
Sometimes an expired message will have the flag "encrypted" set to false (unexpectedly) and not upload. The flag is not actually used anymore and will be removed in another PR, that will affect multiple frameworks.
I'm just changing the predicate here as a quick fix.


mikeger:
No tests affected?

marcoconti83:
@mikeger  In a upcoming PR I will remove the "encrypted" flag completely. I hope to put it in today. I will test properly (or actually, remove tests) there.

PR #229, opened 2017-04-07
Fix: small and large images are mixed up

vytis:
Mapping between ZMImageFormat and size used in other places was reversed


daehn:
@vytis I kicked off a new path release

PR #231, opened 2017-04-10
Do not insert missed call system message from update event if sender is self

daehn:
# What's in this PR?

* We still receive update events for group calls until they use v3. This lead to the issue that we inserted a performed call system message when placing a group call, but also inserted a missed call system message in case we hung up before anyone connected (triggered by the received update event). We do not want to insert a missed call message in this case and ignore the update event if the sender is self from now on.
* This also exposes a couple of methods needed for https://github.com/wireapp/wire-ios-message-strategy/pull/56.


mikeger:
@daehn I started the data model release.

PR #233, opened 2017-04-11
Remove methods inserting assets without specifying protocol version

daehn:
# What's in this PR?

* There are some cases in which the UI is still using the methods which do not specify the protocol version and always use v2 asset sending. This PR removes all of these methods to enforce being explicit about the version that should be used.

### DIFF
--- a/Source/Model/Message/ZMAssetClientMessage.m
+++ b/Source/Model/Message/ZMAssetClientMessage.m
@@ -95,22 +95,6 @@ static NSString * const AssociatedTaskIdentifierDataKey = @"associatedTaskIdenti
 + (instancetype)assetClientMessageWithOriginalImageData:(NSData *)imageData nonce:(NSUUID *)nonce managedObjectContext:(NSManagedObjectContext *)moc expiresAfter:(NSTimeInterval)timeout;
 {
     [moc.zm_imageAssetCache storeAssetData:nonce format:ZMImageFormatOriginal encrypted:NO data:imageData];
-    
-    ZMAssetClientMessage *message = [ZMAssetClientMessage insertNewObjectInManagedObjectContext:moc];
-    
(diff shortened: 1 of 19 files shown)


daehn:
I am still not sure this is the best way (mostly as it introduces a lot of changes in all of the frameworks). Maybe introducing a global flag set on `ZMConversation` makes more sense?

daehn:
Updated to only expose methods which don't specify which version should be used. To enable v3 sending the class method `+setUseVersion3Assets:` on `ZMConversation` should be used from now on.

daehn:
This PR is still on hold until all dependent frameworks have been updated for the new API.

marcoconti83:
We want to remove sending v2 completely. We can remove the flag everywhere and we can just assume v3. It's gonna be a few changes across different frameworks, but we have to do it anyway at some point.

Source/Model/Message/ZMAssetClientMessage.m
```diff
@@ -509,7 +477,10 @@ + (NSPredicate *)predicateForObjectsThatNeedToBeUpdatedUpstream;
 
 - (void)updateWithPostPayload:(NSDictionary *)payload updatedKeys:(NSSet *)updatedKeys
 {
-    if ([updatedKeys containsObject:ZMAssetClientMessageUploadedStateKey] && self.uploadState == ZMAssetUploadStateUploadingPlaceholder) {
+    BOOL shouldUpdate = self.uploadState == ZMAssetUploadStateUploadingPlaceholder
+                     || (self.uploadState == ZMAssetUploadStateUploadingFullAsset && self.v3_isImage);
+
+    if ([updatedKeys containsObject:ZMAssetClientMessageUploadedStateKey] && shouldUpdate) {
```
daehn: We never updated the `serverTimestamp` for sent v3 images, as they will be in a state of `.uploadingFullAsset` (no more preview with v3).

daehn:
@marcoconti83 updated

PR #237, opened 2017-04-19
return ClientMessage when adding a reaction

typfel:
### what changed
Return the client message when adding a reaction

### why this change
I need to observe when a reaction is sent for the new background task API

### DIFF
--- a/Source/Model/Message/ConversationMessage+Reaction.swift
+++ b/Source/Model/Message/ConversationMessage+Reaction.swift
@@ -30,9 +30,9 @@ import Foundation
 
 extension ZMMessage {
     
-    static func appendReaction(_ unicodeValue: String?, toMessage message: ZMConversationMessage) {
-        guard let message = message as? ZMMessage, let context = message.managedObjectContext else { return }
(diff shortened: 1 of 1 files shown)


Source/Model/Message/ConversationMessage+Reaction.swift
```diff
@@ -30,9 +30,9 @@ import Foundation
 
 extension ZMMessage {
     
-    static func appendReaction(_ unicodeValue: String?, toMessage message: ZMConversationMessage) {
-        guard let message = message as? ZMMessage, let context = message.managedObjectContext else { return }
-        guard message.deliveryState == ZMDeliveryState.sent || message.deliveryState == ZMDeliveryState.delivered else { return }
+    static func appendReaction(_ unicodeValue: String?, toMessage message: ZMConversationMessage) -> ZMClientMessage? {
```
mikeger: Should it be `discardableResult` then?

typfel: This is an internal method and I think it makes sense to have the warning here.

PR #245, opened 2017-05-02
Tests: Clearing up some objects created during setUp

vytis:
Part of the ongoing process to optimise test suite memory usage (and possibly running time).


mikeger:
1obtnr

PR #246, opened 2017-05-03
Only ask for more prekeys when we've run out of them

typfel:


Source/Model/UserClient/UserClient.swift
```diff
@@ -464,15 +464,14 @@ public extension UserClient {
...
             numberOfKeysRemaining -= 1
         }
-        if(numberOfKeysRemaining < 0) { // this will recover from the fact that the number might already be < 0
-                                        // from a previous run
+        if numberOfKeysRemaining < 0 { // this will recover from the fact that the number might already be < 0
+                                       // from a previous run
             numberOfKeysRemaining = 0;
         }
-        if(oldValue != 0) {
+        if numberOfKeysRemaining == 0 {
```
marcoconti83: Is this line was causing the bug? Was it basically always uploading keys every time a new session was created?

typfel: Yes, as far as I can tell we were uploading a 100 new prekeys on every new session.

marcoconti83: Looking at the linked ticket, it had nothing to do with getting a "missing client" response from the BE, right? It was only when someone else would create a session with you with a prekey

Source/Model/UserClient/UserClient.swift
```diff
@@ -464,15 +464,14 @@ public extension UserClient {
     
     /// Use this method only for the selfClient
     public func decrementNumberOfRemainingKeys() {
-        let oldValue = numberOfKeysRemaining
```
mikeger: Sorry, I see that it was not there before, should we assert on calling this not on self client? 
`Use this method only for the selfClient`

Source/Model/UserClient/UserClient.swift
typfel: Yes, that's the only place where we are decrementing the number of prekeys

Source/Model/UserClient/UserClient.swift
```diff
@@ -464,15 +464,14 @@ public extension UserClient {
...
-        if(numberOfKeysRemaining > 0) {
+        if numberOfKeysRemaining > 0 {
             numberOfKeysRemaining -= 1
         }
-        if(numberOfKeysRemaining < 0) { // this will recover from the fact that the number might already be < 0
-                                        // from a previous run
+        if numberOfKeysRemaining < 0 { // this will recover from the fact that the number might already be < 0
+                                       // from a previous run
             numberOfKeysRemaining = 0;
```
mikeger: Would be cool to remove the semicolon (worst issue ever).

typfel:
@mikeger updated

PR #249, opened 2017-05-05
Add more buckets to message count clusterizer

daehn:
# What's in this PR?

* Add more buckets to message count clusterizer.

### DIFF
--- a/Source/Model/Analytics/Clusterizer+MessageCount.swift
+++ b/Source/Model/Analytics/Clusterizer+MessageCount.swift
@@ -22,18 +22,28 @@ extension IntegerClusterizer {
         ClusterRange(0, 100),
         ClusterRange(100, 250),
         ClusterRange(250, 500),
-        ClusterRange(500, 1000),
-        ClusterRange(1000, 2000),
-        ClusterRange(2000, 3000),
-        ClusterRange(3000, 4000),
-        ClusterRange(4000, 5000),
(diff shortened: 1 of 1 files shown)


Source/Model/Analytics/Clusterizer+MessageCount.swift
```diff
@@ -22,18 +22,28 @@ extension IntegerClusterizer {
...
+        ClusterRange(75_000, 150_000),
+        ClusterRange(150_000, 300_000),
+        ClusterRange(300_000, 500_000),
+        ClusterRange(500_000, 1_000_000),
+        ClusterRange(1_000_000, 2_000_000),
+        ClusterRange(2_000_000, 5_000_000),
+        ClusterRange(5_000_000, 10_000_000),
+        ClusterRange(10_000_000, 50_000_000),
+        ClusterRange(50_000_000, 100_000_000),
+        ClusterRange(100_000_000, 1_000_000_000)
```
vytis: That should do the job:)

vytis: Ten messages per second for three years and we have it, no biggie

mikeger: We can also figure out a way to autogenerate the buckets, for example by always adding 1/2 of the previous max, or some other algorithm.

PR #250, opened 2017-05-10
Add first batch of model classes

daehn:
# What's in this PR?

* Add first batch of model classes. These are not final yet as this is an intermediary PR, but these classes and helpers are needed in order to start working for example on the `TeamDownloadStrategy`.


Source/Model/Conversation/Team.swift
```diff
@@ -0,0 +1,55 @@
...
+public class Team: ZMManagedObject {
+
+    @NSManaged public var conversations: Set<ZMConversation>?
+    @NSManaged public var members: Set<Member>?
+    @NSManaged public var name: String?
+
+    @NSManaged private var remoteIdentifier_data: Data?
+
+    public var remoteIdentifier: UUID? {
+        get { return remoteIdentifier_data.flatMap { NSUUID(uuidBytes: $0.withUnsafeBytes(UnsafePointer<UInt8>.init)) } as UUID? }
```
daehn: I'm not really happy with this, it would be easier to store the `UUID` as a `String`. However, we will be able to reuse couple of helper methods from the `ZMManagedObject` superclass if the `remoteIdentifier_data` field is present (e.g. `fetch(withRemoteIdentifier:in:`).

Source/Model/Conversation/Team.swift
```diff
@@ -0,0 +1,55 @@
...
+        return "Team"
+    }
+
+    public override static func isTrackingLocalModifications() -> Bool {
+        return false
+    }
+
+    @objc(fetchOrCreateTeamWithRemoteIdentifier:createIfNeeded:inContext:)
+    public static func fetchOrCreate(with identifier: UUID, _ create: Bool, in context: NSManagedObjectContext) -> Team? {
+        precondition(!create || context.zm_isSyncContext, "Needs to be called on the sync context")
```
typfel: 👍

Source/Model/User/ZMUser.m
```diff
@@ -179,6 +180,7 @@ - (void)awakeFromInsert;
 @dynamic clients;
 @dynamic handle;
 @dynamic addressBookEntry;
+@dynamic members;
```
typfel: This reads weird to me. Something like `memberships` sounds more logical to me.

daehn: @typfel updated!

PR #251, opened 2017-05-11
Fix: connection list ordering issue after accepting connection

vytis:
After accepting incoming connection the conversation doesn't have any messages and its `lastModifiedDate` is `1970-01-01`. We use this for sorting the conversation list and it then goes to the absolute bottom.
The fix is to use connection date as `lastModifiedDate`.

### DIFF
--- a/WireDataModel.xcodeproj/xcshareddata/xcschemes/WireDataModel.xcscheme
+++ b/WireDataModel.xcodeproj/xcshareddata/xcschemes/WireDataModel.xcscheme
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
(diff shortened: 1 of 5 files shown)


WireDataModel.xcodeproj/xcshareddata/xcschemes/WireDataModel.xcscheme
```diff
@@ -26,7 +26,8 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "YES">
+      shouldUseLaunchSchemeArgsEnv = "YES"
+      codeCoverageEnabled = "YES">
```
typfel: Intentional?

vytis: Yes, it seems that DataModel didn't have it set, most other frameworks do.

PR #252, opened 2017-05-11
Add functions to create conversations within teams

daehn:
# What's in this PR?

* Add functions to create conversations.
* Add a couple of helper functions.


Source/Model/UserClient/UserClient.swift
```diff
@@ -200,7 +200,7 @@ public class UserClient: ZMManagedObject, UserClientType {
     
     /// Resets the session between the client and the selfClient
     /// Can be called several times without issues
-    public func resetSession() {
+    public func resetSession(in team: Team?) {
```
daehn: Not sure about this, as there will only be one session per user this is actually not needed and we should figure out which conversation it is inside this method.

Source/Model/Conversation/ZMConversation.m
```diff
@@ -1088,14 +1090,61 @@ + (instancetype)conversationWithRemoteID:(NSUUID *)UUID createIfNeeded:(BOOL)cre
     return nil;
 }
 
-+ (instancetype)insertGroupConversationIntoManagedObjectContext:(NSManagedObjectContext *)moc withParticipants:(NSArray *)participants;
++ (instancetype)fetchOrCreateTeamConversationInManagedObjectContext:(NSManagedObjectContext *)moc withParticipant:(ZMUser *)participant team:(Team *)team
```
mikeger: Should it be `... inTeam:` ?

mikeger: Ah, you have `fetchOrCreateTeam` in the name, I guess then it's not necessary.

Source/Model/User/ZMUser.m
```diff
@@ -255,9 +255,15 @@ - (NSString *)initials
     return personName.initials ?: @"";
 }
 
-- (ZMConversation *)oneToOneConversation
```
mikeger: I guess this would break the API, should we leave the old method be too?

mikeger: I have no strong opinion about that, the other methods require team parameter now as well. May be we should mark them as nullable parameters in the headers?

daehn: I thought about it, but it might make sense to break the API here to enforce to change it everywhere on the UI.

daehn: Yeah, I will mark them as `nullable` and also remove the `resetSession(in:)`, as it shouldn't be needed there.

daehn: @mikeger updated!

PR #253, opened 2017-05-12
Parse transport response to create team, permissions and members

daehn:
# What's in this PR?

* Parse transport response to create `Team`, `Permissions` and `Member`.
* Add `TeamType` protocol.

### DIFF
--- a/Source/Model/Conversation/Team.swift
+++ b/Source/Model/Conversation/Team.swift
@@ -18,20 +18,20 @@
 
 
 public protocol TeamType {
-    var conversations: Set<ZMConversation>? { get }
+
+    var conversations: Set<ZMConversation> { get }
     var name: String? { get }
     var teamPictureAssetKey: String? { get }
-    var isActive: Bool { get }
+    var isActive: Bool { get set }
+    var remoteIdentifier: UUID? { get }
 
-    func guests() -> Set<ZMUser>
-    func allUsers() -> Set<ZMUser>
 }
 
 
 public class Team: ZMManagedObject, TeamType {
 
-    @NSManaged public var conversations: Set<ZMConversation>?
-    @NSManaged public var members: Set<Member>?
+    @NSManaged public var conversations: Set<ZMConversation>
+    @NSManaged public var members: Set<Member>
     @NSManaged public var name: String?
     @NSManaged public var teamPictureAssetKey: String?
     @NSManaged public var isActive: Bool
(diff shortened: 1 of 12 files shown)


daehn:
There will be a couple of changes to the API for checking if a user is a guest, I will add them to this PR.

Source/Model/Conversation/Team.swift
```diff
@@ -17,7 +17,18 @@
 //
 
 
-public class Team: ZMManagedObject {
+public protocol TeamType {
+    var conversations: Set<ZMConversation>? { get }
+    var name: String? { get }
```
mikeger: Just noticed, team name is optional, any particular reason? Can be empty string?

daehn: It can be nil when we insert the conversation and did not fetch it yet.

mikeger: Right, but we can pre-fill it with empty string, a bit annoying to have 2 possible negative options to cater for: empty and nil. Same applies for `ZMUser.teams` and other relations, I think CoreData returns empty set when the relation is empty, not nil.

mikeger: Reference: http://stackoverflow.com/questions/21418139/coredata-optional-to-many-relationships-can-never-be-nil

daehn: For `String` it should be possible to use `isEmpty`, which already caters for both of these scenarios. I thought about default values, but checked how we are doing this for other entities were we don't seem to use them.

daehn: Yeah, might be different for relations, there we might get away with adjusting the signature if that's the case.

mikeger: Thanks! Also small request to expose some kind of team id for `TeamType`, so I can compare teams.

mikeger: Also I believe `var isActive: Bool { get }` should be get + set ?

daehn:
@mikeger @basine please have a look at f8dcb48

Source/Model/Conversation/Team.swift
```diff
@@ -17,10 +17,21 @@
 //
 
 
-public class Team: ZMManagedObject {
+public protocol TeamType {
```
mikeger: One more Wunsch: can we make it `Object`, so assignments would not require `var` stuff? Right now it considers that it can be a struct and requires a mutable version.

daehn: Sure, I assume you mean `NSObjectProtocol`?

daehn:
@mikeger updated!

daehn:
I had to add an Objective-C wrapper to make the permissions option set usable form Objective-C 😭.

PR #256, opened 2017-05-18
Update conversation creation rules and transport data parsing

daehn:
# What's in this PR?

* Add `managed` attribute to `ZMConversation`, indicating if a team conversation is managed or not.
* Update team payload parsing from conversation transport data.
* Create new team conversation if existing one has a user defined name.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Transport.m
+++ b/Source/Model/Conversation/ZMConversation+Transport.m
@@ -35,6 +35,8 @@ static NSString *const ConversationInfoOthersKey = @"others";
 static NSString *const ConversationInfoMembersKey = @"members";
 static NSString *const ConversationInfoCreatorKey = @"creator";
 static NSString *const ConversationInfoTeamKey = @"team";
+static NSString *const ConversationInfoTeamIdKey = @"teamid";
+static NSString *const ConversationInfoManagedTeamKey = @"managed";
 static NSString *const ConversationInfoLastEventTimeKey = @"last_event_time";
 
 NSString *const ZMConversationInfoOTRMutedValueKey = @"otr_muted";
@@ -101,9 +103,9 @@ NSString *const ZMConversationInfoOTRArchivedReferenceKey = @"otr_archived_ref";
         ZMLogError(@"Invalid members in conversation JSON: %@", transportData);
     }
 
-    NSUUID *teamId = [transportData optionalUuidForKey:ConversationInfoTeamKey];
-    if (nil != teamId) {
-        self.team = [Team fetchOrCreateTeamWithRemoteIdentifier:teamId createIfNeeded:YES inContext:self.managedObjectContext];
+    NSDictionary *team = [transportData optionalDictionaryForKey:ConversationInfoTeamKey];
+    if (nil != team) {
+        [self updateTeamWithPayload:team];
     }
 }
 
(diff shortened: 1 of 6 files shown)


Source/Model/Conversation/ZMConversation+Transport.m
```diff
@@ -128,6 +130,19 @@ - (void)updateMembersWithPayload:(NSDictionary *)members
...
 
+- (void)updateTeamWithPayload:(NSDictionary *)payload
+{
+    NSUUID *teamId = [payload optionalUuidForKey:ConversationInfoTeamIdKey];
+    if (nil != teamId) {
+        self.team = [Team fetchOrCreateTeamWithRemoteIdentifier:teamId createIfNeeded:YES inContext:self.managedObjectContext];
+
+        NSNumber *managed = [payload optionalNumberForKey:ConversationInfoManagedTeamKey];
+        if (nil != managed) {
+            self.managed = managed.boolValue;
```
vytis: Is there a default value for `Conversation.managed`? What happens if this key is missing?

Source/Model/Conversation/ZMConversation.m
```diff
@@ -111,6 +111,7 @@
 static NSString *const LastReadEventIDDataKey = @"lastReadEventID_data";
 
 static NSString *const TeamKey = @"team";
+static NSString *const TeamManagedKey = @"managed";
```
vytis: Maybe better to add this to `.swift` file and use `#keyPath` to be on the safe side?

Tests/Source/Model/Conversation/ZMConversationTests+Transport.m
```diff
@@ -93,7 +96,10 @@ - (NSDictionary *)payloadForMetaDataOfConversation:(ZMConversation *)conversatio
                                       },
                               @"type" : @(conversationType),
                               @"id" : [conversation.remoteIdentifier transportString],
-                              @"team": [teamID transportString] ?: [NSNull null]
+                              @"team": @{
```
vytis: I would think that in case it's not a team conversation the response will be `"team": null` instead of `"team" : { "teamid" : null }`, but we need to double check with backend docs.

daehn: Yes, wasn't sure how it will look in that case, good point!

Source/Model/Conversation/ZMConversation+Transport.m
daehn: There is a default value `NO`.

Source/Model/Conversation/ZMConversation.m
daehn: It's used in this file only at them moment to exclude this key from syncing, see `ignoredKeys` (the other ones are not referring to the name of the Core Data attribute but to the name of the field in the transport response which happens to be the same at the moment).

vytis: Then probably not worth exposing to outside

daehn: Generally it would make sense to move all of these constants into a swift file and either:
1. Make them accessible through a wrapper object with static functions
2. Move this whole file to Swift 🎉

Tests/Source/Model/Conversation/ZMConversationTests+Transport.m
daehn: Actually I will add a test for this case.

daehn: @vytis updated!

PR #257, opened 2017-05-19
Correct picture keys

daehn:
# What's in this PR?

* Add team `pictureId` and `pictureKey` fields to model schema for consistency.
* Refetch team when receiving when getting it through a conversation update (see the comment).

### DIFF
--- a/Source/Model/Conversation/Team.swift
+++ b/Source/Model/Conversation/Team.swift
@@ -59,14 +59,16 @@ public class Team: ZMManagedObject, TeamType {
         return false
     }
 
-    @objc(fetchOrCreateTeamWithRemoteIdentifier:createIfNeeded:inContext:)
-    public static func fetchOrCreate(with identifier: UUID, create: Bool, in context: NSManagedObjectContext) -> Team? {
+    @objc(fetchOrCreateTeamWithRemoteIdentifier:createIfNeeded:inContext:created:)
+    public static func fetchOrCreate(with identifier: UUID, create: Bool, in context: NSManagedObjectContext, created: UnsafeMutablePointer<Bool>?) -> Team? {
         precondition(!create || context.zm_isSyncContext, "Needs to be called on the sync context")
         if let existing = Team.fetch(withRemoteIdentifier: identifier, in: context) {
+            created?.pointee = false
             return existing
         } else if create {
             let team = Team.insertNewObject(in: context)
             team.remoteIdentifier = identifier
+            created?.pointee = true
             return team
         }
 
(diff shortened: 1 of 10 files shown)


Source/Model/Conversation/ZMConversation+Transport.m
```diff
@@ -135,6 +135,10 @@ - (void)updateTeamWithPayload:(NSDictionary *)payload
     NSUUID *teamId = [payload optionalUuidForKey:ConversationInfoTeamIdKey];
     if (nil != teamId) {
         self.team = [Team fetchOrCreateTeamWithRemoteIdentifier:teamId createIfNeeded:YES inContext:self.managedObjectContext];
+        // FIXME: Do we really need to refetch the team here?
+        // If we are added to a conversation in a team than we should have gotten the
+        // team creation update event and fetched the team before.
+        self.team.needsToBeUpdatedFromBackend = YES;
```
typfel: So we should add question about this in architecture repo?

daehn: Yeah, that might make sense. Another solution to this might be to have an `inout` `BOOL` into the `Team` `fetchOrCreate` method and only set this flag if the team has been newly created, which should never happen at this point, but would make it more error resistant.

daehn: Setting this flag here makes this independent of the backend implementation and as the update event API is not 100% stable yet I would leave it for now or go with the `BOOL *` approach, what do you think @typfel?

typfel: Was wondering if we could use the `isInserted` property on `NSManagedObject` here.

daehn:
@typfel updated!

Source/Model/Conversation/Team.swift
```diff
@@ -59,14 +59,16 @@ public class Team: ZMManagedObject, TeamType {
         return false
     }
 
-    @objc(fetchOrCreateTeamWithRemoteIdentifier:createIfNeeded:inContext:)
-    public static func fetchOrCreate(with identifier: UUID, create: Bool, in context: NSManagedObjectContext) -> Team? {
+    @objc(fetchOrCreateTeamWithRemoteIdentifier:createIfNeeded:inContext:created:)
+    public static func fetchOrCreate(with identifier: UUID, create: Bool, in context: NSManagedObjectContext, created: UnsafeMutablePointer<Bool>?) -> Team? {
```
typfel: Why is not not an `inout` property? ObjectiveC compatibility?

daehn: @typfel yes, unfortunately `inout` doesn't work with Objective-C 😭

daehn:
@vytis @typfel I updated this PR with the `Permissions` bit mask changes (see https://github.com/wireapp/wire-server/pull/12/files#diff-60164b89c380bba0ec9bfaa7bf6356fbR248).

PR #258, opened 2017-05-19
Inject optional team parameter into all conversation list methods

daehn:
# What's in this PR?

* Inject optional team parameter into all conversation list methods.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Predicates.swift
+++ b/Source/Model/Conversation/ZMConversation+Predicates.swift
@@ -18,6 +18,17 @@
 
 import Foundation
 
+fileprivate extension NSPredicate {
+
+    func predicate(for team: Team?) -> NSCompoundPredicate {
+        return .init(andPredicateWithSubpredicates: [
+            ZMConversation.predicateForConversations(in: team),
+            self
+        ])
+    }
+
+}
+
 
 extension ZMConversation {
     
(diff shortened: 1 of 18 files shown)


Source/Model/Conversation/ZMConversation+Predicates.swift
```diff
@@ -18,45 +18,74 @@
...
+    @objc(predicateForConversationsInTeam:)
+    class func predicateForConversations(in team: Team?) -> NSPredicate {
+        if let team = team {
+            return .init(format: "%K == %@", #keyPath(ZMConversation.team), team)
+        }
+
+        return .init(format: "%K == NULL", #keyPath(ZMConversation.team))
+    }
+
+    @objc(predicateForPendingConversationsInTeam:)
```
daehn: Not sure if pending conversations in a team actually make sense as there are no connections and connections to guest will have to be made before adding them to a team conversation. We could also make the conversation list for pending conversations optional and not create it at all in a case a nonnull Team is provided.

typfel: Shouldn't the performance hit for this be pretty minimal?

daehn: Yeah, I don't think the performance is a problem. Was more thinking if this makes sense semantically, as there should never be any pending conversations when there is a team.

PR #262, opened 2017-05-23
Add observation for teams

basine:
Team changes (members added / removed, name change) can be observed by registering as a `TeamObserver`.

Team insertions and deleted can be observed by registering an observer on the selfUser. Insertion & deletions should cause a `UserChangeInfo` with `teamsChanged` set to true.

I also cleaned up the observer tests a bit because we had different methods doing the same thing but being more or less reliable at doing that :-/

### DIFF
--- a/Source/Notifications/ObjectObserverTokens/TeamChangeInfo.swift
+++ b/Source/Notifications/ObjectObserverTokens/TeamChangeInfo.swift
@@ -1,0 +1,103 @@
+//
+// Wire
+// Copyright (C) 2017 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+import WireSystem
+
+extension Team : ObjectInSnapshot {
+    
+    static public var observableKeys : Set<String> {
+        return Set([#keyPath(Team.name),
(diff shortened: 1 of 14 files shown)


Source/Notifications/ObjectObserverTokens/TeamChangeInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+    
+    public var notificationName : Notification.Name {
+        return .TeamChange
+    }
+}
+
+
+@objc public class TeamChangeInfo : ObjectChangeInfo {
+    
+    static let MembersChangedKey = "membersChanged"
```
daehn: Maybe these could also be `#keyPath(TeamChangeInfo. membersChanged)`

Source/Notifications/ObjectObserverTokens/TeamChangeInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+
+@objc public protocol TeamObserver : NSObjectProtocol {
+    func teamDidChange(_ changeInfo: TeamChangeInfo)
+}
+
+
+extension TeamChangeInfo {
+    
+    // MARK: Registering UserObservers
+    /// Adds an observer for the user if one specified or to all ZMUsers is none is specified
```
daehn: The comment is still referring to `ZMUser`

Source/Notifications/ObjectObserverTokens/TeamChangeInfo.swift
```diff
@@ -0,0 +1,103 @@
...
+
+
+import Foundation
+import WireSystem
+
+extension Team : ObjectInSnapshot {
+    
+    static public var observableKeys : Set<String> {
+        return Set([#keyPath(Team.name),
+                    #keyPath(Team.members)])
```
mikeger: Should it also observe the active state, i.e. `isActive` ?

daehn: @mikeger depends if this is needed on the UI, `isActive` is only changed by the UI so it should already know. But maybe it still makes sense e.g. when needing to update a different control in reaction to changing the teams in another one.

mikeger: I would vote to have it observed, then we don't have to build another observer system in the UI ;)

PR #264, opened 2017-05-24
Add logic to check if the self user is guest in a conversation

daehn:
# What's in this PR?

* We need to know if the self user is a guest in a conversation.
* We now set the `teamRemoteIdentifier` on the conversation in addition to setting the `team` relation, this is needed as we will delete the team when we receive a `403` from the backend when fetching `/teams/{id}` for said team but are only a guest. In this case we will now nil the relation but keep the id to know that we are only a guest.


vytis:
The tests fail though, anything related or a flaky one?

daehn:
@vytis checking

daehn:
@vytis fixed!

PR #268, opened 2017-05-26
Exclude self user from team member search

typfel:
Also selfUser was created twice in the `ZMConversationTestsBase`


Source/Model/Conversation/Team.swift
```diff
@@ -98,7 +98,11 @@ extension Team {
     
     public func members(matchingQuery query: String) -> [Member] {
         guard let matchingName = NSPredicate(formatDictionary: ["normalizedName" : "%K MATCHES %@"], matchingSearch: query) else { return [] }
-        return members.filter({ matchingName.evaluate(with: $0.user) })
+        return members.filter({ member in
+            guard let user = member.user else { return false }
+            
+            return matchingName.evaluate(with: user) && !user.isSelfUser
```
daehn: If you reverse the order it doesn't have to evaluate the predicate for the self user as it can return early.

PR #270, opened 2017-05-26
Fix non-unique conversation list identifiers

daehn:
# What's in this PR?

* The `debugDescription` was also used as a unique identifier. We now append the team Id if we have any.


daehn:
I'm checking the tests and will merge once all ✅.

PR #273, opened 2017-05-29
Start UI: Sorting team members by name

vytis:
We didn't sort the search results by anything and ordering was random.


typfel:
@vytis  a test for the sort order maybe?

vytis:
Maybe! 😉

vytis:
@typfel added a test

PR #275, opened 2017-05-30
Add `setMemberPermission` Permissions value

daehn:
# What's in this PR?

* Add `setMemberPermission` Permissions value (see https://github.com/wireapp/architecture/pull/26).

### DIFF
--- a/Source/Model/Conversation/Permissions.swift
+++ b/Source/Model/Conversation/Permissions.swift
@@ -26,22 +26,23 @@ public struct Permissions: OptionSet {
     }
 
     // MARK: - Base Values
-    public static let createConversation       = Permissions(rawValue: 0x001)
-    public static let deleteConversation       = Permissions(rawValue: 0x002)
-    public static let addTeamMember            = Permissions(rawValue: 0x004)
-    public static let removeTeamMember         = Permissions(rawValue: 0x008)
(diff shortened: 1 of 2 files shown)


Source/Model/Conversation/Permissions.swift
```diff
@@ -26,22 +26,23 @@ public struct Permissions: OptionSet {
...
-    public static let removeTeamMember         = Permissions(rawValue: 0x008)
-    public static let addConversationMember    = Permissions(rawValue: 0x010)
-    public static let removeConversationMember = Permissions(rawValue: 0x020)
-    public static let getBilling               = Permissions(rawValue: 0x040)
-    public static let setBilling               = Permissions(rawValue: 0x080)
-    public static let setTeamData              = Permissions(rawValue: 0x100)
-    public static let getMemberPermissions     = Permissions(rawValue: 0x200)
-    public static let getTeamConversations     = Permissions(rawValue: 0x400)
-    public static let deleteTeam               = Permissions(rawValue: 0x800)
+    public static let createConversation       = Permissions(rawValue: 0x0001)
```
vytis: Maybe it's better to use left shift instead of hardcoding the values?

mikeger: Good point

PR #276, opened 2017-05-30
Fix and move canAddUsers check into data model

daehn:
# What's in this PR?

* Fix and move canAddUsers check into data model, add tests.
* The check is currently in the UI project in an extension on `ZMUser`, but makes more sense to be in `wire-ios-data-model`, also the check didn't take the result of `isGuest(in:)` into account.

### DIFF
--- a/Source/Model/User/ZMUser+Teams.swift
+++ b/Source/Model/User/ZMUser+Teams.swift
@@ -42,13 +42,13 @@ public extension ZMUser {
 
     @objc(canAddUserToConversation:)
     public func canAddUser(to conversation: ZMConversation) -> Bool {
-        guard !isGuest(in: conversation) else { return false }
+        guard !isGuest(in: conversation), conversation.isSelfAnActiveMember else { return false }
         return conversation.team.flatMap(permissions)?.contains(.addConversationMember) ?? true
     }
 
     @objc(canRemoveUserFromConversation:)
     public func canRemoveUser(from conversation: ZMConversation) -> Bool {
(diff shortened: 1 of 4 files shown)


Source/Model/User/ZMUser+Teams.swift
```diff
@@ -40,6 +40,18 @@ public extension ZMUser {
         return membership(in: team)?.permissions
     }
 
+    @objc(canAddUserToConversation:)
+    public func canAddUser(to conversation: ZMConversation) -> Bool {
+        guard !isGuest(in: conversation) else { return false }
```
mikeger: Would it work if user is not an active participant in the conversation? Also if conversation has team but no permissions this would return true, but I think it should return false.

daehn: If a member does not have any permissions then something went wrong somewhere else so this should never happen (in theory), regarding active participants I need to double check, AFAIK the conversation should be deleted in a team context when the self member gets removed.

mikeger: May be the conversation is deleted in team, but in personal space you should be still not able to add / remove participants when you are not an active member.

daehn: But then `isGuest` will return true

daehn: But I can add an additional check to make this more explicit

daehn:
@mikeger updated!

PR #278, opened 2017-05-31
Support for observing when a conversation exists remotely

typfel:
Observe `remoteIdentifier` on ZMConversation and add a `onExistsRemotely`
helper method for executing a block when a conversation exists remotely.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+ObserverHelper.swift
+++ b/Source/Model/Conversation/ZMConversation+ObserverHelper.swift
@@ -1,0 +1,50 @@
+//
+// Wire
+// Copyright (C) 2017 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


Source/Model/Conversation/ZMConversation+ObserverHelper.swift
```diff
@@ -0,0 +1,50 @@
...
+        if filter(changeInfo) {
+            block()
+        }
+    }
+    
+}
+
+public extension ZMConversation {
+    
+    public func onExistsRemotely(_ block : @escaping () -> Void) -> NSObjectProtocol? {
```
vytis: It seems that the name `existsRemotely` makes it difficult to make combinations with it that sounds nice - `onExistsRemotely` and `existsRemotelyChanged`. 
Can the message loose the remote identifier? If not then maybe something like 
`onExistsRemotely` -> `whenCreatedRemotely` and `existsRemotelyChanged` -> `createdRemotely` would work better?

typfel: Yes, that sounds nicer but I like the short `on` prefix for blocks. How about `onCreatedRemotely` and `createdRemotely`

vytis: yeah, much clearer!

PR #280, opened 2017-05-31
Respect to one relationships in core date snapshots

daehn:
# What's in this PR?

* Add a separate dictionary to `Snapshot` to store the presence of to-one relations.

# Issue

When a user is added to a team conversation as a guest, the following sequence of events is happening:

1. We receive the event for adding the conversation.
2. We create the conversation from the payload, which contains the `team_id`.
3. We create the team locally and mark it for fetching.
→ The conversation is not yet shown in the general conversation list, since it has a team.
4. We fetch the team and receive a 403 indicating that the access is forbidden.
5. We delete the team.
6. CoreData is sending a save notification in which the conversation is faulted, and our snapshot does not contain the information about the one-to-one relation `team` on the conversation, so we are ignoring the change.
7. The conversation does not appear unless the list is reloaded / app restarted / a message in the conversation is received.

# Solution

Track one-to-one relationship state in our snapshots.

### DIFF
--- a/Tests/Source/Model/Observer/ConversationListObserverTests.swift
+++ b/Tests/Source/Model/Observer/ConversationListObserverTests.swift
@@ -994,7 +994,7 @@ class ConversationListObserverTests : NotificationDispatcherTestBase {
         XCTAssertEqual(movedIndexes(first), [])
     }
     
-    func testThatItNotifiesObserversWhenAConversationsTeamDeletedSoItNowMatch() {
+    func testThatItNotifiesObserversWhenAConversationsTeamDeletedSoConversationAppears() {
         // NB! This test is checking the general behaviour of the conversation list observer, however in real life
         // the `conversation` object can possibly be faulted. It is not possible to implement the desired behaviour 
         // currently in the test setup.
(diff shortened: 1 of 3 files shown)


Tests/Source/Model/Observer/ConversationListObserverTests.swift
```diff
@@ -993,6 +993,53 @@ class ConversationListObserverTests : NotificationDispatcherTestBase {
         XCTAssertEqual(first.updatedIndexes, IndexSet())
         XCTAssertEqual(movedIndexes(first), [])
     }
+    
+    func testThatItNotifiesObserversWhenAConversationsTeamDeletedSoItNowMatch() {
```
typfel: This test name is a bit unclear to me. We are testing that the conversation list observer is notified when a team is deleted? What is the the `SoItNowMatch` referring to?

mikeger: Exactly, so per default the team is there, and list is empty, then team is deleted and conversation is coming into the list "soItNowMatch". I would think on the wording to make it more clear, I guess something along "ConversationAppears"

vytis:
Just a minor point, but maybe adding a more descriptive explanation in PR would help to understand the implications and the reasoning behind this change for someone who was not directly involved into solving this :) I wonder why did we notice the 1-to-1 relationships observation behaviour only now? All other things except teams are 1-to-many?

mikeger:
@vytis yep, sorry, forgot to update the description.

mikeger:
@vytis updated

vytis:
Thanks a bunch, much clearer now!

PR #282, opened 2017-06-01
Enable search by user handle

vytis:
It seems that local search by user handle was not working, it was searching only by `normalizedName`. Updated predicates to handle this situation and re-using it in team members search.

### DIFF
--- a/Source/Model/User/ZMUser+Internal.h
+++ b/Source/Model/User/ZMUser+Internal.h
@@ -57,23 +57,6 @@ extern NSString * __nonnull const ZMUserActiveConversationsKey;
 
 + (nonnull NSOrderedSet <ZMUser *> *)usersWithRemoteIDs:(nonnull NSOrderedSet <NSUUID *>*)UUIDs inContext:(nonnull NSManagedObjectContext *)moc;
 
-/// @method predicateForConnectedUsersWithSearchString:
-/// Retrieves users with name or email matching search string, having ZMConnectionStatusAccepted connection statuses.
-/// @param searchString - a predicate to search users
-+ (nonnull NSPredicate *)predicateForConnectedUsersWithSearchString:(nonnull NSString *)searchString;
-
-/// @method predicateForConnectedNonBotUsers
-/// Retrieves all users (excluding bots), having ZMConnectionStatusAccepted connection statuses.
-+ (nonnull NSPredicate *)predicateForConnectedNonBotUsers;
-
-/// @method predicateForUsersWithSearchString:connectionStatusInArray:
-/// Retrieves users with name or email matching search string, having one of given connection statuses.
-/// @param searchString - a predicate to search users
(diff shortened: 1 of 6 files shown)


Source/Model/User/ZMUser+Internal.h
```diff
@@ -57,23 +57,6 @@ extern NSString * __nonnull const ZMUserActiveConversationsKey;
...
-
-/// @method predicateForConnectedNonBotUsers
-/// Retrieves all users (excluding bots), having ZMConnectionStatusAccepted connection statuses.
-+ (nonnull NSPredicate *)predicateForConnectedNonBotUsers;
-
-/// @method predicateForUsersWithSearchString:connectionStatusInArray:
-/// Retrieves users with name or email matching search string, having one of given connection statuses.
-/// @param searchString - a predicate to search users
-/// @param connectionStatusArray - an array of connections status of the users. E.g. for connected users it is @[@(ZMConnectionStatusAccepted)]
-+ (nonnull NSPredicate *)predicateForUsersWithSearchString:(nonnull NSString *)searchString
```
typfel: Shouldn't the `*.m` implementation be deleted as well?

vytis: 🤦‍♂️  good catch!

Source/Model/User/ZMUser.swift
```diff
@@ -78,6 +79,83 @@ extension ZMUser {
...
+                var withoutAt = query
+                withoutAt.remove(at: query.startIndex)
+                normalizedHandle = withoutAt
+            } else {
+                // User regular normalized query otherwise
+                normalizedHandle = normalizedQuery
+            }
+            
+            let handlePredicate = NSPredicate(format: "%K BEGINSWITH %@", #keyPath(ZMUser.handle), normalizedHandle)
+            allPredicates.append([namePredicate, handlePredicate].flatMap {$0})
```
mikeger: Why `flatMap` is necessary? I think that they are not optional.

vytis: The `handlePredicate` is force-unwrapped from the obj-c land. Was thinking to use regular predicate but this seems to have lots of search-smartness inside and actually sprints query into words and matches them separately.

Source/Model/User/ZMUser+Internal.h
typfel: Were there also tests for the old implementation which can be deleted?

vytis: Doesn't seem like there were any tests for the predicates directly, only for `ZMSearchDirectory` in SE.

vytis: Maybe we should start using some sort of code coverage tool to catch these sort of cases...

mikeger:
@vytis I forgot to mention, let's not merge new stuff yet until we have our RC done, otherwise we are risking not to end it ever :)

PR #283, opened 2017-06-07
Finding out if user is a guest in any team

vytis:
Added a way to fetch all teams where the user is added as a guest to team conversations.

### DIFF
--- a/Source/Model/Conversation/Team.swift
+++ b/Source/Model/Conversation/Team.swift
@@ -81,7 +81,6 @@ public enum TeamError: Error {
     case insufficientPermissions
 }
 
-
 extension Team {
 
     public func addConversation(with participants: Set<ZMUser>) throws -> ZMConversation? {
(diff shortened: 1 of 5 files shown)


Source/Model/Conversation/Team.swift
```diff
@@ -108,4 +107,17 @@ extension Team {
         })
     }
     
+    public static func predicateTeamsWithGuestUserInAnyConversation(guestUser: ZMUser) -> NSPredicate {
+        let notInThisTeam = NSPredicate(format: "NOT (SELF IN %@)", guestUser.teams)
+        let participantInAnyConversation = NSPredicate(format: "SUBQUERY(%K, $conversation, ANY $conversation.%K == %@).@count > 0", #keyPath(Team.conversations), #keyPath(ZMConversation.otherActiveParticipants), guestUser)
```
typfel: `$conversation.otherActiveParticipants == %@` will evaluate to true if the guest is a member in the set? Weird syntax :)

vytis: Yeah so the way it works that subquery will loop on every item in the first param (in this case `Team.conversations`) and then collect all items that evaluate true on last param. However the last param needs to traverse one more to-many relationship and this is done with `ANY` - it loops through all `ZMConversation.otherActiveParticipants` and returns true is any is equal to user.

It is a bit weird mainly because we need to traverse two to-many relationships: Team->Conversations and Conversation-> otherActiveParticipants.

typfel: Hmm so it's not possible to use the `otherActiveParticipants CONTAINS {guest}`?

mikeger: I think it should be possible to do `"SUBQUERY(%K, $conversation, %@ IN $conversation.%K).@count > 0"` as Jacob suggested, but it should not really change the query generated.

mikeger: So the change is only for the readability. But if it works, then may be it is possible to do `ANY %@ IN conversations.otherActiveParticipants`

mikeger: (Sorry for the noise, I think the current is already good, just interested in more details how NSPredicate can handle it for educational purposes)

vytis: Updated with `IN` instead of complicated `ANY`, but seems like not really possible to simplify any further...

PR #285, opened 2017-06-08
Consider team group conversations with one participant 1-to-1

typfel:
We want to consider team group conversations with only 2 participants to be one-to-one. This means supporting the same features (delivery recipes, video calling) and design.

see: https://github.com/wireapp/architecture/blob/master/teams/clients.md#one-to-one-conversation-inside-teams

### DIFF
--- a/Source/Model/Conversation/ZMConversation.m
+++ b/Source/Model/Conversation/ZMConversation.m
@@ -124,6 +124,7 @@ const NSUInteger ZMConversationMaxTextMessageLength = ZMConversationMaxEncodedTe
 
 @property (nonatomic) NSString *normalizedUserDefinedName;
 @property (nonatomic) ZMConversationType conversationType;
+@property (nonatomic, readonly) ZMConversationType internalConversationType;
 
 @property (nonatomic) NSDate *tempMaxLastReadServerTimeStamp;
 @property (nonatomic) NSMutableOrderedSet *unreadTimeStamps;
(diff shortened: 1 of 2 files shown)


Source/Model/Conversation/ZMConversation.m
```diff
@@ -473,6 +481,24 @@ - (void)setUserDefinedName:(NSString *)aName {
...
+        conversationType = ZMConversationTypeOneOnOne;
+    }
+    
+    [self didAccessValueForKey:ZMConversationConversationTypeKey];
+    return conversationType;
+}
+
+- (ZMConversationType)internalConversationType
+{
+    return (ZMConversationType)[[self primitiveConversationType] shortValue];
```
marcoconti83: Is `primitiveConversationType` a new CoreData property? I don't see it being added to the CoreData model in this PR.
Here you call `[self willAccessValueForKey:ZMConversationConversationTypeKey]` but the actual CoreData value for `ZMConversationConversationTypeKey` is never accessed. Unless this points to `primitiveConversationType`.

typfel: `primitiveConversationType` is not a new CoreDara property but is a getter generated by CoreData for accessing the internal storage of `conversationType` (see https://developer.apple.com/documentation/coredata/nsmanagedobject/1506728-primitivevalueforkey)

on ZMConversation.m:145 we declare several other primitive properties which are used in a similar manner.

marcoconti83: Thx for explaining, maybe the `willAccessValue/didAccessValue` should be inside this method then.

marcoconti83:
Looks good now!

PR #288, opened 2017-06-15
[WIP] Unified conversation list

daehn:
# What's in this PR?

* Add a new model version (2.31.0) which removes the `memberships` to-many relation as a user will only be able to be in one team (multi account setup).
* This also includes some changes which had to be made due to said relationship type change.
* Remove `isActive` flag on a team, this will be handles differently in the future as there will only be one team per account.
* Remove the team parameter from a couple of helper methods as there is only one team if any.
* I did not (yet?) add a convenience accessor to `ZMUserSession` to access the team of the current account, however in the most places this was a change from `ZMuser.selfUser().activeTeam` to `ZMUser.selfUser().team` which wasn't to big of a change and I currently don't see the advantage of adding a relation to the session. If we want this behaviour we can create a computed property which gets the self user of the session and returns its team.

### This PR depends on:
- [x] Corresponding PRs in affected repositories.
- [x] A HotFix to cleanup and delete all local teams. The conversations will still appear in the conversation list, but all users will appear as guests.
- [x] Update tests and logic in `wire-ios-sync-engine`.
- [x] Update tests and logic in `wire-ios-message-strategy`.


Source/Model/User/ZMUser+Teams.swift
```diff
@@ -19,41 +19,37 @@
 
 public extension ZMUser {
 
-    public var hasTeams: Bool {
-        return memberships?.any { $0.team != nil } ?? false
+    public var hasTeam: Bool {
+        return nil != team
+        // TODO: Check membership / or check user session
```
vytis: I guess it's fine to only check this, even if we decide to add a relationship to session it won't invalidate this assumption.

Source/Model/User/ZMUser.m
```diff
@@ -77,7 +77,7 @@
 static NSString *const UserClientsKey = @"clients";
 static NSString *const ReactionsKey = @"reactions";
 static NSString *const AddressBookEntryKey = @"addressBookEntry";
-static NSString *const MembershipsKey = @"memberships";
+static NSString *const MembershipKey = @"membership";
```
vytis: Maybe worth moving this to `ZMUser.swift` and add to a list of other keypaths in there

daehn: @vytis I agree, but I wouldn't do this refactoring with this PR, would be nice to clean up these hard coded Strings in other places as well.

daehn:
@vytis updated the model and removed the v2 profile picture attributes.

PR #292, opened 2017-06-20
Remove V2 calling

typfel:
- Simply`ZMCallState` which is now only used to display the correct conversation list indicator.
- Delete `callParticipants` and `needsToUpdateCallStateFromBackend` from data model


daehn:
@typfel can we merge this one?

typfel:
You can but the SE PR is not quite ready yet

PR #293, opened 2017-06-20
Object grouping by key

mikeger:
# Issue

This is necessary to fetch the objects having the same key in the database, mostly to remove the duplicated entities from the database.


Source/Model/ZMManagedObject+Grouping.swift
```diff
@@ -0,0 +1,93 @@
...
+    } catch let error {
+        fatal("Cannot perform the fetch: \(error)")
+    }
+    
+    return result
+}
+
+extension Array where Element: NSObject {
+    public func group<Key: Hashable>(by keyPath: String) -> [Key: [Element]] {
+        return self.map {
```
daehn: Might be more performant to do the creation and merging in one step here.

E.g. instead of mapping over all objects (which we saw could be a lot) and then reducing in `merge` having one loop in which we check if the key exists and if so append to the array, otherwise create the key entry.

Source/Model/UserClient/UserClient.swift
```diff
@@ -141,9 +141,7 @@ public class UserClient: ZMManagedObject, UserClientType {
         
         let existingClients = user.clients.filter({$0.remoteIdentifier == remoteIdentifier})
         
-        if existingClients.count > 1 {
-            zmLog.error("Detected duplicate clients: \(existingClients.map({ $0.remoteIdentifier! }))")
-        }
+        assert(existingClients.count <= 1, "Detected duplicate clients: \(existingClients.map({ $0.remoteIdentifier! }))")
```
marcoconti83: I'm worried how many production clients will this crash? We should try to have the assertion on insertion, not on fetching.

Source/Model/ZMManagedObject+Grouping.swift
```diff
@@ -0,0 +1,93 @@
...
+            return innerFetchRequest
+        }
+        
+        result = distinctIDs.finalResult!
+            .map {
+                $0[keyPath as NSString]! as! Key
+            }.filter {
+                return try! context.count(for: fetchRequest(for: keyPath, value: $0)) > 1
+            }.map {
+                return TupleKeyArray(key: $0, value: context.fetchOrAssert(request: fetchRequest(for: keyPath, value: $0)) as! [T])
```
daehn: We are potentially creating a lot of fetch requests here, maybe we can create one predicate with a list and have a `IN` check to only hit the persistent store once?

daehn: Couldn't we use the `group(by:)` function here instead of doing a map creating the TupleKeyArray and a subsequent merge?

mikeger: We filter out only duplicated clients, see line 76, so there would be a fetch request per duplicated client, I thought those should not be too much in the database.

mikeger: @daehn I am not following with the `group(by:)`, we get as the result of the filter the array of `[Key]`, so at this point we don't have yet any objects fetched, so we cannot perform the group by. Then in map we fetch for each `key` the list of objects matching it: `[T]`, and we don't need to group at this point, since the objects are already placed in their pairs, like `[Key: [T]]`. 

We can as you proposed create one fetch request with IN condition and then do the `group(by:)`.

Source/Model/ZMManagedObject+Grouping.swift
```diff
@@ -0,0 +1,93 @@
...
+    var key: Key { get }
+    var value: [Value] { get }
+}
+
+public struct TupleKeyArray<Key: Hashable, Value: Any>: TupleKeyArrayType {
+    public let key: Key
+    public let value: [Value]
+}
+
+extension Array where Element: TupleKeyArrayType {
```
marcoconti83: If I understand correctly, this transforms` [(key, value)]` to `[key: [value]]` by grouping the value contained in each tuple by the key contained in each tuple, right? `merge` does not explain exactly what's happening, maybe `groupByKey`?
I suggest adding a comment to make this clear, I had to look up the other definitions to understand this. If I didn't get this right, then it's even more important to add a comment :)

Source/Model/ZMManagedObject+Grouping.swift
```diff
@@ -0,0 +1,93 @@
...
+            var objectsForKey = $0[$1.key] ?? []
+            objectsForKey.append(contentsOf: $1.value)
+            var result = $0
+            result[$1.key] = objectsForKey
+            return result
+        }
+    }
+}
+
+public func findDuplicated<T: ZMManagedObject, Key: Hashable & CVarArg>(in context: NSManagedObjectContext, by keyPath: String) -> [Key: [T]] {
```
daehn: Also this would maybe read nicer at the caller site having it in an extension on `NSManagedObjectContext` instead of being a free function.

mikeger: Good point!

marcoconti83: Is this method used to check how many entries in Core Data have the same value for a given key? Do you need only the count, or do you need the exact values?
If you need only the count, a `GROUP BY` is easier: https://stackoverflow.com/questions/4865686/group-by-equivalent-for-core-data

Source/Model/UserClient/UserClient.swift
daehn: I would suggest we add our own function which only crashes internal builds (and use the same when inserting as well).

Source/Model/ZMManagedObject+Grouping.swift
daehn: We need the objects (or their `NSManagedObjectID`) to delete the duplicates after retrieving them, as I think this will only be used in a HotFix to clean up the database.

mikeger:
@daehn @marcoconti83 I would do the group by to fetch the duplicated clients and then another fetch with IN predicate for getting the objects. I need the objects since I would have to merge the relations.

Source/Model/ZMManagedObject+Grouping.swift
marcoconti83: That hotfix will need to be very comprehensive, to navigate all relationships, e.g.:

- we have 100 copies of a client
- pick the first client (1) as the only client to keep
- go through all the relationships of the 99 that we want to delete
- for each object reached this way, assign the one that we will keep (1) to the inverse relationship
- now you can delete the 99

marcoconti83: It makes sense though, just saying as a reminder :)

Source/Model/UserClient/UserClient.swift
mikeger: I would revert this part and add the new internal-assert function as a separate thing on the appropriate framework level.

mikeger:
@marcoconti83 @daehn updated, please re-review!

mikeger:
@marcoconti83 @basine @vytis @typfel please review

PR #294, opened 2017-06-26
Add lightweight managed object context change observer

daehn:
# What's in this PR?

Add a lightweight managed object context observer `ManagedObjectContextChangeObserver`, which should be used when in the background. For example, when replying to a message from a push notification we want to observe the delivery state of the sent message. If the send operation fails we want to present a "Failed to send message" notification to the user. As our custom observer system is not active while the application is in the background and would be quite expensive to start just for observing one message (due to the snapshot creation), the added observer can be used to get notified when objects in a specified context change.


daehn:
@typfel updated!

PR #296, opened 2017-06-27
Try to fetch user client if it is not available from relation

mikeger:
# Issue

Sometimes we see in the user databases the duplicated clients with the same user ID.

It is possible to reproduce (see `testThatItFetchesUserClient_OtherMOC`).

Solution is to try to fetch the object with the fetch request instead of looking in the relation's set.


Source/Model/UserClient/UserClient.swift
```diff
@@ -145,17 +145,28 @@ public class UserClient: ZMManagedObject, UserClientType {
...
-                return newClient
-            }
-            return nil
+        if let client = existingClients.first {
+            return client
         }
-        return client
+        
+        let fetchRequest = NSFetchRequest<UserClient>(entityName: UserClient.entityName())
+        fetchRequest.predicate = NSPredicate(format: "%K == %@", ZMUserClientRemoteIdentifierKey, remoteIdentifier)
```
daehn: You could set a `fetchLimit` of 1 here as well (see https://github.com/wireapp/wire-ios-data-model/blob/1d5d7d9881032979b74f17851b4785df6b199f38/Source/Model/UserClient/UserClient.swift#L281).

daehn: Would also be nice not to duplicate the logic in `createOrUpdateSelfUserClient` and reuse this method there.

basine: I think you can just call `UserClient.fetch(withRemoteIdentifier: , in:)` which does all this. It's a method on ZMManagedObject

daehn: I think `ZMManagedObject.fetch(withRemoteIdentifier:in:)` expects the remote identifier to be a `UUID` whereas a `UserClient` has a remoteId of type `String`.

mikeger:
@basine I checked and `UserClient.fetch(withRemoteIdentifier: , in:)` is not a good option, since it needs a UUID, whereas `UserClient` is using a string as identifier. I refactored the code to have a single method that creates the fetch request and added the limit.

mikeger:
@basine please re-review

PR #298, opened 2017-06-28
Set a members users remoteIdentifier on the member itself

daehn:
# What's in this PR?

* Add needsToBeUpdatedFromBackend and add persistent data patch migrating identifiers
* Add `needsToBeUpdatedFromBackend` flag to `Member` entity.
* Add `PersistedDataPatch` to set a members users `remoteIdentifier` as the `remoteIdentifier` of the member itself as well.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -9,7 +9,7 @@ GEM
       tzinfo (~> 1.1)
     addressable (2.5.1)
       public_suffix (~> 2.0, >= 2.0.2)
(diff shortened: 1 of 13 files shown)


basine:
@daehn something's up with Danger

daehn:
@basine please re-review, I added a helper method and tests to update member permissions with a response payload (will be used when fetching member permissions from `/teams/{id}/members/{is}`).

PR #302, opened 2017-06-30
revert user changes

basine:
Otherwise the user will end up in a loop 
team -> member -> user


mikeger:
So when it's merged the user's team membership would be updated when we fetch the team's members?

daehn:
@mikeger yes, that's already the case.

PR #303, opened 2017-07-03
Add first account manager scaffolding

daehn:
# What's in this PR?

* Adds `Account` class, which holds information like user identifier, user name, team name and image about an account. This class can be serialised to disk.
* Adds `AccountStore` class which is used to store and retrieve `Accounts`. It stores accounts in files named by their id like this:
```
– Root url passed to init
    – Accounts
         – 47B3C313-E3FA-4DE4-8DBE-5BBDB6A0A14B
         – 0F5771BB-2103-4E45-9ED2-E7E6B9D46C0F
```
* Adds `AccountManager` class which is the entry point for other objects that need to handle with accounts. This internally uses an `AccountStore` and exposes a sorted array of accounts, functions to add and remove accounts, as well as a currently selected account or `nil` if there is none.

### What's not in this PR?

* There is no way to store and retrieve cookies associated with a specific account, yet. In order for this to be implemented in the `AccountManager` the transport session, specifically `ZMPersistentCookieStorage`, has to be adjusted to store multiple cookies (keyed by user identifiers).


Source/Model/AccountStore.swift
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    // MARK: - Private Helper
+
+    /// Loads the urls to all stored accounts.
+    /// - returns: The urls to all accounts stored in this `AccountStore`.
+    private func loadURLs() -> Set<URL> {
+        do {
+            let paths = try fileManager.contentsOfDirectory(atPath: directory.path)
+            return Set<URL>(paths.filter { $0 != ".DS_STORE" }.map(directory.appendingPathComponent))
```
vytis: Do these files exist for iOS as well? Maybe a more robust solution would be to filter by having a parseable `UUID` in name?

daehn: Yeah good point, it's at least not there when checking the simulator contents, but checking for valid `UUID` names is safer anyways.

daehn: @vytis updated!

Source/Model/Account.swift
```diff
@@ -0,0 +1,98 @@
...
+    public override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? Account else { return false }
+        return userName == other.userName
+            && teamName == other.teamName
+            && userIdentifier == other.userIdentifier
+            && imageData == other.imageData
+    }
+
+    public override var hash: Int {
+        return userName.hash ^ userIdentifier.hashValue
```
mikeger: Should it be enough to have a hash on `userIdentifier`?

daehn: Yeah it should, was thinking about this as well.

daehn: @mikeger updated!

Source/Model/Account.swift
```diff
@@ -0,0 +1,98 @@
...
+// MARK: - NSSecureCoding
+
+extension Account: NSSecureCoding {
+
+    public static var supportsSecureCoding = true
+
+    public convenience init?(coder aDecoder: NSCoder) {
+        guard let id = aDecoder.decodeString(forKey: #keyPath(Account.userIdentifier)).flatMap(UUID.init),
+            let name = aDecoder.decodeString(forKey: #keyPath(Account.userName)) else { return nil }
+        self.init(
```
mikeger: On `NSCoding`: It can get slightly tricky with the migration, knowing you are not allowed to rename classes and it is using the swift, may be would be better to use JSON? I am trying to be not too picky, but this one would be used in the future to store the accounts and we have to keep it in mind.

Source/Model/AccountStore.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    /// Creates the `Accounts` subdirectory to store accounts data in.
+    /// Excludes the directory from backend as well as set the correct `FileProtectionType`.
+    private func createDirectoryIfNeeded() {
+        do {
+            guard !fileManager.fileExists(atPath: directory.path) else { return }
+            try fileManager.createDirectory(at: directory, withIntermediateDirectories: true, attributes: nil)
+            try (directory as NSURL).setResourceValue(true, forKey: .isExcludedFromBackupKey)
+            let attributes = [FileAttributeKey.protectionKey: FileProtectionType.completeUntilFirstUserAuthentication]
+            try fileManager.setAttributes(attributes, ofItemAtPath: directory.path)
```
vytis: This should not be needed because we have it set for all files already.

daehn: @vytis updated!

Source/Model/Account.swift
daehn: Renaming classes should be possible when providing a delegate of the archiver, but its easy to convert it to use JSON and I agree that this might make it easier in the future.

daehn: @mikeger updated!

PR #305, opened 2017-07-04
Fix not using normalised query string when stripping leading at sign

daehn:
# What's in this PR?

* We were not using the normalised query string when the user entered a query string with a leading @ symbol. In case the user searched for a username and started uppercase no results were found.


mikeger:
@daehn potentially a test can be added?

daehn:
@mikeger already on it, but wanted to get this in for the release.

PR #308, opened 2017-07-04
Unauthenticated session

vytis:
Small change needed after API update


Source/ManagedObjectContext/ZMTestSession.m
```diff
@@ -85,7 +85,6 @@ - (void)prepareForTestNamed:(NSString *)testName
     
     [ZMPersistentCookieStorage setDoNotPersistToKeychain:!self.shouldUseRealKeychain];
     [self resetUIandSyncContextsAndResetPersistentStore:YES];
-    [ZMPersistentCookieStorage deleteAllKeychainItems];
```
mikeger: Should it be replaced with `deleteAllUserItems` ?

PR #311, opened 2017-07-06
Do not crash when sending the ephemeral to an empty group

mikeger:
- Optional here and there to allow indication of the fact that the payload cannot be generated.


daehn:
@mikeger what happens when we are unable to generate an encrypted message payload? Will we endlessly retry?

mikeger:
@daehn message would expire without sending: https://github.com/wireapp/wire-ios-message-strategy/pull/86/files

basine:
Hmm, I thought we had fixed that before?

mikeger:
Closed in favor of https://github.com/wireapp/wire-ios-data-model/pull/316

PR #314, opened 2017-07-11
Correctly handle device verification for team users

mikeger:
- Check if the users are the members of the same team.


Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -369,7 +369,19 @@ extension ZMConversation {
     }
     
     fileprivate var containsUnconnectedParticipant : Bool {
```
daehn: We might also want to consider if the name is still valid, something like `containsUnconnectedParticipantOrExternalUser` might be more accurate now.

mikeger:
@daehn updated! More tests!

PR #316, opened 2017-07-13
Do not allow sending the ephemeral message to the empty conversation

mikeger:
# Issue

When sending the message to group conversation without the participants the sending is not possible, since there are no receivers (ephemeral is not sent to your own devices).

# Fix

Appending the message is failing in such conversations.

### DIFF
--- a/Tests/Source/Model/Conversation/ZMConversationTests+Ephemeral.swift
+++ b/Tests/Source/Model/Conversation/ZMConversationTests+Ephemeral.swift
@@ -88,7 +88,6 @@ class ZMConversationTests_Ephemeral : BaseZMMessageTests {
     func testThatSendingEphemeralToEmptyConversationIsNotPossible() {
         // given
         let conversation = ZMConversation.insertNewObject(in: uiMOC)
-        conversation.conversationType = .oneOnOne
         conversation.conversationType = .group
 
         // when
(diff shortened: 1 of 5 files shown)


Tests/Source/Model/Conversation/ZMConversationTests+Ephemeral.swift
```diff
@@ -84,6 +84,35 @@ class ZMConversationTests_Ephemeral : BaseZMMessageTests {
         // then
         XCTAssertEqual(conversation.messageDestructionTimeout, 5)
     }
+ 
+    func testThatSendingEphemeralToEmptyConversationIsNotPossible() {
+        // given
+        let conversation = ZMConversation.insertNewObject(in: uiMOC)
+        conversation.conversationType = .oneOnOne
+        conversation.conversationType = .group
```
typfel: Setting it twice? typo?

Tests/Source/Model/Conversation/ZMConversationTests+Ephemeral.swift
```diff
@@ -84,6 +84,35 @@ class ZMConversationTests_Ephemeral : BaseZMMessageTests {
...
+                
+        // then
+        XCTAssertEqual(conversation.canSendEphemeral, false)
+    }
     
+    func testThatSendingEphemeralToNormalConversationIsPossible() {
+        // given
+        let conversation = ZMConversation.insertNewObject(in: uiMOC)
+        conversation.conversationType = .oneOnOne
+        conversation.conversationType = .group
```
typfel: same here

Tests/Source/Model/Conversation/ZMConversationTests+Ephemeral.swift
mikeger: Yeah, my bad!

mikeger:
@typfel please re-review

PR #319, opened 2017-07-19
Fixed leaking CFString

mikeger:
# Issue

The instance of `Unmanaged` must be released.

### DIFF
--- a/Source/Model/Message/V2Asset.swift
+++ b/Source/Model/Message/V2Asset.swift
@@ -24,10 +24,14 @@ import MobileCoreServices
 extension String {
 
     var isGIF: Bool {
-        guard let UTIString = UTTypeCreatePreferredIdentifierForTag(kUTTagClassMIMEType, self as CFString, nil)?.takeUnretainedValue() else { return false }
-        return UTIString == kUTTypeGIF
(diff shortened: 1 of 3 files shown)


Source/Model/Message/V2Asset.swift
```diff
@@ -24,10 +24,14 @@ import MobileCoreServices
 extension String {
 
     var isGIF: Bool {
-        guard let UTIString = UTTypeCreatePreferredIdentifierForTag(kUTTagClassMIMEType, self as CFString, nil)?.takeUnretainedValue() else { return false }
-        return UTIString == kUTTypeGIF
+        guard let UTIStringUnmanaged = UTTypeCreatePreferredIdentifierForTag(kUTTagClassMIMEType, self as CFString, nil) else { return false }
```
typfel: `CFString` is toll free bridged. Shouldn't we be able to hand over the owner ship to ARC.

mikeger:
Changing to `takeRetainedValue()` seems to be sufficient:
```
Basically, if a function name contains the word "Create" or "Copy", use .takeRetainedValue(). If a function name contains the word "Get", use .takeUnretainedValue().
```
Reference: https://stackoverflow.com/questions/28978200/whats-the-difference-between-takeunretainedvalue-and-takeretainedvalue

PR #320, opened 2017-07-21
Properly tear down CoreData stack

marcoconti83:
# Description
The `StorageStack` singleton should be used to create the stack, using the `createManagedObjectContextDirectory:…` method. It will asynchronously return a `ManagedObjectContextDirectory` that contains all the necessary `NSManagedObjectContext`. It will internally take care of migration, waiting for the filesystem to be unlocked if awaiting first unlock, and any other internal operation.

To tear down the stack, call `StoreStack.reset()` after releasing any reference to the managed object contexts. After calling this method, the stack is ready to be initialised again with  `createManagedObjectContextDirectory:…`.

# TODO
- Expose whatever extra property (store URL, ...) is needed somewhere else up the chain
- **TESTING**, tests are not compiling

# Notes
- Backing up the database in case of corrupted database has been removed, as the original bug that made us introduce this debug function is long gone.
- `ZMTestSession` is gone, the same functionality should be achievable with `StorageStack`


Source/Notifications/ConversationListObserverCenter.swift
```diff
@@ -32,7 +32,8 @@ extension NSManagedObjectContext {
     static let ConversationListObserverCenterKey = "ConversationListObserverCenterKey"
     
     public var conversationListObserverCenter : ConversationListObserverCenter {
-        assert(zm_isUserInterfaceContext, "ConversationListObserver does not exist in syncMOC")
+        // FIXME: Uncomment and fix crash when running tests
```
daehn: This is fixed in #326

PR #322, opened 2017-07-25
Cherry-picked memory fixes from fix/stack-initialization

mikeger:
Copy of memory-related fixes from https://github.com/wireapp/wire-ios-data-model/pull/320


daehn:
@mikeger looks like an API breaker with the changes to the name generator?

mikeger:
Good point, I will check now.

mikeger:
I cannot find the usage of `nameGenerator` in other frameworks, most likely they are going directly to the users for the names.

daehn:
Either way I would make this a major release as some methods have been removed, irregardless whether someone was actually using them or not.

PR #325, opened 2017-07-26
Mark migration as Optional, add storeURL properties

daehn:


Source/ManagedObjectContext/NSPersistentStoreCoordinator+Wire.swift
```diff
@@ -24,7 +24,7 @@ extension NSPersistentStoreCoordinator {
     convenience init(
         localStoreAt url: URL,
         model: NSManagedObjectModel,
-        startedMigrationCallback: @escaping (Void)->(Void)
+        startedMigrationCallback: (() -> Void)?
```
vytis: Is optional closure escaping by default? Seems weird that the annotation is not needed...

daehn: Yes, the annotation is not possible for `Optional` closures.

Source/Notifications/ConversationListObserverCenter.swift
```diff
@@ -32,7 +32,8 @@ extension NSManagedObjectContext {
     static let ConversationListObserverCenterKey = "ConversationListObserverCenterKey"
     
     public var conversationListObserverCenter : ConversationListObserverCenter {
-        assert(zm_isUserInterfaceContext, "ConversationListObserver does not exist in syncMOC")
+        // FIXME: Uncomment and fix crash when running tests
```
daehn: This is fixed in https://github.com/wireapp/wire-ios-data-model/pull/326

PR #328, opened 2017-07-27
Fixing part of tests after stack initialization changes

vytis:
It is still work in progress...


Source/ManagedObjectContext/NSPersistentStoreCoordinator+Wire.swift
```diff
@@ -60,17 +60,39 @@ extension NSPersistentStoreCoordinator {
...
+        removePersistentStoreFromDisk(at: url)
         let options = NSPersistentStoreCoordinator.persistentStoreOptions(supportsMigration: false)
         do {
             try self.addPersistentStore(
                 ofType: NSSQLiteStoreType,
                 configurationName: nil,
                 at: url,
                 options: options)
-        } catch {
+        } catch let error {
```
daehn: 😭

Source/ManagedObjectContext/Storage.swift
```diff
@@ -209,12 +213,14 @@ extension PersistentStorageInitialization {
...
 }
 
-/// Loads the CoreData model from the current bundle
-private func loadManagedObjectModel() -> NSManagedObjectModel {
-    let modelBundle = Bundle(for: ZMManagedObject.self)
-    guard let result = NSManagedObjectModel.mergedModel(from: [modelBundle]) else {
-        fatal("Can't load data model bundle")
+extension NSManagedObjectModel {
+    /// Loads the CoreData model from the current bundle
+    @objc public static func loadManagedObjectModel() -> NSManagedObjectModel {
```
daehn: As this now is a method on `NSManagedObjectModel` it might make sense to change the name to `load()` or something similar.

Tests/Source/ManagedObjectContext/Migrations/DatabaseBaseTest.m
```diff
@@ -40,7 +41,7 @@ - (void)setUp
 {
     [super setUp];
     
-    [NSManagedObjectContext setUseInMemoryStore:NO];
+    StorageStack.shared.createStorageAsInMemory = YES;
```
daehn: Before it was not using an in memory store, no it is. Intended change?

vytis: Good catch!

Source/ManagedObjectContext/NSPersistentStoreCoordinator+Wire.swift
vytis: Not sure actually where did the `error` come from before this change in the line below...

daehn: It's implicit

vytis: Ah nice, didn't know that!

PR #329, opened 2017-07-27
Do not remove context type keys at all

daehn:
# What's in this PR?

There were cases in which tests checked the context type after the context has been torn down. This should not happen, as no test should be running anymore and thus means some tests are not properly torn down or some asynchronous operations do not use the proper dispatch groups, but this also crashes the test suit, which is why we now keep the context type information present.


daehn:
Closed in favour of https://github.com/wireapp/wire-ios-data-model/pull/330

PR #335, opened 2017-08-02
Feature/adding tests

marcoconti83:
Added tests for initialization of store when:
- migration from legacy location
- migration from old model version


mikeger:
🎏 conflict

marcoconti83:
Fixed 🎉

Tests/Source/ManagedObjectContext/StorageStackTests.swift
```diff
@@ -0,0 +1,273 @@
...
+        
+        StorageStack.shared.createManagedObjectContextDirectory(
+            forAccountWith: userID,
+            inContainerAt: self.baseURL
+        ) { directory in
+            contextDirectory = directory
+            expectation.fulfill()
+        }
+        
+        self.waitForExpectations(timeout: 500)
```
daehn: 500 seconds seems a bit long considering it will get multiplied by ~100 when running.

vytis: In this case the base class is vanilla XCTestCase, so it uses the real expectations!

daehn: 😱

daehn: Still might be a bit long I guess.

marcoconti83: It is!

marcoconti83: I forgot to revert it after setting it that high for testing the test (!)

PR #337, opened 2017-08-08
Fetch user ID from legacy store

marcoconti83:
This PR contains a lot of tweaks, because:
- there was a lot of code duplication
- there were still a few memory leaks
- I ported some stuff to Swift, in particular some test were using a lot of helper methods that are not needed anymore (e.g. migration tests)


marcoconti83:
Fixed merge issues

Source/ManagedObjectContext/StorageStack.swift
```diff
@@ -33,83 +36,118 @@ import UIKit
...
+            if let managedObjectContextDirectory = self.inMemoryStores[accountDirectory] {
+                completionHandler(managedObjectContextDirectory)
+            } else {
+                let managedObjectContextDirectory = InMemoryStoreInitialization.createManagedObjectContextDirectory(accountDirectory: accountDirectory, applicationContainer: applicationContainer)
+                self.inMemoryStores[accountDirectory] = managedObjectContextDirectory
+                completionHandler(managedObjectContextDirectory)
+            }
+        } else {
+            let storeFile = accountDirectory.appendingPersistentStoreLocation
+            DispatchQueue(label: "Storage stack").async {
```
daehn: There a multiple places were a `DispatchQueue` with this label is created, maybe it would make sense to extract it into an instance variable instead.

Tests/Source/ManagedObjectContext/DatabaseBaseTest.swift
```diff
@@ -0,0 +1,119 @@
...
+        StorageStack.shared.createManagedObjectContextDirectory(
+            accountIdentifier: userID,
+            applicationContainer: self.applicationContainer
+        ) { directory in
+            contextDirectory = directory
+        }
+        
+        guard self.waitOnMainLoop(until: { contextDirectory != nil }, timeout: 5) else {
+            XCTFail()
+            fatalError()
```
daehn: Not sure if fatal error is needed in addition to making the test fail, IIRC it will crash the whole test suit.

Source/ManagedObjectContext/PersistentStoreRelocator.swift
```diff
@@ -18,119 +18,100 @@
...
-    fileprivate func appendingSuffixToLastPathComponent(suffix: String) -> URL {
+
+    /// Appends a suffix to the last path (e.g. from `/foo/bar` to `/foo/bar_1`)
+    func appendingSuffixToLastPathComponent(suffix: String) -> URL {
         let modifiedComponent = lastPathComponent + suffix
         return deletingLastPathComponent().appendingPathComponent(modifiedComponent)
     }
+    
+    /// Appends the name of the store to the path
+    var appendingStoreFile: URL {
```
daehn: `appending` sounds more like a function to me, maybe `storeFileURL` makes more sense when this is a variable.

WireDataModel.xcodeproj/project.pbxproj
```diff
@@ -2504,9 +2488,9 @@
 			buildSettings = {
 				APPLICATION_EXTENSION_API_ONLY = YES;
 				CLANG_ENABLE_MODULES = YES;
-				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
+				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "iPhone Developer";
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = W5KEQBF9B5;
```
vytis: Probably leftover after updating project settings?

Source/ManagedObjectContext/StorageStack.swift
```diff
@@ -33,83 +36,118 @@ import UIKit
...
-    /// - parameter completionHandler: this callback is invoked on the main queue. It is responsibility
-    ///     of the caller to switch back to the same queue that this method was invoked on.
-    @objc public func createManagedObjectContextFromLegacyStore(
-        inContainerAt containerUrl: URL,
-        dispatchGroup: ZMSDispatchGroup? = nil,
-        startedMigrationCallback: (() -> Void)? = nil,
-        completionHandler: @escaping (ManagedObjectContextDirectory) -> Void
+    /// Attempts to access the legacy store and fetch the user ID of the self user.
+    /// - parameter completionHandler: this callback is invoked with the user ID, if it exists, else nil.
+    @objc public func fetchUserIDFromLegacyStore(
```
daehn: We could also add a default to the optional `startedMigrationCallback` closure in order to being able to omit it if not needed like in `createManagedObjectContextDirectoryForAccountIdentifier:applicationContainer:startedMigrationCallback:completionHandler:`. Then again I am unsure if a caller should ignore this callback in general (except maybe in tests).

Source/ManagedObjectContext/StorageStack.swift
```diff
@@ -33,83 +36,120 @@ import UIKit
...
+        completionHandler: @escaping (UUID?) -> Void
         )
     {
-        directory(forAccountWith: nil, inContainerAt: containerUrl, dispatchGroup: dispatchGroup, startedMigrationCallback: startedMigrationCallback, completionHandler: completionHandler)
+        guard let oldLocation = PersistentStoreRelocator.exisingLegacyStore(applicationContainer: applicationContainer) else {
+            completionHandler(nil)
+            return
+        }
+        
+        self.currentPersistentStoreInitialization = PersistentStorageInitialization()
```
daehn: @marcoconti83 This method always assumes we are using a persistent store on disk, is there a way to test this behaviour using an in memory store? All integration tests in `wire-ios-sync-engine` use in memory stores and call this (which is fine as it will not find an existing store on disk, but I assume we want to add integration tests for this as well).

PR #341, opened 2017-08-11
Fix/migration order

johnxnguyen:
# What's in this PR?

## Problem:
- We migrate the legacy stores by checking for the existence of `store.wiredabase` in any of the legacy locations. If it exists, we then move the store files `store.wiredabase`, `store.wiredabase.-wal` and `store.wiredabase-shm` in that order. If the app crashes after moving the first file, but before the last file, then when the app restarts it sees that it does not need to migrate (since `store.wiredabase` does not exist in the legacy store), even though the previous attempt to migrate did not complete successfully.

## Solution:
- We reverse the order in which the store files are moved. More explicitly, we move all the necessary files (including the SUPPORT and OTR files) and then finally the `store.wiredabase` store file. In this way, if the migration is interrupted, we will continue the migration when the app is restarted

## Refactoring:
- Moving `NSPersistentStoreCoordinator` extension and `PersistentStorageInitialization` into a separate file
- code for generating support folder path extracted into a separate function
- removed `migrateIfNeeded` parameter from `NSPersistentStoreCoordinator` initializer since this functionality is already present in `moveLegacyStoreIfNecessary`

## Note:
- some tests still need to be implemented/fixed.


Source/ManagedObjectContext/ManagedObjectContextDirectory.swift
```diff
@@ -72,7 +72,9 @@ extension ManagedObjectContextDirectory {
             moc.markAsUIContext()
             moc.configure(with: persistentStoreCoordinator)
             ZMUser.selfUser(in: moc)
-            dispatchGroup.apply(moc.add)
```
typfel: Why are you reverting this?

johnxnguyen: for some strange reason Xcode was complaining saying that dispatchGroup does not have a member called apply. Marco suggested to fix it by writing it out explicitly

typfel: `apply` is a utility function which can be imported from `WireUtilities`

daehn: Probably just need to run `carthage bootstrap`

PR #345, opened 2017-08-15
Improve StorageStack Tests

johnxnguyen:
Previously when "creating" the keystore in the tests we were simply creating the `otr` directory and inserting a dummy file in the legacy location. Then after the migration we check that the keystore directory and the dummy file exist in the new account folder.

Now we create an actual keystore with an encrypted session in the legacy directory, and then check that the keystore is migrated to the new account directory and the encrypted session is still intact.


Tests/Source/ManagedObjectContext/DatabaseBaseTest.swift
```diff
@@ -84,18 +84,26 @@ import WireTesting
...
-    @objc public func doesFileExistInKeyStore(accountDirectory: URL, filename: String) -> Bool {
-        let path = accountDirectory.appendingPathComponent("otr/\(filename)")
-        return FileManager.default.fileExists(atPath: path.path)
+    /// Returns true if the given session exists in the keystore for the given account
+    public func doesSessionExistInKeyStore(accountDirectory: URL, applicationContainer: URL, sessionId: EncryptionSessionIdentifier) -> Bool {
+        
+        var hasSession = false
+        
+        let keyStore = UserClientKeysStore(accountDirectory: accountDirectory, applicationContainer: applicationContainer)
+        keyStore.encryptionContext.perform { sessionsDirectory in
```
daehn: The `perform` function sounds asynchronous to me, but unsure if it actually is. If so it might be better to wait until the block has been executed before returning the result, e.g. either by checking if there is a synchronous version (`..andWait`) or by utilising a semaphore or dispatch group.

marcoconti83: It's not async

Tests/Source/ManagedObjectContext/StorageStackTests.swift
```diff
@@ -329,17 +329,18 @@ class StorageStackTests: DatabaseBaseTest {
...
         }
         
+        let accountDirectory = StorageStack.accountFolder(accountIdentifier: userID, applicationContainer: self.applicationContainer)
+        
         // check all legacy stores deleted
         for oldPath in self.previousDatabaseLocations {
             let oldStoreFile = oldPath.appendingStoreFile()
             XCTAssertFalse(checkSupportFilesExists(storeFile: oldStoreFile))
-            XCTAssertFalse(self.doesFileExistInKeyStore(accountDirectory: oldPath, filename: "foo"))
+            XCTAssertFalse(self.doesSessionExistInKeyStore(accountDirectory: oldPath, applicationContainer: self.applicationContainer, sessionId: sessionID))
```
marcoconti83: I think just by trying to open it for checking, we actually create a new one (at the old location). I would check that the folder does not exist anymore instead.

marcoconti83:
I will fix this PR

PR #349, opened 2017-08-18
Detect memory leaks

marcoconti83:
Added detection of memory leaks on managed object context. 
Weak-fy some references to fix found leaks, fixed failing tests.

Needs release with:
- [x] https://github.com/wireapp/wire-ios-testing/pull/14


Source/ManagedObjectContext/StorageStack.swift
```diff
@@ -36,6 +36,11 @@ import UIKit
         return currentStack!
     }
     
+    /// Created managed object context directory. If I don't retain it here, it will 
+    /// eventually de-init and call the tear down of the contexes at the wrong time,
+	/// even if someone is still holding on to it
```
vytis: Broken indentation, maybe tabs mixed with spaces?

vytis:
Needs a release of WireTesting, update the description

PR #350, opened 2017-08-21
Fix: notification for change is not fired in certain cases

vytis:
CoreData has a concept of “Uniquing” which prevents creation of multiple NSManagedObject instances pointing to the same underlying CoreData record, but it does not guarantee the same behavior for faulted objects. When the last reference to the faulted object dissapears CoreData is free to re-create the object again when needed. This poses a problem when used together with NotificationCenter because it captures the “sender” object weakly. If the sender gets deallocated the notification is not delivered to the subscriber.

To work around this issue we keep a strong reference to the object until the token is alive.

### DIFF
--- a/Tests/Source/Model/Observer/NotificationDispatcherTests.swift
+++ b/Tests/Source/Model/Observer/NotificationDispatcherTests.swift
@@ -194,6 +194,34 @@ class NotificationDispatcherTests : NotificationDispatcherTestBase {
         UserChangeInfo.remove(observer: token, forBareUser: user)
     }
     
+    func testThatItNotifiesAboutChangeWhenObjectIsFaultedAndDissapears(){
+        // given
+        var user: ZMUser? = ZMUser.insertNewObject(in: uiMOC)
+        user?.name = "foo"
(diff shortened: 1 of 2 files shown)


vytis:
According to codecov `dealloc` is never called for `NotificationCenterObserverToken`. Will investigate

vytis:
Oh yeah, known issue - code in dealloc is somehow not reported as covered. `NotificationCenterObserverToken` does get deallocated, confirmed both with visual memory debugger and adding a breakpoint in the method.

Tests/Source/Model/Observer/NotificationDispatcherTests.swift
```diff
@@ -194,6 +194,34 @@ class NotificationDispatcherTests : NotificationDispatcherTestBase {
         UserChangeInfo.remove(observer: token, forBareUser: user)
     }
     
+    func testThatItNotifiesAboutChangeWhenObjectIsFaultedAndDissapears(){
```
typfel: typo: `Dissapears` > `Disappears`

PR #356, opened 2017-08-31
Add function to copy a folder recursively

marcoconti83:


typfel:
I thought the old code was already supposed to copy recursively. Could you explain what changed?

marcoconti83:
The old code moves. The new code copies. 
I refactored in one single function as the only difference was whether we were calling "copy" or "move" on the individual items.

vytis:
Does this have any CoreData related code? Maybe WireUtilities is a better place for it?

marcoconti83:
@vytis good point. I think there are more file-system related extensions that would be good to move. I would put in this PR and then go through all those and move them in another PR.

PR #357, opened 2017-09-01
Remove implementation of file system functions

marcoconti83:
These have been moved to https://github.com/wireapp/wire-ios-utilities/pull/26


Cartfile.resolved
```diff
@@ -7,7 +7,7 @@ github "wireapp/wire-ios-cryptobox" "10.0.0"
 github "wireapp/wire-ios-images" "10.0.0"
 github "wireapp/wire-ios-link-preview" "5.1.0"
 github "wireapp/wire-ios-protos" "11.0.1"
-github "wireapp/wire-ios-system" "18.2.1"
-github "wireapp/wire-ios-testing" "12.0.1"
-github "wireapp/wire-ios-transport" "25.0.0"
-github "wireapp/wire-ios-utilities" "17.0.0"
+github "wireapp/wire-ios-system" "18.2.2"
+github "wireapp/wire-ios-testing" "12.1.0"
```
vytis: Maybe let's keep `wire-ios-testing` to `12.0.0` which doesn't have the check for leaks. Seems to be unstable sometimes and needs a bit of work to make it work nicely on CI.

PR #358, opened 2017-09-05
Re-organize folder structure

typfel:
### Why the change
The various data we persist for an account is scattered in different folders so
we decided we need clean it up and put everything inside one folder per account.

### New folder structure
We store our data in a shared container. A shared container has the same structure
as the main bundle container. We want keep all the account related data in a folder
named by the account-id (UUID) so that it's easy to delete.

```
Accounts/
    <UUID>                    -- JSON file storing basic account information
    ...

AccountData/
    <UUID>/
        otr/                  -- folder containing cryptobox sessions
        store/                -- persistence store for messages etc.
        events/               -- persistence store for unprocessed events
        drafts/               -- persistence store for message drafts
        sharedObjectStore     -- booking keeping to sync changes between share extension and main app
    ...

Library/Caches/
    wire-account-<UUID>/
        <PIN cache folders>   -- store images and other data we can download on demand.
```

### What changed
We need migrate the `store` and `otr` folder. Since we don't want break users upgrading from the internal version we also need to support migration for users with the new multi account folder structure.

### DIFF
--- a/Source/ManagedObjectContext/PersistentStoreRelocator.swift
+++ b/Source/ManagedObjectContext/PersistentStoreRelocator.swift
@@ -33,19 +33,30 @@ extension URL {
 }
 
 public struct MainPersistentStoreRelocator {
-    /// Returns the list of possible locations for legacy stores
-    static func possiblePreviousStoreFiles(applicationContainer: URL) -> [URL] {
-        let locations = possibleLegacyAccountFolders(applicationContainer: applicationContainer)
+    /// Returns the list of possible locations for legacy stores. If accountIdentifier is supplied it also 
+    /// includes account directories used after multiple account support was added.
+    static func possiblePreviousStoreFiles(applicationContainer: URL, accountIdentifier: UUID?) -> [URL] {
+        let locations = possibleLegacyAccountFolders(applicationContainer: applicationContainer, accountIdentifier: accountIdentifier)
         return locations.map{ $0.appendingStoreFile() }
     }
     
-    static func possibleLegacyAccountFolders(applicationContainer: URL) -> [URL] {
+    static func possibleLegacyAccountFolders(applicationContainer: URL, accountIdentifier: UUID?) -> [URL] {
+        var accountsFolders = possibleCommonLegacyDirectories()
+        
         let sharedContainerAccountFolder = applicationContainer.appendingPathComponent(Bundle.main.bundleIdentifier!)
-        return possibleCommonLegacyDirectories() + [sharedContainerAccountFolder]
+        accountsFolders.append(sharedContainerAccountFolder)
+        
+        if let accountIdentifier = accountIdentifier {
+            accountsFolders.append(sharedContainerAccountFolder.appendingPathComponent(accountIdentifier.transportString()).appendingPathComponent("store"))
+        }
+        
+        return accountsFolders
     }
 
-    static func possibleLegacyKeystoreFolders(applicationContainer: URL) -> [URL] {
-        return possibleCommonLegacyDirectories() + [applicationContainer]
+    static func possibleLegacyKeystoreFolders(applicationContainer: URL, accountIdentifier: UUID) -> [URL] {
+        let bundleIdFolder = applicationContainer.appendingPathComponent(Bundle.main.bundleIdentifier!)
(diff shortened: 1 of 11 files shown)


Tests/Source/ManagedObjectContext/DatabaseBaseTest.swift
```diff
@@ -73,14 +75,41 @@ import WireTesting
     /// Create storage stack at a legacy location
     @objc public func createLegacyStore(filePath: URL, customization: ((ManagedObjectContextDirectory)->())? = nil) {
         
-        StorageStack.shared.createOnDiskStack(
-            accountDirectory: filePath.deletingLastPathComponent(),
+//        StorageStack.shared.createOnDiskStack(
```
mikeger: Unnecessary any more? Also in other places in the same file (method).

typfel: Oops. 😳 will clean this up.

typfel: So what I did here was to bypass the StorageStack when creating legacy db test fixtures since it will try do a migration.

typfel: Or it wont do the migration but it requires you to pass all the necessary information to do it.

Source/ManagedObjectContext/PersistentStoreRelocator.swift
```diff
@@ -33,19 +33,30 @@ extension URL {
...
-    static func possibleLegacyAccountFolders(applicationContainer: URL) -> [URL] {
+    static func possibleLegacyAccountFolders(applicationContainer: URL, accountIdentifier: UUID?) -> [URL] {
+        var accountsFolders = possibleCommonLegacyDirectories()
+        
         let sharedContainerAccountFolder = applicationContainer.appendingPathComponent(Bundle.main.bundleIdentifier!)
-        return possibleCommonLegacyDirectories() + [sharedContainerAccountFolder]
+        accountsFolders.append(sharedContainerAccountFolder)
+        
+        if let accountIdentifier = accountIdentifier {
+            accountsFolders.append(sharedContainerAccountFolder.appendingPathComponent(accountIdentifier.transportString()).appendingPathComponent("store"))
```
vytis: So this will append "<UUID>/store" as the legacy folder? Isn't this the current place for data store?

Source/ManagedObjectContext/StorageStack.swift
```diff
@@ -179,13 +185,8 @@ public extension StorageStack {
     /// Returns the URL that holds the data for the given account
     /// It will be in the format <application container>/<bundle ID>/<account identifier>
```
vytis: Need to update comment here

Tests/Source/ManagedObjectContext/DatabaseBaseTest.swift
```diff
@@ -22,6 +22,8 @@ import WireTesting
 
 @objc public class DatabaseBaseTest: ZMTBaseTest {
     
+    var accountID : UUID = UUID.create()
```
vytis: Just a nitpick, but it's better to create things in `setUp` so that we could nil it in `tearDown`. Otherwise a copy of it will stay in memory for each `func testFooBar()` in any of the subclasses.

typfel: But this is just an UUID :)

vytis: Yeah in this case is not much of a problem, just wanted to point out that it multiplies the number of these things in memory by number of test cases in all subclasses of `DatabaseBaseTest`.

Source/ManagedObjectContext/PersistentStoreRelocator.swift
typfel: This will append "<SharedContainer>/<BundleID>/<AccountID>/store" as a legacy folder which is current place for the store yes. But this PR will move it to "<SharedContainer>/AccountData/<AccountID>/store"

vytis: Makes sense:)

PR #360, opened 2017-09-12
Handle notification from multiple instances

marcoconti83:
Currently, only one instance of the full data model stack can be alive at any given time. If more than one instance is alive, they will interfere with each other because some internal mechanisms rely on `NSNotification` and `NSNotificationCenter.default`. Notifications sent from one instance (if send to a `nil` object, or if the observer is register for all objects) will affect the behavior of the other instance.

In order to avoid this, it's necessary to make sure that no notification is sent with `nil` object, and that no observer registers for all objects.

# Changes
- Created a new class, `NotificationInContext`, which in addition to an `object` also has a `context`. The idea is to always use the same context (in this case, the UI `managedObjectContext`) as context for all notifications within one instance of the data model stack. Internally it uses `NSNotification` but notifications that are from a different context will be ignored.
- `ZMAssetClientMessage` has been ported from Objc to Swift, because it makes use of the new notifications which are hard to work with in Objc. The class was huge, I broke it up in multiple files.

# Known issues
The code diff is huge :-/


marcoconti83:
In the Core Data model definition, I changed the underlying class of one instances. I'm not sure if this requires a new data model version of not, AFAIK a new version is required only when the entity name or fields change. 
We should test this in a development build to see if the migration succeed.

Source/Notifications/NotificationDispatcher.swift
```diff
@@ -159,71 +173,89 @@ public class NotificationDispatcher : NSObject {
...
     }
     
     public func tearDown() {
         NotificationCenter.default.removeObserver(self)
-        conversationListObserverCenter.tearDown()
-        tornDown = true
+        self.notificationTokens.forEach {
+            NotificationCenter.default.removeObserver($0)
+        }
+        self.notificationTokens = []
```
mikeger: I don't think that we have right now the contract to use `self.` in cases when the compiler can figure it out. Are there considerations to change it and start using `self.` mandatory?

Source/Notifications/ZMMoveIndex.swift
```diff
@@ -1,29 +1,38 @@
...
+        super.init()
+    }
+    
+    public override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? ZMMovedIndex else { return false }
+        return other.from == self.from && other.to == self.to
+    }
+    
+    public override var hash: Int {
+        return self.from.hashValue ^ self.to.hashValue
```
mikeger: I would change the hash function to something where the arguments are treated differently, since now it's commutative, i.e. hash(a, b) == hash(b, a).
Would be enough to do something like (a ^ b * K) % MAX where K is a big enough prime.

mikeger:
@marcoconti83 please check the comments, I can also update the PR myself if you are busy.

Source/Notifications/NotificationDispatcher.swift
marcoconti83: Whether to add `self.` is a stylistic choice. For the compiler it's the same, but I personally prefer having it explicit because:
- without `self.` there's a possible error if later you add a local variable with the same name as the instance variable, now in the rest of that function you are accessing the wrong variable without noticing
- if you move the code inside a closure, you will need to manually add `self.` anyway (or it will complain abut implicit capturing reference to self)

PR #368, opened 2017-09-25
Persist unread count for Account

marcoconti83:
- Add unread count to the persisted JSON of the account
- Also renamed the method `AccountManager.add` to `AccountManager.addOrUpdate`


Source/Model/Account.swift
```diff
@@ -127,7 +136,7 @@ extension Account {
         let data = try? Data(contentsOf: url)
         return data.flatMap {
             (try? JSONSerialization.jsonObject(with: $0, options: [])) as? [String: Any]
-        }.flatMap(Account.init)
+            }.flatMap(Account.init)
```
nicorsm: Tab instead of spaces

marcoconti83: image

PR #370, opened 2017-09-26
Additional test to verify that it obfuscates the emoji and Russian text

mikeger:
Also exposes `String.obfuscated()`


typfel:
So we were creating the obfuscating string in the UI before?

mikeger:
@typfel the issue happens here https://github.com/wireapp/wire-ios/blob/develop/Wire-iOS/Sources/UserInterface/Overlay/InAppNotifications/ChatHeadsViewController.swift#L177

where we just set the font on the text label of the chat head. I need to expose `obfuscated()` to call it from the UI.

PR #371, opened 2017-09-26
obfuscate with seeded random generator

johnxnguyen:
By using the string’s hash value as the seed, this string will always generate the same obfuscated value.

As a result, when this obfuscated string is shown in the redacted type font, it will always appear the same, which finally allows us to include ephemeral messages using the redacted type font in snapshot tests.

### DIFF
--- a/Source/Utilis/Protos/ZMGenericMessage+Obfuscation.swift
+++ b/Source/Utilis/Protos/ZMGenericMessage+Obfuscation.swift
@@ -24,11 +24,12 @@ extension String {
     static func randomChar() -> UnicodeScalar {
         let string = "abcdefghijklmnopqrstuvxyz"
         let chars = Array(string.unicodeScalars)
-        let random = Int(arc4random_uniform(UInt32(chars.count)))
+        let random = Int((drand48() * 100)) % chars.count
         return chars[random]
     }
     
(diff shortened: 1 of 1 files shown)


Source/Utilis/Protos/ZMGenericMessage+Obfuscation.swift
```diff
@@ -24,11 +24,12 @@ public extension String {
     static func randomChar() -> UnicodeScalar {
         let string = "abcdefghijklmnopqrstuvxyz"
         let chars = Array(string.unicodeScalars)
-        let random = Int(arc4random_uniform(UInt32(chars.count)))
+        let random = Int((drand48() * 100)) % chars.count
```
typfel: why do you use `drand48`

johnxnguyen: `arc4random` can't be seeded, and `srand()`, `rand()`, `srandom()` and `random()` aren't available in swift :(

johnxnguyen: well more precisely, the seed methods `srand()` and `srandom()` are, but the two generators are not

PR #373, opened 2017-10-04
Xcode 9 support

vytis:
# Changes

### Implementation
* More specific type annotation in closure

### Tests
* Standard library has `KeyPath`, prefixing our implementation with module name.
* `XCTest.name` is an optional
* Disabled failing tests that were checking name splitting into first and last names
* OCMock crashes when mocking NSManagedObject subclass. There was no need to mock that class anyway.
* One test failed because we were mocking an already mocked object.


Tests/Source/Model/User/ZMPersonNameTests.m
```diff
@@ -537,7 +537,7 @@ - (void)testThatTheFirstComposedCharacterReturnsNilWhenTheStringIsEmpty;
 }
 
 
-- (void)testThatTheInitialsAreGeneratedCorrectlyIfThereIsOnlyOneCompositeCharacter
+- (void)DISABLED_testThatTheInitialsAreGeneratedCorrectlyIfThereIsOnlyOneCompositeCharacter
```
mikeger: This one is broken n XC9?

vytis: Yeah those are broken. We should think about if we need them at all. It seems to rely on underlying implementation which obviously changed...

mikeger: What do you mean about the changed underlying implementation?

vytis: Well it uses `NSLinguisticTagger` under the hood which splits the text into tokens. Its behaviour seems to have changed with iOS11 and it returns different values now.

vytis:
Yeah maybe we should remove them altogether. What do you think @typfel @mikeger?

PR #384, opened 2017-10-20
pin Xcode to "9.0.0"

mikeger:


vytis:
It seems CircleCI still picks up Xcode 9.0.1 even if the config says differently. Expanding the dependencies step we can see that Carthage rebuilds all of them and takes around 5 mins (https://circleci.com/gh/wireapp/wire-ios-data-model/1496)

PR #385, opened 2017-10-23
move filename logic from wire-ios

billypchan:
depends on: 
- [x] https://github.com/wireapp/wire-ios-utilities/pull/29


mikeger:
@billypchan please do not forget to bump the dependency that would allow using the `normalizedFilename`

vytis:
Also a test would be good!

billypchan:
@vytis @mikeger Will pull again later with unit test.

PR #387, opened 2017-10-24
WIP: adding unique constraints

vytis:
Still work in progress. The outline of changes:

- [x] Added unique constraint for UserClient.
- [x] Tests for unique constraint for UserClient.
- [x] Added unique constraint for ZMUser.
- [x] Tests for unique constraint for ZMUser.
- [x] Added unique constraint for ZMConversation.
- [x] Tests for unique constraint for ZMConversation.
- [x] Hotfix for populating newly created `uniqueIdentifier` property
- [ ] Tests for the hotfix
- [ ] Fix broken duplicate removal tests. It needs to use data model from before adding unique constraints.
- [ ] Investigate what happens if we have duplicates in the store from before and migrate to a data model with unique constraints
- [ ] Investigate if there are no problems that `uniqueIdentifier` is optional


vytis:
Will not be implementing any time soon, partial workaround the problem is in #394

PR #389, opened 2017-10-27
Fixing orphan clients

vytis:
After deleting duplicate users we did not remove their clients and they ended up not associated with user. We can remove the invalid ones because they will be recreated when needed.


Resources/Configurations/zmc-config/project.xcconfig
```diff
@@ -40,7 +40,6 @@ GCC_PREPROCESSOR_DEFINITIONS = DEBUG=0 ZM_MAJOR_VERSION=$(MAJOR_VERSION) $(inher
 
 
 // Swift Compiler - Code Generation
-//
-// this causes a crash - must leave it commented for now
-// SWIFT_WHOLE_MODULE_OPTIMIZATION = YES
+
+SWIFT_WHOLE_MODULE_OPTIMIZATION = YES
```
mikeger: Nice!

PR #390, opened 2017-11-01
remove unneccessary check

johnxnguyen:
This check was preventing system messages being created for first time self member join events, b/c the brand new conversation has not had it’s `conversationType` set to `group` yet, and it defaults to `invalid`. Thus we were returning nil. 

However, in the first line of the method we already do event type checking (only `memberJoin`, `memberLeave`, `conversationRename` and `connectRequest` are permitted), so most of these removed checks are unneccessary. Having said that, we are assuming/expecting that the conversation referenced on the Backend actually is a group conversation.


mikeger:
@johnxnguyen  seems like some tests are still failing here.

PR #391, opened 2017-11-03
support filename with custom file suffix

billypchan:


Source/Model/User/ZMUser+Filename.swift
```diff
@@ -28,13 +28,23 @@ public extension ZMUser {
     /// return a file name with length <= 255 - 4(reserve for extension) - 37(reserve for WireDataModel UUID prefix for meta) characters
     ///
     /// - Returns: a string <= 214 characters
-    var filename: NSString {
+    func filename(suffix: NSString? = nil)-> NSString {
```
vytis: I guess we could just use Swift `String` here instead of `NSString`? Or are there any problems with ObjC interoperability then?

Tests/Source/Model/User/ZMUserTests.swift
```diff
@@ -523,4 +526,25 @@ extension ZMUserTests {
...
+        let filename = user.filename(suffix: suffix)
+        
+        // Then
+        XCTAssertEqual(filename.length, 214)
+        XCTAssertTrue(filename.hasPrefix("Some"))
+        XCTAssertTrue(filename.contains("body"))
+        
+        /// check ends with a date stamp, e.g. -2017-10-24-11.05.43
+        let pattern = "^.*[0-9-.]{20,20}\(suffix)$"
+        let regexp = try! NSRegularExpression(pattern: pattern, options: [])
```
vytis: You can mark the test method with `throws` and then do not need to force-unwrap here. This is nicer because the whole test-suite will stop if something goes wrong. However if the test throws an error it will just mark it as failed and continue with all others

Source/Model/User/ZMUser+Filename.swift
billypchan: Yes, ConversationInputBarViewController+Files.m still needs this function.

Tests/Source/Model/User/ZMUserTests.swift
billypchan: Thanks, I will update this.

Source/Model/User/ZMUser+Filename.swift
vytis: I mean more about using `String` instead of `NSString` than removing this function alltogether

PR #392, opened 2017-11-03
Remove usage of depreciated BE fields

typfel:
Remove all usage of the following fields: 

remove `status`, `status_ref`,` status_time` from the self member object
remove `status` from the other member object (it is always 0)
remove `last_event` and `last_event_time` from the conversationMeta

in addition to that the following fields has also been removed:
`archived`, `clearedEventID`, `muted`, `mutedTime`, `lastRead`

### Fixes
After removing these fields the following changes were necessary:

* remove/add participants based if they are part of participants list instead of checking the `status` flag.
* Set default `lastModifiedDate` and `lastServerTimestamp` when creating conversations since we will no longer set one when reading the `last_event_time`.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Transport.m
+++ b/Source/Model/Conversation/ZMConversation+Transport.m
@@ -129,8 +129,8 @@ NSString *const ZMConversationInfoOTRArchivedReferenceKey = @"otr_archived_ref";
     NSMutableOrderedSet<ZMUser *> *removedUsers = [lastSyncedUsers mutableCopy];
     [removedUsers minusOrderedSet:users];
     
-    [self internalAddParticipants:addedUsers isAuthoritative:YES];
-    [self internalRemoveParticipants:removedUsers sender:[ZMUser selfUserInContext:self.managedObjectContext] isAuthoritative:YES];
+    [self internalAddParticipants:addedUsers.set isAuthoritative:YES];
+    [self internalRemoveParticipants:removedUsers.set sender:[ZMUser selfUserInContext:self.managedObjectContext] isAuthoritative:YES];
 }
 
 - (void)updateTeamWithIdentifier:(NSUUID *)teamId
--- a/Source/Model/Conversation/ZMConversation+Transport.m
+++ b/Source/Model/Conversation/ZMConversation+Transport.m
@@ -134,8 +134,7 @@ NSString *const ZMConversationInfoOTRArchivedReferenceKey = @"otr_archived_ref";
     }
     
     for (ZMUser *user in removedUsers) {
-        [self.mutableOtherActiveParticipants removeObject:user];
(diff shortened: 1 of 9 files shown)


Source/Model/Conversation/ZMConversation+Transport.m
```diff
@@ -109,22 +106,36 @@ - (void)updateWithTransportData:(NSDictionary *)transportData serverTimeStamp:(N
 
 - (void)updateMembersWithPayload:(NSDictionary *)members
 {
-    NSArray *users = [members arrayForKey:ConversationInfoOthersKey];
-    for(NSDictionary *userDict in [users asDictionaries]) {
+    NSArray *usersInfos = [members arrayForKey:ConversationInfoOthersKey];
+    NSMutableOrderedSet<ZMUser *> *users = [NSMutableOrderedSet orderedSet];
+    NSMutableOrderedSet<ZMUser *> *lastSyncedUsers = [NSMutableOrderedSet orderedSet];
+    
+    if (self.mutableLastServerSyncedActiveParticipants != nil) {
```
mikeger: This is for the case when we have the local modifications?

mikeger: Bump

typfel: `mutableLastServerSyncedActiveParticipants ` contains the last known server state. `mutableActiveParticipants` contains the locally modified state.

In this method we get the new `mutableLastServerSyncedActiveParticipants `. We don't update it directly because the security level also get updated when we use `internalAddParticipants` and `internalRemoveParticipants`

typfel: Actually now I wonder why we weren't using `internalRemoveParticipants `?

Source/Model/Conversation/ZMConversation+Transport.m
```diff
@@ -109,22 +106,35 @@ - (void)updateWithTransportData:(NSDictionary *)transportData serverTimeStamp:(N
...
+        [users addObject:[ZMUser userWithRemoteID:userId createIfNeeded:YES inContext:self.managedObjectContext]];
+    }
+    
+    NSMutableOrderedSet<ZMUser *> *addedUsers = [users mutableCopy];
+    [addedUsers minusOrderedSet:lastSyncedUsers];
+    NSMutableOrderedSet<ZMUser *> *removedUsers = [lastSyncedUsers mutableCopy];
+    [removedUsers minusOrderedSet:users];
+    
+    for (ZMUser *user in addedUsers) {
+        [self internalAddParticipants:[NSSet setWithObject:user] isAuthoritative:YES];
```
mikeger: Any reason we do for here? We can pass the set to `internalAddParticipants`, same applies to remove

typfel: No :) I was blindly keeping old code structure

PR #393, opened 2017-11-06
replace the type NSString to String

billypchan:
Clean up NSString logic and clean up unit tests

### DIFF
--- a/Tests/Source/Model/User/ZMUserTests.swift
+++ b/Tests/Source/Model/User/ZMUserTests.swift
@@ -513,17 +513,13 @@ extension ZMUserTests {
         XCTAssertTrue(filename.hasPrefix("Some"))
         XCTAssertTrue(filename.contains("body"))
 
-        do {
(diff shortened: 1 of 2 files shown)


Tests/Source/Model/User/ZMUserTests.swift
```diff
@@ -507,6 +507,22 @@ extension ZMUserTests {
...
+        XCTAssertTrue(filename.hasPrefix("Some"))
+        XCTAssertTrue(filename.contains("body"))
+
+        do {
+            let regexp = try NSRegularExpression(pattern: pattern, options: [])
+            let matches = regexp.matches(in: filename as String, options: [], range: NSMakeRange(0, filename.count))
+            
+            XCTAssertTrue(matches.count > 0)
+        } catch {
+            XCTFail("failed to create a NSRegularExpression")
```
vytis: I mean instead of doing try/catch you can say that the test method throws:
```
func testFilenameForUser() throws {
<...>
}
```

PR #394, opened 2017-11-06
Fix crash for duplicate users

vytis:
## Reason

After a bug where we were running multiple sync engines for the same account we had quite a lot of duplicated objects in the local store. We have cleaned that up in a hotfix, but unfortunately it still happens that we create users with same remote identifiers twice.

## Solution

The crashes seem to happen in sync engine when creating users from payload. The fix was to add a method that merges the users when fetching and use it in several cases where it could be created.


Source/Utilis/DuplicatedEntityRemoval.swift
```diff
@@ -36,67 +36,42 @@ enum DuplicatedEntityRemoval {
...
-            firstConversation.needsToBeUpdatedFromBackend = true
+            ZMConversation.merge(conversations, in: context)
         }
     }
     
 }
 
 extension UserClient {
+
+    static func merge(_ clients: [UserClient], in context: NSManagedObjectContext) {
```
mikeger: [Style] You can infer the context from the object itself, we had to attach the context before before since we had to have it for fetching.

vytis: make sense, will update!

vytis:
@mikeger updated

PR #399, opened 2017-12-04
Added support for Availability

nicorsm:
### Issues

We added support for Availability on the Message Strategy framework.

### Solutions

- We added new methods in order to support Availability, also referenced as "broadcast", mainly in `ZMGenericMessage` and its extension.
- We extended `ZMUser` in order to compute which connections and team members should be informed about changes. We also handle changes of the four available `WorkStatus`: `none, vacation, sick, workFromHome`.
- We updated tests.

## Dependencies

Needs releases with:
- [x] https://github.com/wireapp/wire-ios-protos/pull/17
- [x] https://github.com/wireapp/wire-ios-mocktransport/pull/63


vytis:
I guess we also need to add a migration test because we bumped the data store version

Source/Model/User/ZMUser.m
```diff
@@ -414,7 +415,6 @@ - (NSSet *)ignoredKeys;
                                            LocalMediumRemoteIdentifierDataKey,
                                            LocalSmallProfileRemoteIdentifierDataKey,
                                            NormalizedEmailAddressKey,
-                                           NormalizedNameKey,
```
mikeger: The change seems off from the PR topic. Intended?

PR #400, opened 2017-12-04
ZMConversation: add lastEditableMessage

vorobievalex:
In order to support 'edit last message' shortcut ⌘↑ (https://github.com/wireapp/wire-ios/issues/1125), ZMConversation should be able to return lastEditableMessage.

This is a port of the same method from webapp:
https://github.com/wireapp/wire-webapp/blob/589e86613234ded742c32acdba075888cffd17de/app/script/entity/Conversation.js#L574

PS: As an additional outcome of this change, isEditableMessage check can later be moved from wire-ios to wire-ios-data-model to be incapsulated in the message class. I.e.:
https://github.com/wireapp/wire-ios/blob/1dfadd4c65843f7e2824bbfe24d8f2199b25dd21/Wire-iOS/Sources/UserInterface/Conversation/Content/Cells/TextMessageCell.m#L394 can be replaced with message.isEditableMessage().


vytis:
Thanks for a solid pull request and sorry it took so long to look through it! 
I see this is only a part of the implementation as it needs integration with the UI (in `wire-ios`). Would you have some time to work on that as well? The iOS team is happy to help if you have any questions how to go about doing it.

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -111,11 +111,14 @@ extension ZMClientMessage {
     override var isEditableMessage : Bool {
         if let genericMsg = genericMessage {
             return  genericMsg.hasEdited() ||
-                   (genericMsg.hasText() && (deliveryState == .sent || deliveryState == .delivered))
+                    (genericMsg.hasText() &&
+                        (self.sender?.isSelfUser ?? false) &&
+                        !isEphemeral &&
```
vytis: The only change that we were not sure about in this PR is changing the logic slightly here. Why did you decide to add the check for ephemeral messages? We don't allow it from the UI, so it makes sense, but we need to make sure this doesn't break any other functionality.

vorobievalex:
The code for the UI part is ready locally. I just did not find an easy way to recompile wire-ios-data-model and other 2 frameworks that depend on it. So I decided to submit this code first. When it is released, I can test and submit the UI part.
At the same time I would greatly appreciate if anybody explain me how to recompile required packages locally. Carthage just downloads frameworks binaries for me when I run wire-ios/setup.sh

Source/Model/Message/ConversationMessage+Deletion.swift
vorobievalex: I believe the messenger never supports editing of ephemeral messages since web app does not support it in the corresponding function:
https://github.com/wireapp/wire-webapp/blob/a0dcb1fa2777a84ea083fbc0b8e0978a42ea4d07/app/script/entity/message/Message.js#L275

vytis:
Ah I see, there is a trick - you can directly edit `Cartfile.resolved` and change `wire-data-model` to point to `vorobievalex/wire-ios-data-model` and instead of version add your branch or commit hash. Then don't run the whole `setup.sh`, but just do `carthage bootstrap`. It should download and recompile that dependency and all should work. You can also add me on Wire - vytis_wire, might be quicker to troubleshoot any other issues you have!

vorobievalex:
Second part of the feature was sent to UI - https://github.com/wireapp/wire-ios/pull/1511

PR #404, opened 2017-12-11
Delete asset messages if they are cancelled

typfel:
## Issue
If a remote user cancels a file upload the receiver will be stuck with placeholder message which never goes away.

## Cause  / Discussion
The desired behaviour is that a cancelled upload gets deleted and gets re-inserted at the bottom if the sender decides to resume the upload.

## Solution
We delete the message if get a `.cancelled` update. Due to a race condition with preview image downloader we need to guard against a zombie message, because accessing the `genericAssetMessage` property on a deleted  object will trigger a force unwrap.

## Note
I also changed the behaviour when a file message is canceled after it was uploaded. Previously we would mark it as canceled now we ignore it.


mikeger:
@daehn if you have time to glimpse over it, something related to your background area. Thanks!

typfel:
Need to fix some integration tests fail because of these changes.

PR #407, opened 2017-12-15
Fix not sending availability update to users without a handle

typfel:
## Issue
Some users could not set their status because they didn't include all required clients as recipeints

## Cause
The `predicateForConnectedNonBotUsers` were also filtering out users without a handle.

## Solution
Only filter based on connection status


mikeger:
@typfel I am confused, should this be a major release on data model?

typfel:
@mikeger no this is a patch release

PR #408, opened 2017-12-16
Changed order of statuses - swapped Away with Busy

nicorsm:
## What's new in this PR?

Changed order of statuses - swapped Away with Busy. The new order, which will also appear in the UIActionSheet, is: None, Available, Busy, Away.


nicorsm:
Tests are working fine locally, I'll merge the branch.

PR #412, opened 2018-01-02
Fix/obfuscation

johnxnguyen:
## What's in this PR?

Changes in https://github.com/wireapp/wire-ios-data-model/pull/371 weakened the security of obfuscated text messages, this PR reverts the changes as well as adding a test to check that the same message content obfuscates differently each time.


Source/Utilis/Protos/ZMGenericMessage+Obfuscation.swift
```diff
@@ -24,12 +24,12 @@ public extension String {
     static func randomChar() -> UnicodeScalar {
         let string = "abcdefghijklmnopqrstuvxyz"
         let chars = Array(string.unicodeScalars)
-        let random = Int((drand48() * 100)) % chars.count
+        let random = Int(arc4random_uniform(UInt32(chars.count)))
```
vytis: `arc4random` is not really secure for cryptography purposes, only stuff like randomly assigning accent colors etc.
In WireUtilities there is `Data.secureRandomData(length: UInt)`, but it returns `Data` which will have to be converted into index, or somehow this method needs to be refactored.

mikeger: I guess `SecRandomCopyBytes` can be used to copy enough bytes to fill the integer, or something like that.

johnxnguyen: Ok good to know! I'll make the changes.

vytis: This function is basically a wrapper around `SecRandomCopyBytes`. But yeah, just getting enough bytes to fill it would work fine! Or another way would be to get more bytes in one go and use them in a `map` over original chars.

Source/Utilis/Protos/ZMGenericMessage+Obfuscation.swift
```diff
@@ -24,12 +24,13 @@ public extension String {
     static func randomChar() -> UnicodeScalar {
         let string = "abcdefghijklmnopqrstuvxyz"
         let chars = Array(string.unicodeScalars)
-        let random = Int((drand48() * 100)) % chars.count
-        return chars[random]
+        var data = Data.secureRandomData(length: 1)
+        let randomByte = Int(data.removeFirst())
```
johnxnguyen: @vytis is this what you meant?

vytis: We need to be careful with sizes here, you get one byte, but create an Int that is 8 bytes on 64bit machine. You can get the real size with `MemoryLayout<Int>.size`. However in this specific case we are only picking from 26 different values, so maybe it's not a problem.

Also there is a bit of a catch with % operator - https://crypto.stackexchange.com/questions/22767/does-using-modulo-affect-quality-of-randomness

vytis:
This post seems to get the things right with modulo bias - https://medium.com/hownetworks/dont-waste-cycles-with-modulo-bias-35b6fdafcf94, pointing to code how `arc4random` is implemented. Maybe we should add a new method to Utilities for this sort of thing as well.

mikeger:
@johnxnguyen as I understood it, this is not relevant any more, and the test would be removed?

johnxnguyen:
@mikeger No we need this PR (which is only waiting on the Utilities release). The test that needs to be removed is in the UI, which I have ready to go as well.

PR #413, opened 2018-01-02
Mark conversation as read or unread

mikeger:
This feature is useful to mark some conversation as the one requiring attention when you already finished reading it or marking a non-important conversation as read.

Not every conversation can be marked as unread, since it must contain at least one message with the timestamp.

# Todo

- Verify syncing of updated timestamps.

### DIFF
--- a/Tests/Source/Model/Conversation/ZMConversationTests.m
+++ b/Tests/Source/Model/Conversation/ZMConversationTests.m
@@ -1853,7 +1853,7 @@
     XCTAssertEqualObjects(conversation.lastReadServerTimeStamp, serverTimeStamp);
 }
 
-- (void)testThatItMakrsTheMessagesAsRead;
+- (void)testThatItMarksTheMessagesAsRead;
 {
     // GIVEN
     ZMConversation *conversation = [ZMConversation insertNewObjectInManagedObjectContext:self.uiMOC];
@@ -1893,12 +1893,14 @@
     XCTAssertTrue([conversation canMarkAsUnread]);
 }
 
-- (void)testThatItMakrsTheMessagesAsUnread;
+- (void)testThatItMakrksTheMessagesAsUnread;
 {
     // GIVEN
     ZMConversation *conversation = [ZMConversation insertNewObjectInManagedObjectContext:self.uiMOC];
     conversation.conversationType = ZMConversationTypeConnection;
     conversation.remoteIdentifier = NSUUID.createUUID;
+    
+    [self.uiMOC saveOrRollback];
     
     ZMMessage* unreadMessage = [self insertDownloadedMessageAfterMessageIntoConversation:conversation];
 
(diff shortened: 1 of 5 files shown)


typfel:
Are we syncing unread conversation between clients?

mikeger:
It's WIP, but should work, clients at the moment do not accept the timestamps earlier than they know about, so it's not possible to mark as unread.

typfel:
Ok, but disregarding earlier timestamps is a good protection against race conditions between clients.

mikeger:
You are right, then I guess we need to think may be to update the protobuf to add additional flag that the update is user's conscious decision.

vytis:
Could this somehow be used to potentially fix the issue when deleting unread message still keeps the conversation as unread? Of course we can now mark all messages as read, but maybe avoid that situation altogether?

Tests/Source/Model/Conversation/ZMConversationTests.m
```diff
@@ -1853,6 +1853,67 @@ - (void)testThatItSetsTheLastReadServerTimeStampToTheLastEventAfterTheLastMessag
...
+    ZMConversation *conversation = [ZMConversation insertNewObjectInManagedObjectContext:self.uiMOC];
+    conversation.conversationType = ZMConversationTypeConnection;
+    conversation.remoteIdentifier = NSUUID.createUUID;
+    [self insertDownloadedMessageAfterMessageIntoConversation:conversation];
+    [conversation markAsRead];
+    // WHEN & THEN
+    XCTAssertTrue([conversation canMarkAsUnread]);
+}
+
+- (void)testThatItMakrsTheMessagesAsUnread;
```
vytis: Typo - Makrs

Tests/Source/Model/Conversation/ZMConversationTests.m
```diff
@@ -1853,6 +1853,67 @@ - (void)testThatItSetsTheLastReadServerTimeStampToTheLastEventAfterTheLastMessag
     XCTAssertEqualObjects(conversation.lastReadServerTimeStamp, serverTimeStamp);
 }
 
+- (void)testThatItMakrsTheMessagesAsRead;
```
vytis: Same typo here

mikeger: karl_marx

mikeger:
@vytis it could be the way to fix it, but I don't know the details of the bug that you are describing.

vytis:
Also maybe you can fix this analyzer warning we have for quite some time now?

mikeger:
@vytis which one, with `let bundleIdentifier = Bundle.main.bundleIdentifier` ?

vytis:
Yeah, it has been present in quite some time now...

vytis:
BTW the issue pointed by the analyzer is already fixed in develop.

typfel:
To clarify: Is this a internal feature only? I would be cautious to release `marked as unread` to the public with the current limitations. It also has cross platform implications.

mikeger:
@typfel it would be tested for internal people only, just to verify if we need / want it.

PR #419, opened 2018-01-12
Split imageData logic into personal and team avatars

nicorsm:
### Issues

The `imageData` field of the `Account` object is actually designed to contain the avatar of a personal account, or an eventual image of a team account (which actually is nil).

This is causing the `SkeletonViewController` to show a grey screen while switching from personal to team account.

### Solutions

I've added a new field to the `Account` class, `teamImageData`. In this way, `imageData` always contains the personal avatar for both personal and team accounts, and `teamImageData` could contain a team avatar in the future - so we have a more consistent behavior. 

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-sync-engine/pull/665
- [ ] https://github.com/wireapp/wire-ios/pull/1600

## Notes

Issue linked to the ticket 8936.

### DIFF
--- a/Source/Model/Account.swift
+++ b/Source/Model/Account.swift
@@ -34,6 +34,7 @@ public final class Account: NSObject {
     public var teamName: String?
     public let userIdentifier: UUID
     public var imageData: Data?
+    public var teamImageData: Data?
     
     public var unreadConversationCount: Int = 0 {
         didSet {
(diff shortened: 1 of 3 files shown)


typfel:
Since you are adding to the API this will be  a <b>minor</b> release

PR #422, opened 2018-01-22
isServiceUser helper bool getter for ServiceUser protocol

billypchan:
## What's new in this PR?

Add isServiceUser bool getter for ServiceUser protocol and implement in ZMSearchUser and ZMUser. This property is for UI to know the user is a bot or not in one call.

### DIFF
--- a/Source/Model/User/ServiceUser.swift
+++ b/Source/Model/User/ServiceUser.swift
@@ -21,6 +21,7 @@ import Foundation
 @objc public protocol ServiceUser: class, ZMBareUser {
     var providerIdentifier: String? { get }
     var serviceIdentifier: String? { get }
+    var isServiceUser: Bool { get }
 }
 
 @objc public protocol ServiceUserChecker: ServiceUser {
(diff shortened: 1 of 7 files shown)


Source/Model/User/ServiceUser.swift
```diff
@@ -21,6 +21,7 @@ import Foundation
 @objc public protocol ServiceUser: class, ZMBareUser {
     var providerIdentifier: String? { get }
     var serviceIdentifier: String? { get }
+    var isServiceUser: Bool { get }
```
vytis: It is a bit inconvenient to use this flag as you first need to cast to `ServiceUser` and then check `isServiceUser`. There are a lot of APIs where we pass `ZMBareUser` as it is an abstraction for `ZMUser` and `ZMSearchUser`. It would be much more useful to put the flag in `ZMBareUser`.

Also looking into `ServiceUserChecker` protocol it seems we don't use it anywhere in the app. It would make everything much more straightforward to just remove it...

billypchan: The reason for creating ServiceUserChecker is we do not want to expose 
````var serviceUser: ServiceUser? { get }````
in ZMUser. If we have a isServiceUser checker in ZMBareUser, we can move the above getter to extension of ZMSearchUser.

billypchan: @vytis after discussion with @mikeger, ServiceUserChecker is required and we should keep it.

PR #423, opened 2018-01-23
Service users cleanup

vytis:
## What's new in this PR?

Slightly cleaned up protocols around service users - now there is only one protocol `ServiceUser` and a flag on `ZMBareUser` which should be enough to see if user is a bot.
Also rewrote display name function in Swift to get proper nullability annotations.


vytis:
@mikeger Maybe I don't know the complete idea about `if let` to judge then! The code was not used anywhere yet. Having an extra protocol acting as a cast basically seemed a bit too much. We could a property to `ZMBareUser` together with the flag, but will we actually use any properties inside `ServiceUser` anywhere?

mikeger:
I was thinking to have some kind of this API:

```swift

protocol ServiceUser {
    let providerIdentifier: String
    let serviceIdentifier: String
}

protocol ServiceUserProvider {
    var serviceUser: ServiceUser? {
        if self.providerIdentifier != nil && self.serviceIdentifier != nil {
            return self
        }
    }
}
```

Public:
```swift
extension ZMUser: ServiceUserProvider {
}
extension ZMBareUser: ServiceUserProvider {
}
```

Private:
```swift
extension ZMUser: ServiceUser {
}
extension ZMBareUser: ServiceUser {
}
```

And usage like:

```swift
if let serviceUser = user.serviceUser {
    ...show service user related stuff
}
```

But I think it's not a big deal to have it with the bool property.

vytis:
This would make sense if we actually used the service user anywhere. At the moment we would just need to check `if user.serviceUser != nil`. Maybe we will add some more properties there in the future and in that case your approach would be better!

PR #424, opened 2018-01-23
Fix crashes while handling SVG files

nicorsm:
### Issues

The app was:
1) not displaying SVG files (blank view with three dots inside)
2) crashing when opening the Collections view in a conversation with an SVG file shared inside.

### Causes

These issues are caused by the MIME type of the SVG format, which is `image/svg+xml` and it's recognized as an image - which is correct, but from our side, we're unable to display SVG natively, because they're not supported by `UIImage` (they're vectorial drawings and not raster bitmaps).

### Solutions

We need to treat the SVG as a file, not as an image. This means that we have to check both if the shared media is an image and if its MIME type is not `image/svg+xml`. This is done by the extension methods available in `RasterImages+Protobuf.swift` from Data Model.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-message-strategy/pull/108
- [ ] https://github.com/wireapp/wire-ios-transport/pull/76


Source/Model/Message/RasterImages+Protobuf.swift
```diff
@@ -0,0 +1,43 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+public extension ZMAssetOriginal {
+    var hasRasterImage: Bool {
+        return hasImage() && mimeType != "image/svg+xml"
```
daehn: Just played around with this a bit as I thought it might be nice to remove the String comparison from this. It seems like it should be possible to convert the mime type into an UTI and use methods from MobileCoreServices to check wether it is an SVG or not, which could have the advantage of being more future proof, like this:

screen shot 2018-01-24 at 10 36 07

nicorsm: Cool! I'll take a look. Many thanks, @daehn!

daehn: I put a more type-safe version into a gist here for reference https://gist.github.com/daehn/b8849171c34ed964cca99c5d2a1205d4

PR #426, opened 2018-01-26
Forbid connecting to the service user

mikeger:
## What's new in this PR?

It should not be possible to connect to the service user.


mikeger:
I am re-running the tests locally.

mikeger:
@vytis should be fixed now, will wait for the tests to run.

PR #427, opened 2018-01-30
Recalculate displayName map when a name becomes available

daehn:
### Issues

We did not refresh the display name map when a user name becomes available, this can be an issue when a non-local user is added to a conversation as it might not have been fetched yet when the display name map is populated. 

### Solutions

This PR adds a check whether the stored name is empty and the user name is not and will discard and recalculate the user name map in that case.

### DIFF
--- a/Tests/Source/Model/User/DisplayNameGeneratorTests.swift
+++ b/Tests/Source/Model/User/DisplayNameGeneratorTests.swift
@@ -104,7 +104,7 @@ class DisplayNameGeneratorTests : ZMBaseManagedObjectTest {
         // when
         user1.name = "\u{00C5}ron Meister"
         user2.name = "A\u{030A}ron Hans"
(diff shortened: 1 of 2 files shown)


Tests/Source/Model/User/DisplayNameGeneratorTests.swift
```diff
@@ -97,33 +97,33 @@ class DisplayNameGeneratorTests : ZMBaseManagedObjectTest {
         
         let generator = DisplayNameGenerator(managedObjectContext: uiMOC)
         let emptyString = ""
-        XCTAssertEqual(generator.givenName(for: user1), emptyString);
-        XCTAssertEqual(generator.givenName(for: user2), emptyString);
-        XCTAssertEqual(generator.givenName(for: user3), emptyString);
```
vytis: Bye 👋 semicolons

mikeger:
Please also take care to remove the f-word from the tests.

PR #429, opened 2018-01-30
Parse and return the summary of the integration

mikeger:
## What's new in this PR?

The backend introduced the new property of the integration: summary. It is a string that summarizes the functionality of the service.

### DIFF
--- a/Source/Model/User/ServiceUser.swift
+++ b/Source/Model/User/ServiceUser.swift
@@ -22,3 +22,7 @@ import Foundation
     var providerIdentifier: String? { get }
     var serviceIdentifier: String? { get }
 }
+
+@objc public protocol SearchServiceUser: ServiceUser {
+    var summary: String? { get }
+}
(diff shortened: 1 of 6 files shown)


Source/Model/User/ServiceUser.swift
```diff
@@ -22,3 +22,7 @@ import Foundation
     var providerIdentifier: String? { get }
     var serviceIdentifier: String? { get }
 }
+
+@objc public protocol SearchServiceUser: ServiceUser {
+    var summary: String? { get }
```
vytis: Does this get exported to `@objc` without annotation?

mikeger: Not sure, but should be used from Obj-C. Currently @objc is automatically inferred when possible. However if we want to move to Swift 4 would be useful to mark it anyways since https://github.com/apple/swift-evolution/blob/master/proposals/0160-objc-inference.md

mikeger:
Tests are passing locally.

PR #432, opened 2018-02-09
Conversation creation system messages

vytis:
### Issues

With new conversation creation flow we have custom name for the conversation at the time of creation. System messages need to be adjusted to display it nicely.

### Solutions

Added a method to create conversation with name as a (nullable) parameter. Later we should remove the APIs for creating conversation without name as it is not possible anymore from UI.

Added another type of system message that gets shown when created conversation has a custom name defined. Tweaked the existing `.newConversation` system message to have conversation name inside `text` property, so that UI could display it differently when name is set.

There are three cases possible now:
1. Conversation is created without name. This is for backwards compatibility with existing conversations. We show only one system message, no changes here.
2. Empty conversation with name. We show a single system message with conversation name.
3. Group conversation with name. We show system message with name and another one with participants. 

## Notes

Deleted one method to create conversation that was left from the "beta" teams implementation that is no longer used anywhere.


typfel:
I guess we need add a case for this in push notifications as well.

vytis:
I think for now we will keep it as it is, but it should definitely be added!

vytis:
I will re-do the functionality to use only single system message, there are quite a few complications with having two sometimes. It is a UI problem more and data model should not be changed that much to accommodate it.

daehn:
**Update:** We decided to go ahead with this approach as the spec changed again and packing all of this information into one system message cell means replicating a lot of the layout logic as the new cell would not fit into the `IconSystemCell` system very well.

PR #434, opened 2018-02-13
Add option to filter modified keys

daehn:
## What's new in this PR?

* Add method to filter modified keys while ensuring they are only set once in the super class.
* Before it was not possible to remove keys that have been added by `ZMManagedObject` in `updateKeysThatHaveLocalModifications`, as setting them would result in a save, which in turn would result in a recalculation which will trigger `updateKeysThatHaveLocalModifications` being called again, leaving the object in a dirty state.
* This PR adds a separate method that subclasses can override to filter the modified keys to be set by the super class, this way the object will not be changed again.

### DIFF
--- a/Source/Model/Conversation/ZMConversation.m
+++ b/Source/Model/Conversation/ZMConversation.m
@@ -905,10 +905,9 @@ const NSUInteger ZMConversationMaxTextMessageLength = ZMConversationMaxEncodedTe
     return [NSSet setWithObjects:ZMConversationMessagesKey, ZMConversationLastReadServerTimeStampKey, nil];
 }
 
-- (void)updateKeysThatHaveLocalModifications;
+- (NSSet<NSString *> *)filterUpdatedLocallyModifiedKeys:(NSSet<NSString *> *)updatedKeys
 {
-    [super updateKeysThatHaveLocalModifications];
-    NSMutableSet *newKeys = [self.keysThatHaveLocalModifications mutableCopy];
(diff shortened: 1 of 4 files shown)


Source/Model/Conversation/ZMConversation.m
```diff
@@ -905,10 +905,9 @@ + (NSSet *)keyPathsForValuesAffectingLastReadMessage
     return [NSSet setWithObjects:ZMConversationMessagesKey, ZMConversationLastReadServerTimeStampKey, nil];
 }
 
-- (void)updateKeysThatHaveLocalModifications;
+- (NSSet<NSString *> *)filterUpdatedLocallyModifiedKeys:(NSSet<NSString *> *)updatedKeys
 {
-    [super updateKeysThatHaveLocalModifications];
-    NSMutableSet *newKeys = [self.keysThatHaveLocalModifications mutableCopy];
+    NSMutableSet *newKeys = [super filterUpdatedLocallyModifiedKeys:updatedKeys].mutableCopy;
     if (self.unsyncedInactiveParticipants.count > 0) {
```
typfel: I guess this is technically adding keys and not filtering. Maybe this should stay in the `updateKeysThatHaveLocalModifications`

typfel: Ah right, doesn't quite work with they way the method is called.

daehn: We can rename the method, but the important part here is that `updateKeysThatHaveLocalModifications` will set the keys which should only happen once and not again in the subclass.

PR #436, opened 2018-02-16
Change asset cache key

typfel:
## What's new in this PR?

- Base cache keys on conversation and sender in addition to the nonce.
- Changed `ZMMessage` initializer to require a `nonce`
- The `nonce` on the `ZMMessage` is now marked as `nullable`.
- Made helper methods for creating messages internal in order to force dependencies to use the public API:s for adding messages to a conversation. This should hopefully make tests less prone to break in future.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-utilities/pull/36

## Notes

Would have like to have `nonce` be `nonull`. Well investigate a bit more to see if it's safe.

## Todo
- [x] Decide on a caching solution.
- [x] Investigate `nonce` nullability.


Source/Model/Conversation/ZMConversation.m
```diff
@@ -1491,7 +1499,7 @@ - (NSUInteger)sortedAppendMessage:(ZMMessage *)message;
         self.lastModifiedDate = message.serverTimestamp;
     }
     
-    [message prepareToSend];
+//    [message prepareToSend];
```
mikeger: Debug comment?

Source/Model/Message/AssetCache.swift
```diff
@@ -21,20 +21,20 @@ import WireSystem
...
+//    public var zm_imageAssetCache : ImageAssetCache! {
+//        get {
+//            return self.userInfo[NSManagedObjectContextImageAssetCacheKey] as? ImageAssetCache
+//        }
+//        
+//        set {
+//            self.userInfo[NSManagedObjectContextImageAssetCacheKey] = newValue
+//        }
+//    }
+//}
```
mikeger: Most likely not necessary any more?

Source/Model/Message/AssetCache.swift
```diff
@@ -108,52 +108,53 @@ struct PINAssetCache : Cache {
...
+//        let encryptedComponent = encrypted ? "_encrypted" : ""
+//        return "\(messageID.transportString())_\(String(describing: tagComponent))\(encryptedComponent)"
+//    }
+//}
+//
+//public extension ImageAssetCache {
+//    func wipeCache() {
+//        cache.wipeCache()
+//    }
+//}
```
mikeger: Most likely not necessary any more?

Source/Model/Message/AssetEncryption.swift
```diff
@@ -81,58 +81,78 @@ extension Cache {
...
+//        return self.cache.decryptAssetIfItMatchesDigest(plaintextCacheKey, encryptedEntryKey: encryptedCacheKey, encryptionKey: encryptionKey, sha256Digest: sha256Digest)
+//    }
+//    
+//    /// Encrypts a plaintext cache entry to an encrypted one, also computing the digest of the encrypted entry
+//    public func encryptFileAndComputeSHA256Digest(_ nonce: UUID, format: ZMImageFormat) -> ZMImageAssetEncryptionKeys? {
+//        let plaintextCacheKey = type(of: self).cacheKeyForAsset(nonce, format: format, encrypted: false)
+//        let encryptedCacheKey = type(of: self).cacheKeyForAsset(nonce, format: format, encrypted: true)
+//        return self.cache.encryptFileAndComputeSHA256Digest(plaintextCacheKey, encryptedEntryKey: encryptedCacheKey)
+//    }
+//}
```
mikeger: Most likely not necessary any more?

Tests/Source/Model/Messages/AssetCacheTests.swift
```diff
@@ -41,7 +41,7 @@ class FileAssetCacheTests: CacheTests {}
 
 
 // MARK: - Storing and retrieving image assets
-extension ImageAssetCacheTests {
+extension ImageAssetCacheTests { // TODO jacob removed from target
```
mikeger: TODO left

mikeger:
Good ol' sync engine refactoring days!! 👍

typfel:
So I think nonce could be `nonnull` since we were always setting a nonce in `willSave` so even local system message should have a nonce.

typfel:
@vytis @mikeger remember why had this behaviour https://github.com/wireapp/wire-ios-data-model/pull/436/commits/97a87f48b57781b3327a19188f56688822086c66#diff-d9b8d4f5dd3f3e72ccafe339f45b2fa8R1530 ?

vytis:
From my limited search it seems that it was used as cache key at some point (v2 assets), so makes sense that it would change after data is changed - 
https://github.com/wireapp/wire-ios-data-model/blob/0105feaeeb2ad322f57440b545f061c3c28d97c4/Source/Model/Message/V2Asset.swift#L64

https://github.com/wireapp/wire-ios-data-model/blob/8d584785a3c339819c9654ebdc1d9d23fe2f4602/Tests/Source/Model/Messages/ZMClientMessageTests%2BTextMessage.swift#L129

typfel:
@vytis 
The cache key is only used for the in-memory cache in UI AFAIK. Changing the cache key would only be necessary if the image becomes invalid which is only the case when a user changes his/her profile image.

For normal messages the assets are not allowed to be replaced after they've been uploaded. That's why hoping I could get away with using the FileAssetCache cache keys also for the image identifiers.

Source/Model/Message/FileAssetCache.swift
```diff
@@ -163,61 +163,118 @@ open class FileAssetCache : NSObject {
...
+    static func cacheKeyForAsset(_ message : ZMConversationMessage, identifier: String? = nil, encrypted: Bool = false) -> String? {
+        guard let messageId = message.nonce?.transportString(),
+              let senderId = message.sender?.remoteIdentifier?.transportString(),
+              let conversationId = message.conversation?.remoteIdentifier?.transportString()
         else {
-            return "\(messageID.transportString())_\(suffix)"
+            return nil
+        }
+    
+        var key = "\(messageId)_\(senderId)_\(conversationId)"
```
vytis: This maybe could be done slightly more simple by having array of strings and joining it with separator afterwards.

Source/Model/Message/FileAssetCache.swift
```diff
@@ -163,61 +193,118 @@ open class FileAssetCache : NSObject {
...
+              let senderId = message.sender?.remoteIdentifier?.transportString(),
+              let conversationId = message.conversation?.remoteIdentifier?.transportString()
         else {
-            return "\(messageID.transportString())_\(suffix)"
+            return nil
         }
+    
+        var key = "\(messageId)_\(senderId)_\(conversationId)"
+        
+        if let identifier = identifier {
```
vytis: (comment moved from another place):
This maybe could be done slightly more simple by having array of optional strings, flatmapping it and joining with separator afterwards.

Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
```diff
@@ -115,7 +115,24 @@ extension ZMAssetClientMessage: ZMFileMessageData {
...
+        }
+        
+        do {
+            try FileManager.default.createDirectory(at: temporaryFileURL.deletingLastPathComponent(), withIntermediateDirectories: true, attributes: nil)
+            try FileManager.default.linkItem(at: assetURL, to: temporaryFileURL)
+        } catch {
+            return nil
+        }
+        
+        return temporaryFileURL;
```
mikeger: Semicolon 😰

Source/Model/Message/ZMMessage+Internal.h
```diff
@@ -167,7 +169,7 @@ inManagedObjectContext:(NSManagedObjectContext * _Nonnull)moc;
 @interface ZMMessage ()
 
 @property (nonatomic) NSString * _Nullable senderClientID;
-@property (nonatomic) NSUUID * _Null_unspecified nonce;
+@property (nonatomic) NSUUID * _Nullable nonce;
```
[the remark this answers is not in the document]
typfel: I still think we can make it `nonull` but changing will affect so many tests that I leaning towards doing it in a separate PR.

PR #437, opened 2018-02-22
Team-only support: conversation access mode support

mikeger:
## Missing parts:

- [x] Support for the most recent BE status.
- [x] Migration test.
- [x] Payload loading.
- [x] Conversation creation API.


Tests/Source/Model/Conversation/ZMConversationTests+AccessMode.swift
```diff
@@ -0,0 +1,176 @@
...
+    var sut: ZMConversation!
+    
+    override func setUp() {
+        super.setUp()
+        sut = conversation()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+        sut = nil
```
vytis: Tear down should be called at the end of the method

Tests/Source/Model/Conversation/ZMConversationTests+AccessMode.swift
```diff
@@ -0,0 +1,176 @@
...
+    }
+    
+    func testThatItCanSetTheMode() {
+        // when
+        sut.accessMode = .allowGuests
```
vytis: I guess worth checking if access mode is different before this

Tests/Source/Model/Conversation/ZMConversationTests+AccessMode.swift
```diff
@@ -0,0 +1,176 @@
...
+            XCTAssertEqual(sut.allowGuests, $0.0)
+        }
+    }
+    
+    func testThatTheConversationIsInsertedWithCorrectAccessModeAccessRole_Default() {
+        // when
+        let conversation = ZMConversation.insertGroupConversation(into: self.uiMOC,
+                                                                  withParticipants: [],
+                                                                  name: "Test Conversation",
+                                                                  in: nil)!
```
vytis: What should be the behaviour when we create conversations outside of the team? Maybe it should default to something different than inside teams?

mikeger: Good point, I need to research is it safe to set the access mode of `code`, `invite` for the non-team user's conversations.

mikeger:
@vytis the default mode is now defined by the backend: if user specifies the team when creating the conversation the access mode is set, otherwise it's nil and the backend is going to set it to the default values.

PR #439, opened 2018-02-26
Added default values for the conversation creation options

mikeger:
### Issues

Using the new Swift API is not that simple, since all the values for the conversation creation options should be passed in.

### Solutions

Set the default values and let user change them.


typfel:
You don't need default values in the constructor?

mikeger:
@typfel good point, can be further improved.

mikeger:
Tests are succeeding locally.

PR #445, opened 2018-03-08
Introduce keyword to send message to services

alexaubry:
### Issues

When added to the conversation, services/bots receive all the messages, as if they were regular users. This could be a security issue, since a bot could save all the discussions on their servers without users noticing.

### Solutions

The data model includes services in the recipients list only if the message starts with the explicit service mention keyword (`@bots`). If this is not the case, the services will not know about the message. 

## Dependencies

When @ mentions are fully implemented, we will not require the `@bots` keyword.

## Attachments

Bot Handle Demo

## To-Do

- [x] Create Set utility to access service users
- [x] Rename `mentionedServices` to reflect that it works with users
- [x] Create a constant for the keyword


Source/Model/Conversation/ZMConversation+Mentions.swift
```diff
@@ -0,0 +1,56 @@
...
+
+extension ZMConversation {
+
+    // MARK: User Categories
+
+    var serviceUsers: Set<ZMUser> {
+        return services(in: otherActiveParticipants.set as! Set<ZMUser>)
+    }
+
+    func services(in set: Set<ZMUser>) -> Set<ZMUser> {
```
mikeger: I think that this method is not relevant to the class where it's defined, would be possible to define on `Set` that contains users.

mikeger: Same applies to `textMentionsServices` and `categorizeUsers`.

Source/Model/Message/ZMClientMessage+Mentions.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ZMGenericMessage {
+
+    func mentionedServices(within activeServices: Set<ZMUser>) -> Set<ZMUser> {
```
mikeger: Can be renamed to `mentionedUsers`: would work for both users and services.

Source/Model/Conversation/ZMConversation+Mentions.swift
```diff
@@ -0,0 +1,56 @@
...
+    func categorizeUsers(in usersSet: Set<ZMUser>) -> (services: Set<ZMUser>, users: Set<ZMUser>) {
+        let services = self.services(in: usersSet)
+        let users = usersSet.subtracting(services)
+        return (services, users)
+    }
+
+    // MARK: Mentions
+
+    func textMentionsServices(_ text: String) -> Bool {
+        return text.starts(with: "@bots ")
```
mikeger: May be worth putting to the constant property, so it would be easy to find it and modify in case it's needed.

Source/Model/Message/ZMClientMessage+LinkPreview.swift
```diff
@@ -159,19 +159,29 @@ extension ZMClientMessage: ZMImageOwner {
         let original = ZMAssetOriginal.original(withSize: UInt64(imageData.count), mimeType: properties?.mimeType ?? "", name: nil, imageMetaData: imageMetaData)
         
         let updatedPreview = linkPreview.update(withOtrKey: keys.otrKey, sha256: keys.sha256!, original: original)
-        
+        var updatedMentions: [ZMMention] = []
+
+        if let conversation = conversation {
```
typfel: Why do we need to update the mentions in this function? This function update the image data associated with a link preview, the message content should not change.

Source/Model/Message/ZMClientMessage+Mentions.swift
typfel: Why is it useful to pass in the list of `activeServices`? The message has a reference to its conversation so it can look it up itself.

mikeger:
@alexaubry one of the bots developers asked to exclude the "@bots " text from the messages that are going to be delivered to the bots, I guess it should be fine from our side?

mikeger:
@alexaubry one more thing: I guess if the conversation only has 1 other user and it's a bot then it's fine to always send the message to the bot.

Source/Model/Message/ZMClientMessage+Mentions.swift
alexaubry: Since we already filter users and bots before calling this method, this allows us to reuse the results of the filter and not doing it a second time.

Source/Model/Message/ZMClientMessage+LinkPreview.swift
alexaubry: Okay, I was confused by the role of this method. But if the original message contained mentions, shouldn't you re-include them in the updated message?

mikeger: I think we should, would be fine to copy them over I guess.

alexaubry: Okay. I changed the variable name to just `mentions`, to show that we're just copying.

PR #449, opened 2018-03-19
Allow marking message as unread, improved conversation unread

mikeger:
- Added the method to check if the message can be considered unread. The messages sent by the current user cannot be marked as read.


daehn:
The naming confused me a bit at the beginning. Maybe there is a better way to make it more clear, like `canBeMarkedUnread`?

PR #450, opened 2018-03-21
Fire notification to mark assets to be remotely deleted

daehn:
## What's new in this PR?

* Fire notification to mark assets to be remotely deleted.


Tests/Source/Model/Messages/ZMClientMessageTests+Deletion.swift
```diff
@@ -85,6 +91,7 @@ class ZMClientMessageTests_Deletion: BaseZMClientMessageTests {
         
         // then
         assertDeletedContent(ofMessage: sut, inConversation: conversation)
+        XCTAssertEqual(observer.deletedIdentifiers, [assetId])
```
vytis: Also would be good to check if preview asset gets deleted

daehn: @vytis updated!

PR #452, opened 2018-03-22
Add BackupMetadata struct, move ZMVersion from SyncEngine

daehn:
## What's new in this PR?

* Add BackupMetadata struct + validation methods.
* Move ZMVersion from `wire-ios-sync-engine` + convert it to Swift in the process.


Source/ManagedObjectContext/BackupMetadata.swift
```diff
@@ -0,0 +1,129 @@
...
+public extension BackupMetadata {
+    
+    enum VerificationError: Error {
+        case backupFromNewerAppVersion
+        case noVersion
+        case clientMismatch
+        case userMismatch
+    }
+    
+    func verify(using client: UserClient) throws {
```
vytis: I think we will need to create a new client on import anyway, so this is probably not needed.  Also this method does nothing except throwing, maybe it's then better to just return optional error?

Source/Utilis/ZMVersion.m
```diff
@@ -0,0 +1,101 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+#import "ZMVersion.h"
+
+@import Foundation;
+@import WireUtilities;
+
+@interface ZMVersion ()
```
vytis: Should probably move tests as well

Source/ManagedObjectContext/BackupMetadata.swift
daehn: Good point!

Source/Utilis/ZMVersion.m
daehn: Done in d34f5da ;)

Source/ManagedObjectContext/BackupMetadata.swift
daehn: @vytis updated!

daehn:
Tests on SE are failing with these changes, I need to investigate.

PR #454, opened 2018-03-23
Add option to prepare a database for being imported from a backup

typfel:
## What's new in this PR?

When importing a database from a backup we need make sure that there's no:
-  self client
-  stored push tokens
-  stored last fetched event id

## Todo
- [x] tests


Source/ManagedObjectContext/NSManagedObjectContext+BackupImport.swift
```diff
@@ -0,0 +1,31 @@
...
+
+import Foundation
+
+extension NSManagedObjectContext {
+    
+    /// Prepare a backed up database for being imported, deleting self client, push token etc.
+    func prepareToImportBackup() {
+        setPersistentStoreMetadata(nil as String?, key: ZMPersistedClientIdKey)
+        setPersistentStoreMetadata(nil as Data?, key: "pushToken") // TODO jacob: move string constant to data model
+        setPersistentStoreMetadata(nil as Data?, key: "ZMPushKitToken") // TODO jacob: move string constant to data model
```
vytis: Also last fetched notification ID, or it lives somewhere else?

typfel: No just forgot about it :)

vytis:
So the flag that we are restoring from backup will be in memory only for now? In the unfortunate case that something goes wrong and we have this DB already copied it, would there be a way to recover?

typfel:
@vytis  yes for now it's in memory but we can add flag later. Can see that we either we add a flag when we export the database or we store the flag in the account as Mike suggested.

typfel:
Settings the flag when exporting will make the API:s a bit cleaner don't know if it have any other drawbacks.

PR #458, opened 2018-03-26
Add method to import local store from backup

typfel:
## What's new in this PR?

Adds a method to StorageStack for importing a database from a backup.
## Notes

This method assumes the backup directory passed in contains a uncompressed and decrypted backup.

### DIFF
--- a/Source/ManagedObjectContext/StorageStack+Backup.swift
+++ b/Source/ManagedObjectContext/StorageStack+Backup.swift
@@ -94,4 +94,62 @@ extension StorageStack {
             }
         }
     }
+    
+    public enum BackupImportError: Error {
+        case incompatibleBackup(error: Error)
+        case failedToCopy
+    }
+    
+    /// Will import a backup for a given account
+    ///
+    /// - Parameters:
+    ///   - accountIdentifier: account for which to import the backup
+    ///   - backupDirectory: root directory of the decrypted and uncompressed backup
+    ///   - applicationContainer: shared application container
+    ///   - dispatchGroup: group for testing
+    ///   - completion: called on main thread when done. Result will contain the folder where all data was written to.
+    public static func importLocalStorage(accountIdentifier: UUID, from backupDirectory: URL, applicationContainer: URL, dispatchGroup: ZMSDispatchGroup? = nil, completion: @escaping ((Result<URL>) -> Void)) {
+        func fail(_ error: BackupImportError) {
+            DispatchQueue.main.async {
+                completion(.failure(error))
+                dispatchGroup?.leave()
+            }
+        }
+        
+        let queue = DispatchQueue(label: "Database import", qos: .userInitiated)
+        
+        dispatchGroup?.enter()
+        
+        let accountDirectory = accountFolder(accountIdentifier: accountIdentifier, applicationContainer: applicationContainer)
+        let accountStoreFile = accountDirectory.appendingPersistentStoreLocation()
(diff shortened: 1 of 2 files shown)


Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -94,4 +94,62 @@ extension StorageStack {
...
+        }
+        
+        let queue = DispatchQueue(label: "Database import", qos: .userInitiated)
+        
+        dispatchGroup?.enter()
+        
+        let accountDirectory = accountFolder(accountIdentifier: accountIdentifier, applicationContainer: applicationContainer)
+        let accountStoreFile = accountDirectory.appendingPersistentStoreLocation()
+        let backupStoreFile = backupDirectory.appendingPathComponent(databaseDirectoryName).appendingPersistentStoreLocation()
+        let metadataURL = backupDirectory.appendingPathComponent(metadataFilename)
```
vytis: Maybe there is a way to factor this out, as the steps are pretty similar to creating backup

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -94,4 +94,62 @@ extension StorageStack {
...
+        
+        queue.async() {
+            do {
+                let metadata = try BackupMetadata(url: metadataURL)
+                
+                if let verificationError = metadata.verify(using: accountIdentifier) {
+                    fail(.incompatibleBackup(error: verificationError))
```
typfel: should return here

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -94,4 +94,62 @@ extension StorageStack {
             }
         }
     }
+    
+    public enum BackupImportError: Error {
+        case incompatibleBackup(error: Error)
```
mikeger: So that's the error that has error, may be easier to re-throw the original error?

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -94,4 +94,62 @@ extension StorageStack {
...
+                
+                // Import the persistent store to the account data directory
+                try coordinator.replacePersistentStore(at: accountStoreFile, destinationOptions: options, withPersistentStoreFrom: backupStoreFile, sourceOptions: options, ofType: NSSQLiteStoreType)
+                
+                DispatchQueue.main.async {
+                    completion(.success(accountDirectory))
+                    dispatchGroup?.leave()
+                }
+            } catch {
+                fail(.failedToCopy)
```
mikeger: In the future this can start catching some other errors, so this can happen implicitly. May be makes sense to re-throw here too.

daehn: I think it's nice to also (or only) forward the catches error to the user of the API here.

daehn:
@typfel you can pull in develop to use the unzipping methods added there already!

typfel:
@daehn nice! but that's in SE or?

Source/ManagedObjectContext/StorageStack+Backup.swift
typfel: Yes, I can also just forward the errors :) Depends on what want to do with them

daehn:
@typfel Ah yeah, sorry mixed it up.

PR #463, opened 2018-03-28
Using only user identifier to decide if accounts are equal

vytis:
### Issues

To decide if two `Account` objects are equal we compare all properties. We really care only about user identifier, because if team image changes it's still same account.


daehn:
Might be possible to adjust this check here then as well https://github.com/wireapp/wire-ios-sync-engine/blob/develop/Source/SessionManager/SessionManager.swift#L653

PR #464, opened 2018-03-28
Store metadata creation date in ISO8601 format

daehn:
## What's new in this PR?

* Store metadata creation date in ISO8601 format.

### DIFF
--- a/Source/ManagedObjectContext/BackupMetadata.swift
+++ b/Source/ManagedObjectContext/BackupMetadata.swift
@@ -65,7 +65,7 @@ public func ==(lhs: BackupMetadata, rhs: BackupMetadata) -> Bool {
     return lhs.platform == rhs.platform
         && lhs.appVersion == rhs.appVersion
         && lhs.modelVersion == rhs.modelVersion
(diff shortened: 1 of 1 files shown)


Source/ManagedObjectContext/BackupMetadata.swift
```diff
@@ -65,7 +65,7 @@ public func ==(lhs: BackupMetadata, rhs: BackupMetadata) -> Bool {
     return lhs.platform == rhs.platform
         && lhs.appVersion == rhs.appVersion
         && lhs.modelVersion == rhs.modelVersion
-        && lhs.creationTime == rhs.creationTime
+        && lhs.creationTime.timeIntervalSince1970 - rhs.creationTime.timeIntervalSince1970 < 0.001 // We only store 3 floating points
```
vytis: maybe brackets would be good, just to make it more obvious

daehn:
Yeah, there are tests that check serialisation and deserialisation in place already which should be sufficient I think. As I understand it right now the metadata is not really for the user to look at anyways.

PR #467, opened 2018-04-03
Add new model version including ZMUser.teamIdentifier field

daehn:
## What's new in this PR?

* Add new model version including `ZMUser.teamIdentifier` field.

### This PR is still missing:

- [x] A model version schema migration test.


vytis:
Should we refetch all users after this change? Now some will have the team set, but others no?

daehn:
@vytis HotFix added in https://github.com/wireapp/wire-ios-sync-engine/pull/735.

PR #472, opened 2018-04-12
zmessaging 2.47 cleanup

mikeger:
- Cascade delete `ZMGenericMessageData` -- integrated version does not work for some reason, had to override `prepareForDeletion`.
- Added tests.
- New data model version.


Source/ManagedObjectContext/NSManagedObjectContext+Patches.swift
```diff
@@ -25,3 +25,21 @@ extension NSManagedObjectContext {
         PersistedDataPatch.applyAll(in: self)
     }
 }
+
+
+extension NSManagedObjectContext {
+    func batchDeleteEntities(named entityName: String, matching predicate: NSPredicate) throws {
```
vytis: A test for this method would be good

mikeger: It's extracted from `InvalidClientsRemoval`, so I guess it was already tested?

vytis: Well for stuff like do other contexts get notified and so on, since we have it as a separate method now it maybe needs some more test-love :)

mikeger: Ok, will do!

mikeger: @vytis added more tests 🌮

mikeger:
@vytis updated the cleaning up of the store.

PR #473, opened 2018-04-13
Fix: Gif sent via giphy isn't animated

billypchan:
### Issues

Gif sent via giphy isn't start anination

### Causes

The conversation screen's collection view cell will be updated two times after the user selected a GIF. The first update missing MIME data, which creates a cache in ImageMessageCell and the second update will use the cache and animation would not start.

### Solutions

Write MIME type when creating ZMGenericMessage

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-images/pull/15


Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -26,16 +26,23 @@ import Foundation
...
         self.init(nonce: nonce, managedObjectContext: managedObjectContext)
         
-        // We update the size and mimeType once the preprocesing is done
+        // We update the size once the preprocessing is done
+        // mimeType is assigned first, to make sure UI can handle animated GIF file correctly
+        var mimeType = ""
+
+        if let contentType = ZMAssetMetaDataEncoder.contentType(forImageData: imageData) {
+            mimeType = contentType
+        }
```
mikeger: All can be replaced with `let mimeType = ZMAssetMetaDataEncoder.contentType(forImageData: imageData) ?? ""`

PR #475, opened 2018-04-17
Sanitize generic messages for sensitive information when writing internal debug logs

typfel:
### Issues

For internal builds we allow users to send debug logs which may contain sensitive information.

### Solutions

Sanitize all messages from user generated content before writing them to the internal debug log.

### DIFF
--- a/Source/Model/Message/ZMGenericMessage+Debug.swift
+++ b/Source/Model/Message/ZMGenericMessage+Debug.swift
@@ -68,7 +68,7 @@ fileprivate extension ZMArticle {
 
 extension ZMGenericMessage {
     
-    public var sanitizedDebugDescription: String {
+    open override var debugDescription: String {
         
         guard let builder = self.toBuilder() else { return "" }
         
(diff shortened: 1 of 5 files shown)


Source/Model/Message/ZMGenericMessage+Debug.swift
```diff
@@ -0,0 +1,88 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+fileprivate let retractedValue = "<retracted>"
```
vytis: Probably should be "redacted"

Source/Model/Message/ZMGenericMessage+Debug.swift
```diff
@@ -0,0 +1,88 @@
...
+        builder.setPermanentUrl(retractedValue)
+        builder.setSummary(retractedValue)
+        return builder.build()
+    }
+    
+}
+
+extension ZMGenericMessage {
+    
+    public var sanitizedDebugDescription: String {
```
vytis: Maybe we should just use the regular `debugDescription`? Or there are some cases where we want to print the otherwise sensitive contents?

typfel: Nah let's override `debugDescription`

PR #477, opened 2018-04-19
Fix/last read after restore

johnxnguyen:
### Issues

After restoring from backup & opening a conversation with unread messages, then immediately returning to the conversation list, the unread badge is not cleared.

### Causes

The timestamp of the last message in the visible message window is used update the last unread server timestamp, but only if it is not from the self user. Since we have restored a backup and are on a new device, a new device system message is appended to the conversation. And since this system message is 'sent' from the self user, the server timestamp is not updated until the user scrolls up, and the pulls the system message offscreen.

### Solutions

When attempting to update the last read server time stamp, we check if the given message (which is the last one when the conversation loads) is a system message from the self user. If so, then we find the last received message and use its timestamp to update the server timestamp.

**EDIT:** Instead of backtracking to the last received unread message, we now use the system message's timestamp. This simplifies the code and rests upon the likelihood that if you see a system message that follows received messages, you have also seen the previous messages.

## Notes

The logic here in this part of the codebase is quite tricky, and I'm not 100% confident that my solution does not introduce unwanted side effects, so if you have any concerns please raise them!

### DIFF
--- a/Source/Model/Conversation/ZMConversation.m
+++ b/Source/Model/Conversation/ZMConversation.m
@@ -586,28 +586,6 @@ const NSUInteger ZMConversationMaxTextMessageLength = ZMConversationMaxEncodedTe
         return;
     }
     
-    // If the message is a self system message, then we would not update the
-    // last read server time stamp, even if there are previous unread messages
-    if ([message isKindOfClass:[ZMSystemMessage class]] && senderIsSelfUser) {
-        // find the last message where sender is not self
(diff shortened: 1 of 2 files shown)


johnxnguyen:
@typfel yeah that makes sense. But if I may ask, what is the exact reason why we don't want to update the timestamp for messages that the self user sends? In which circumstance would it happen that I can send a message without seeing the latest unread message from another user?

PR #479, opened 2018-04-20
ZIOS-8814: Fix crash for messages with invalid ID

alexaubry:
### Issues

When receiving a message where the identifier field is invalid, the app crashed. The crash would also happen when re-opening the app, thus crashing the app forever.

### Causes

The crash happened because the message was inserted with a `nil` nonce. This caused a crash when sending the confirmation (the conversation was not linked to the message).

### Solutions

This PR introduces systematic validation of messages when decoding them from an update event. We check that every ID field id a valid UUID to prevent this particular crash and potential future undiscovered problems related to invalid nonces.

Furthermore, we changed the parameters of every factory method for messages and assets from `String` to `UUID` to enforce the use of valid UUID strings and prevent this invalid nonce problem from our side.

When the user receives an invalid message, we insert a system message with type `ZMSytsemMessageTypeInvalid` to let the user know.

## Notes

This PR introduces potentially breaking changes. Once it is merged and the changes will be introduced in a new release, this will need to be done:

- Audit frameworks depending on data model for required changes from `String` to `UUID` in factory methods 
- Create a new cell for the`ZMSytsemMessageTypeInvalid` in the app (currently, the app will display the "unsupported message" cell)


Source/Model/Conversation/ZMConversation+Invalid.swift
```diff
@@ -0,0 +1,33 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ZMConversation {
+
+    /// Appends a "message invalid" system message
+    public func appendInvalidSystemMessage(at date: Date, sender: ZMUser, moc: NSManagedObjectContext) -> ZMSystemMessage {
```
mikeger: Passing a `moc` here is not a good idea, and here is why:
Every non-deleted managed object in CoreData belongs to some managed object context. One context live on one queue. The objects from one context can only have relations between each other. If you try to establish the relation between the objects that belong to different contexts this would break this principle and CoreData would have to work with objects that are designed to be accessed only from one queue from another queue. `self`, i.e. the conversation in this case belongs to some managed object context. You cannot insert system message to _some_ context and then establish the relation to `self`.

alexaubry: Thank you for explaining. Updated the PR with @vytis's help.

PR #480, opened 2018-04-20
ZIOS-8814: Fix asset ID verification

alexaubry:
### Issues

#479 introduced a bug that caused all messages with an asset to fail verification and be replaced by a system message.

### Causes

We mistakenly validated the asset key and token as a UUID, but these fields are never sent as UUIDs by the server. This caused the validation to fail all the time, even when the asset can be used.

### Solutions

We reverted the changes and made sure that the asset keys and tokens are never validated as UUIDs.

### DIFF
--- a/Source/Model/Validation/PBMessage+Validation.swift
+++ b/Source/Model/Validation/PBMessage+Validation.swift
@@ -1,0 +1,295 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+import WireProtos
+
+extension UUID {
+
+    fileprivate static func isValid(object: Any?) -> Bool {
+        guard let string = object as? String else { return false }
+        return UUID(uuidString: string) != nil
+    }
+
+    fileprivate static func isValid(bytes: Data?) -> Bool {
+        return bytes?.count == 16
+    }
+
+    fileprivate static func isValid(array: [Any]?) -> Bool {
+        return array?.map(UUID.isValid).contains(false) == false
+    }
+
+}
+
+// MARK: - Specific Validation
+
+// MARK: Generic Message
+
+extension ZMGenericMessage {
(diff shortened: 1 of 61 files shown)


Source/Model/Validation/PBMessage+Validation.swift
```diff
@@ -0,0 +1,258 @@
...
+// MARK: Last Read
+
+extension ZMLastRead {
+    @objc public func validatingFields() -> ZMLastRead? {
+        guard UUID.isValid(object: conversationId) else { return nil }
+        return self
+    }
+}
+
+extension ZMLastReadBuilder {
```
typfel: Is this used somewhere?

typfel: Some question applies for all other builder except for `ZMMentionBuilder`

vytis:
The test coverage is pretty good, but there are some spots where it could be better - https://codecov.io/gh/wireapp/wire-ios-data-model/pull/480/diff

Source/Model/Validation/PBMessage+Validation.swift
alexaubry: Ah that's right, they're not used anywhere. Should we remove them?

alexaubry: Removed them, as discussed.

alexaubry:
@vytis I tried to increase the coverage in my latest commit. Any other cases I should test?

Source/Model/Validation/PBMessage+Validation.swift
```diff
@@ -21,21 +21,67 @@ import WireProtos
...
         return array?.map(UUID.isValid).contains(false) == false
     }
 
 }
 
+// MARK: - String Formatting
+
+extension String {
+
+    public var isValidAssetID: Bool {
```
vytis: Maybe would be simpler to just split the string into prefix and UUID parts and check them separately. The character set approach doesn't really take into account that asset id structure is very strict.

alexaubry: Right. But when we receive the asset, we don't really know if it's in the v2 or v3 format, do we?

vytis: Asset v2 is accessed through completely different API, the call for fetching them is `/conversations/<conv_id>/otr/assets/<asset_id>`. Uploading is disabled for quite some time now (1-1.5 years I think), so there will be no new events coming that contain v2 asset. They are only kept for compatibility reasons, e.g. someone has uploaded a picture long time ago to a conversation with you and you want to download it.

typfel: On the other hand the asset id created and managed by the server, they could change the format . To us it's just a random string so I think it makes sense to only check that it doesn't contain illegal characters.

vytis: I might be wrong but I think the format is documented and could be considered as part of public API. In this case it's a user-controller value, and while it's probably enough to just check characters, but it's hard to say if it would completely solve the problem.

vytis: After a short talk with @tiago-loureiro it seems it's better to treat it as opaque string and avoid strict validation.

PR #493, opened 2018-06-18
Global destruction timer: save, parse payload

mikeger:
~~For the normal expiring messages we currently use the non-synced properties `destructionTimeout` and `messageDestructionTimeout`. `messageDestructionTimeout` is the "private" one: it's of the `TimeInterval` type, we use it whereas the other one `destructionTimeout` is public.~~

~~We have 2 properties is in order to expose the "nice" swift enum value, and maintain the compatibility with the CoreData storage.~~

~~In the addition to the existing properties, we are adding 2 more properties:`globalDestructionTimeout` and `globalMessageDestructionTimeout`. They would serve the same purpose, but for the global (synced) timer.~~

Better Swift-ObjC API is introduced. Rejoice!

# Pending

- [x] Observer system update


Resources/zmessaging.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>zmessaging2.47.0.xcdatamodel</string>
+	<string>zmessaging2.48.0.xcdatamodel</string>
```
typfel: So in theory it's not necessary to create a new model any longer.

Source/Notifications/ObjectObserverTokens/ConversationChangeInfo.swift
```diff
@@ -115,6 +116,10 @@ extension ZMConversation : ObjectInSnapshot {
                changedKeysContain(keys: #keyPath(ZMConversation.accessRoleString))
     }
     
+    public var globalDestructionTimeoutChanged : Bool {
+        return changedKeysContain(keys: #keyPath(ZMConversation.globalMessageDestructionTimeout))
```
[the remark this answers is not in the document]
daehn: Yeah I think we should convert all occurences of `#keyPath` to the new keypath type shorthand at some point.

daehn:
Not sure about the naming `global`, maybe a prefix like `shared` or `synchronised` would indicate better that this will be synced between all participants in the conversation.

Source/Model/Conversation/ZMConversation+Ephemeral.swift
```diff
@@ -83,20 +83,70 @@ public extension ZMConversationMessageDestructionTimeout {
...
+        }
+        set {
+            let currentValue = messageDestructionTimeout
+            
+            if let newTimeout = newValue {
+                switch (currentValue, newTimeout) {
+                case (_, .synced(let value)):
+                    syncedMessageDestructionTimeout = value.rawValue
+                case (.some(.synced(_)), .local(_)):
+                    fatal("Not allowed to set synced timeout when local is set")
```
daehn: The error message seems misleading to me, as far as I can understand the new value we are trying to set is `.local` and the old one is `.synced`, which shouldn't be allowed.

daehn: BTW there is a nice shorthand syntax for switching over optionals, instead of `.some(.synced(_))` it is also possible to write `.synced?`.

Resources/zmessaging.xcdatamodeld/.xccurrentversion
mikeger: I tried it, but it was not working properly somehow. Needs more investigation / separate task.

mikeger:
@zenkins release major

PR #497, opened 2018-06-22
Restored isWeeks variable for ephemeral messages

nicorsm:
## What's new in this PR?

Restored `isWeeks` variable from UI project.


Source/Model/Conversation/ZMConversation+Ephemeral.swift
```diff
@@ -141,6 +142,10 @@ public extension MessageDestructionTimeoutValue {
     var isDays: Bool {
         return rawValue >= 86400
```
daehn: Maybe worth changing is Days to check `86400..< 604800 ~= rawValue` in that case.

nicorsm:
@zenkins release minor

vytis:
The reason why magic phrase didn't work - "3 of 4 required status checks are expected.". Basically we can't merge until we have circleci passing. I will turn it off for now and try again!

nicorsm:
@zenkins release minor

daehn:
@zenkins release minor

PR #504, opened 2018-06-26
Don't update last modified date for message timer update system messages

daehn:
## What's new in this PR?

* Don't update last modified date for message timer update system messages.


daehn:
@zenkins release patch

PR #505, opened 2018-06-27
Use query generations on UI and search MOC

vytis:
### Issues

Because we have two MOCs and data can be updated in either one of them it can get out of date quickly. For example we try to display a conversation message in the UI, but before data is loaded the actual message is deleted. There are several known bugs with ephemerals about this.

### Solutions

Starting from iOS 10 there is a new CoreData feature that allows to pin the MOC to data snapshot at certain time. All fetches will then show data as it was at that time, even if it was changed afterwards. On the DB level this is done by keeping the log of all changes (in the .wal file) that allows recreating state of each transaction.


mikeger:
Suspiciously small changeset! :)

vytis:
And unexpectedly large amount of failing tests :)

vytis:
Closing this PR as there seem to be some issues with this and in-memory stores used in tests. The reason is unclear, will try to maybe isolate it somehow.

PR #507, opened 2018-06-27
Expire self assets from other devices

mikeger:
### Issues

We start expiring the text messages from the self device, but ignore images and assets.

### Causes

Expire messages and assets, too.


mikeger:
Tests pending

PR #508, opened 2018-06-28
Filter deleted messages from message window

daehn:
# What's in this PR?

**This PR adds filtering of deleted messages from the message window when recalculating.**

When deleting a message, we delete its content and then set `visibleInConversation` to nil, and `hiddenInConversation` to the conversation it was a part of. However, in some rare cases the inverse relationship that field, which is `messages` on `ZMConversation` has not been updated yet after a save when we recalculate the message window. I am not sure why this happens yet and it seems to be some kind of race-condition as it doesn't always happen.

## Alternative solutions for this problem:

There are ways to manually enforce refreshing of inverse relationships using `processPendingChanges` as this 

> Forces the context to process changes to the object graph.

according to the documentation. This should already be called at some point after a save during the event loop according to the docs. I went with the aforementioned workaround for now as I want to avoid calling this method every time after a save and only when a message has been deleted, which needs more changes to be achieved in a nice way (the only way I could come up with so far is to set a flag in the context user info that we can check after saving, `forceProcessPendingChanges` or something like that).


vytis:
I have recently noticed we have a custom [merge policy](https://github.com/wireapp/wire-ios-data-model/blob/develop/Source/ManagedObjectContext/ZMSyncMergePolicy.m) which fires sometimes when ephemerals are deleted. Maybe worth taking a look what happens there.

mikeger:
🎏 merge conflict somehow

daehn:
@vytis I looked into the merge policy but couldn't find anything ephemeral specific so far. @mikeger updated!

vytis:
@daehn it's not ephemeral-specific, but it does something about merging the ZMConversation.messages from multiple contexts. Didn't look into the code long enough to understand what it is solving though. If you put a breakpoint on `prepareMessagesInConversationForMergeWithConflicts:` and try to replicate the bug it will be triggered I think.

daehn:
@vytis I'd still merge this workaround while investigating a proper fix.

PR #509, opened 2018-06-28
Add helper properties for timeout existence

johnxnguyen:
## What's new in this PR?

Added some helpful properties to determine if a conversation has a synced and/or local destruction timeouts.


Source/Model/Conversation/ZMConversation+Ephemeral.swift
```diff
@@ -249,5 +249,8 @@ public extension ZMConversation {
     
     @NSManaged internal var localMessageDestructionTimeout: TimeInterval
     @NSManaged internal var syncedMessageDestructionTimeout: TimeInterval
+    
+    public var hasSyncedDestructionTimeout: Bool { return syncedMessageDestructionTimeout != 0 }
+    public var hasLocalDestructionTimeout: Bool { return localMessageDestructionTimeout != 0 }
```
daehn: In theory both internal vars could be set (when setting a synced one after a local one has been set). To avoid this ambiguity that both flags return true (not sure if this is wanted) you could check whether the destructionTimeout property is .local or .synced as these are mutually exclusive.

johnxnguyen: Indeed, I was doing that before, but I thought it would be cleaner to have these properties rather than unwrapping the timeout and case checking its value. My use case for this is inside the processing of the update event. In the case that we turn off the sync value, reading the timeout from the conversation would return the local timeout (given that it exists). It was super confusing why after setting the timeout to `.synced(.none)` and reading it back moments later to get a local timeout instead.

johnxnguyen: But yes, perhaps it is not immediately obvious that the synced timeout has higher priority than the local timeout.

johnxnguyen:
@zenkins release patch

PR #511, opened 2018-07-02
Add logs for ephemeral deletion and obfuscation timer

daehn:
## What's new in this PR?

* Add logs for ephemeral deletion and obfuscation timer.


daehn:
@zenkins release patch

PR #512, opened 2018-07-03
Fix/audio message timer

johnxnguyen:
## What's new in this PR?

This PR introduces a new method to extend a destruction timer of a `ZMAssetClientMessage`.  This will be used in the UI to handle the case when an ephemeral audio message is longer than its timeout (thus making it impossible to hear the entire message).


zenkins:

vytis:
Somehow you were not whitelisted with @zenkins, fixed it so commenting here to make a release should work fine

johnxnguyen:
@zenkins release minor

PR #513, opened 2018-07-06
Add unavailable transfer state and check it before starting download

daehn:
## What's new in this PR?

* Add unavailable transfer state and check it before starting download.


vytis:
Small test for this maybe?

daehn:
@vytis there are some tests added in the request strategy PR, but I'll make sure to add one to check the added branch here as well.

daehn:
@vytis updated!

daehn:
@zenkins release minor

daehn:
Actually thinking about it adding an enum value is an API breaker (exhaustive switches), should probably have been a major instead.

PR #515, opened 2018-07-12
Feature/new new conversation system message cell zios 10271

daehn:
## What's in this PR?

* Add new model schema version including `allTeamUsersAdded` and `numberOfGuestsAdded` fields which will be used for the updated `.newConversation` system message.
* Add test for `.newConversation` system message when inserting new conversation.

### This PR depends on: 

- [x] The new model schema version is still missing a migration test.


Source/Model/Message/ZMMessage.m
```diff
@@ -856,6 +856,8 @@ + (instancetype)createOrUpdateMessageFromUpdateEvent:(ZMUpdateEvent *)updateEven
     message.users = usersSet;
     message.text = messageText != nil ? messageText : name;
     
+    [message updateNewConversationPropertiesWithUsers:usersSet context:moc conversation:conversation];
```
typfel: Only necessary to run when `systemMessageType` is `ZMSystemMessageTypeNewConversation ?

daehn: Yeah, misleading naming, should be referring to new conversation system messages so maybe better to include that in the name.

daehn: @typfel changed the name slightly now.

daehn:
@vytis updated!

vytis:
@daehn ready to merge? Tests are green.

daehn:
@vytis merged!

PR #516, opened 2018-07-12
Store push token inside user client

vytis:
### Issues

We were storing push token in the NSManagedObjectContext's persisted metadata but this makes it hard to keep track of when to update it. 

### Solutions

Swiftified version of PushToken was added and it is stored as a property on UserClient. To migrate from the old format, we have also moved simplified implementation of `ZMPushToken` from Sync Engine.

## Notes

1. After https://github.com/wireapp/wire-ios-data-model/pull/515 is merged there will be conflicts because of updated store version.
2. Migration tests need to be added, will be done after 1. is done


Source/Model/UserClient/ZMPushToken.m
```diff
@@ -0,0 +1,99 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+// 
+
+#import "ZMPushToken.h"
+
+#import <stdio.h>
```
mikeger: I know that it's just copied from the SE, got flashback from old C days with this `stdio.h` :)

vytis: This is not needed actually anymore, will remove so that all the sad memories can rest in peace.

vytis:
@zenkins release minor

PR #518, opened 2018-07-13
Fix: ZIOS-10275 check messages are expired in ZMConversationMessageWindow's recalculateMessages method

billypchan:
### Issues

Some expired messages in the conversation are still visible and displayed as unknown-type system messages.

### Causes

For some reason, there are some ZMMessages in Core Data that the message has no data, destructionDate are expired but isObfuscated flags are false. 

### Solutions

In recalculateMessages method, filter the messages which have expired destructionDate and isObfuscated flags are false.

## Notes

Converted ZMConversationMessageWindow to Swift. Simplify the filtering by replacing NSPredicate with closure.


Source/Model/Message/ZMMessage.m
```diff
@@ -1035,7 +1035,7 @@ - (BOOL)startDestructionIfNeeded
 - (void)obfuscate;
 {
     ZMLogDebug(@"obfuscating message %@", self.nonce.transportString);
-    self.isObfuscated = true;
+    self.isObfuscated = TRUE;
```
daehn: Maybe it's better to use `YES` here.

Source/Model/Conversation/ZMConversationMessageWindow.swift
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+
+    @objc func recalculateMessages() {
+        let messages = conversation.messages
+        let numberOfMessages = Int(activeSize)
+        let range = NSRange(location: messages.count - numberOfMessages, length: numberOfMessages)
+        let newMessages = NSMutableOrderedSet(orderedSet: messages, range: range, copyItems: false)
+
+        let filtered = newMessages.filter{ message in
```
daehn: Missing space after `filter`.

Source/Model/Conversation/ZMConversationMessageWindow.swift
```diff
@@ -0,0 +1,114 @@
...
+            guard let message = message as? ZMMessage else { return false }
+
+            var filterResult: Bool!
```
daehn: Not sure I'd make this a implicitly unwrapped optional, maybe better to just initialize it `false` directly when declaring it?

Source/Model/Conversation/ZMConversationMessageWindow.swift
```diff
@@ -0,0 +1,114 @@
...
+            self.size = max(0, UInt(conversation.messages.count) - firstUnreadIndex + size)
+        }
+
+        recalculateMessages()
+        conversation.managedObjectContext?.messageWindowObserverCenter.windowWasCreated(self)
+    }
+
+    deinit {
+        if let zm_isValidContext = conversation.managedObjectContext?.zm_isValidContext,
+            zm_isValidContext == true {
```
daehn: No need to check `== true` if the optional is already unwrapped, I think you can either do
```swift
if true == conversation.managedObjectContext?.zm_isValidContext {}

// Or with the unwrapping in place:

if let zm_isValidContext = conversation.managedObjectContext?.zm_isValidContext, zm_isValidContext {}

```

Source/Model/Conversation/ZMConversationMessageWindow.swift
daehn: I think it should even be possible to make this a `let filterResult: Bool` as the compiler should be smart enough to figure out it's only initialized once.

PR #519, opened 2018-07-17
Fix/ephemeral deletion

johnxnguyen:
### Issues

An ephemeral message sent in a group conversation is not deleted on the clients of all recipients when the original sender deletes for everyone.

### Causes

When an ephemeral message is deleted, we decide who the recipients should be for the `ZMMessageDelete` message. The logic is to only send this deletion message to the self user and the sender of the ephemeral message, which I believe is to ensure that the message is still available for other users in the conversation. However, this logic fails to check if the sender and self user are one and the same, in which case no one else will delete the ephemeral.

### Solutions

If the sender and self user are the same, include all authorized users in the conversation as recipients of the deletion message.


mikeger:
@zenkins release patch

PR #520, opened 2018-07-18
Test improvements

vytis:
### Issues

* Tests using ZMTestSession were defaulting to on-disk storage, but it should be in memory. Initialising this flag got lost during a refactoring ~1 year ago.
* In tests a situation can happen that UserClient is not committed to the database and doesn't have permanentID yet, so all consecutive calls to `syncMOC.existingObject(with:)` will fail to return anything.
* Tearing down context directory a bit earlier.


vytis:
@zenkins release patch

PR #523, opened 2018-07-23
ZIOS-10243: Fix "Declined call" state

nicorsm:
### Issues

Calls rejected by the recipient should not be counted towards the amount of missed calls inside the conversations list.

### Solutions

Since we need this filter inside the conversations list only (system messages are still generated in the conversation), I've added a `isRelevant` field, in order to determine which notifications should be considered or not.


Resources/zmessaging.xcdatamodeld/zmessaging2.48.0.xcdatamodel/contents
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="14133" systemVersion="17C205" minimumToolsVersion="Xcode 4.3" sourceLanguage="Objective-C" userDefinedModelVersionIdentifier="2.48.0">
+<model type="com.apple.IDECoreDataModeler.DataModel" documentVersion="1.0" lastSavedToolsVersion="14135" systemVersion="17C205" minimumToolsVersion="Xcode 4.3" sourceLanguage="Objective-C" userDefinedModelVersionIdentifier="2.48.0">
```
mikeger: Please confirm the migration is not necessary.

nicorsm: FYI I see that Vytis has released v2.49 of the data model (that's why tests are failing). I'll move the property there.

typfel:
The name `isRelevant` is a bit unspecific I think. The property seems to similar to another property on messages `shouldGenerateUnreadDot`but as I understood it is about controlling weather it shows up in conversation list status row. Maybe a name like `relevantForConversationStatus` is more clear about what it is used for.

nicorsm:
@zenkins release major

PR #524, opened 2018-07-23
Make message window size public

alexaubry:
### Issues

The `size` property of the message window was not publicly accessible, which caused compilation problems in SE tests (cf. [Circle CI](https://circleci.com/gh/wireapp/wire-ios-sync-engine/3781?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)).

### Causes

When this class was re-written in Swift, the size property was left with internal access.


alexaubry:
@zenkins release minor

PR #526, opened 2018-07-25
Feature: I want to automatically switch to the last keyboard language selected in the conversation

billypchan:
## What's new in this PR?

Add language attribute to the conversation entity and create tests.


Source/Model/Conversation/ZMConversation+Language.swift
```diff
@@ -0,0 +1,35 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public extension ZMConversation {
+    @NSManaged internal var language: String
+
+    public var inputLanguage: String? {
```
mikeger: Can you explain the necessity of having two properties? Can't we just expose the CoreData property?

Resources/zmessaging.xcdatamodeld/zmessaging2.50.0.xcdatamodel/contents
```diff
@@ -47,6 +47,7 @@
         <attribute name="internalIsArchived" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="isSelfAnActiveMember" attributeType="Boolean" defaultValueString="YES" usesScalarValueType="NO" syncable="YES"/>
         <attribute name="isSilenced" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="NO" syncable="YES"/>
+        <attribute name="language" optional="YES" attributeType="String" syncable="YES"/>
```
mikeger: Have you tested that the app still migrating correctly?

vytis: But 2.50 is already merged to develop and is used in internal clients. You have to create a new model version and add a migration test to make sure it all works fine.

Tests/Source/Model/Conversation/ZMConversationTests+Language.swift
```diff
@@ -0,0 +1,36 @@
...
+
+class ZMConversationTests_Language : BaseZMMessageTests {
+
+    func testThatItAllowsSettingLanguageOnConversation(){
+        // given
+        let conversation = ZMConversation.insertNewObject(in: uiMOC)
+        let germanLanguage = "de-DE"
+
+        // when
+        conversation.inputLanguage = germanLanguage
```
vytis: A more thorough test would be to insert the object, change the language and the re-fetch it again. The current test would pass even if inputLanguage would be transitive property and never saved to the store.

billypchan:
@zenkins release minor

billypchan:
@zenkins release minor

PR #527, opened 2018-07-25
Asynchronous fetching of images

typfel:
## What's new in this PR?

Add methods for fetching images on a background queue.

## Notes

Some re-factoring had to be made in order to implement these methods in Swift. In particular:

* `ZMBareUser` protocol has been ported to Swift and is now called `UserType`.
* `ZMSearchUser` has been ported to Swift


Source/Model/Message/V2Asset.swift
```diff
@@ -30,7 +30,21 @@ extension String {
 
 
 @objcMembers public class V2Asset: NSObject, ZMImageMessageData {
-
+    
+    public var isDownloaded: Bool {
```
mikeger: So is it not possible to download the V2 asset any more?

typfel: Ah right :) Need to have second look at that. Was confused about what V2 referring to.

Source/Model/User/UserType.swift
```diff
@@ -0,0 +1,85 @@
...
+    var smallProfileImageCacheKey: String? { get }
+    var mediumProfileImageCacheKey: String? { get }
+    
+    var previewImageData: Data? { get }
+    var completeImageData: Data? { get }
+    
+    func requestPreviewProfileImage()
+    func requestCompleteProfileImage()
+    
+    /// AAA Wheater this user is a guest in a conversation
```
mikeger: AAA?

Source/Utilis/UserImageLocalCache.swift
```diff
@@ -162,6 +162,19 @@ extension NSManagedObjectContext
         }
     }
     
+    open func userImage(_ user: ZMUser, size: ProfileImageSize, queue: DispatchQueue, completion: @escaping (_ imageData: Data?) -> Void) {
```
mikeger: Can't this method just synchronously return an image? Not sure that I see the point of having the callback and queue when the method itself can be called in the synchronous way.

mikeger: You are right, however this other method also can be synchronous :D we can leave it, or we can also fix the public method too...

mikeger: Or is the method defined in the protocol? Was not reading it careful enough to understand.

typfel: No this method is not defined on any protocol. I'll take a second look at this.

Source/Model/Conversation/ZMConversation+DisplayName.swift
```diff
@@ -75,7 +75,7 @@ public extension ZMConversation {
         let selfUser = managedObjectContext.map(ZMUser.selfUser)
 
         let activeNames: [String] = lastServerSyncedActiveParticipants.compactMap { (user) -> String? in
-            guard let user = user as? ZMUser, user != selfUser && user.displayName?.count > 0 else { return nil }
+            guard let user = user as? ZMUser, user != selfUser && user.displayName.count > 0 else { return nil }
             return user.displayName
```
alexaubry: Are we checking the length of the string here? If yes maybe we should check `isEmpty`, as it consumes less resources IIRC

typfel: `user.displayName.count > 0` is indeed checking the length.

Source/Utilis/UserImageLocalCache.swift
```diff
@@ -162,6 +162,19 @@ extension NSManagedObjectContext
...
+        guard let cacheKey = user.resolvedCacheKey(for: size) else { return completion(nil) }
+        
+        queue.async {
+            switch size {
+            case .preview:
+                completion(self.smallUserImageCache.object(forKey: cacheKey) as? Data)
```
daehn: Right now we are calling the completion handler on the passed in queue, but are also doing a lot of work on that same queue. I am not sure if the passed in queue is supposed to be a background work queue or for the completion handler to be called on. I assume the expensive operation we want to offload is accessing the data, e.g. `self.largeUserImageCache.object(forKey: cacheKey)`.

typfel: The passed in queue is a work queue where the heavy work should happen.

daehn: @typfel ah I see, makes sense then 👍 .

Source/Model/Message/V3Asset.swift
```diff
@@ -50,7 +53,22 @@ private let zmLog = ZMSLog(tag: "AssetV3")
 
 
 @objcMembers public class V3Asset: NSObject, ZMImageMessageData {
-
+    
+    @objc(fetchImageDataWithQueue:completionHandler:)
+    public func fetchImageData(with queue: DispatchQueue!, completionHandler: ((Data?) -> Void)!) {
```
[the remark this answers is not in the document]
typfel: It is part of the `AssetProxyType`
https://github.com/wireapp/wire-ios-data-model/blob/aae30c9b31c93c9978d3a6fd6454188ea63363d5/Source/Model/Message/V3Asset.swift#L46-L47

typfel:
@zenkins release major

PR #528, opened 2018-07-26
ZIOS-10243: Declined call - exclude rejected calls from the amount of unread messages

nicorsm:
## What's new in this PR?

Changes about the declined call behavior in conversation list were not affecting the counter badge, so I've:

- Moved `relevantForConversationStatus` extension to datamodel and renamed to `messageIsRelevantForConversationStatus`.
- Added filtering of `unreadMessagesIncludingInvisible` by `messageIsRelevantForConversationStatus` property, so rejected calls are not counted towards the amount of unread messages.
- Overrided `initWithNonce:managedObjectContext:` of `ZMSystemMessage` in order to set the default value of `relevantForConversationStatus` to `YES` and avoid unnecessary variable setup for all the tests
- Added tests


Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -315,7 +315,7 @@ extension ZMConversation {
                                              ZMMessageServerTimestampKey, lastReadServerTimestamp as NSDate)
         fetchRequest.sortDescriptors = ZMMessage.defaultSortDescriptors()
         
-        return managedObjectContext.fetchOrAssert(request: fetchRequest)
+        return managedObjectContext.fetchOrAssert(request: fetchRequest).filter { $0.messageIsRelevantForConversationStatus }
     }
```
alexaubry: If you update `WireUtilities`, you can use the new key-path based filtering here:

~~~swift
filter(\.messageIsRelevantForConversationStatus)
~~~

nicorsm:
@zenkins release minor

PR #529, opened 2018-07-27
Guarantee that the completion handler is always called

typfel:
## What's new in this PR?

Always call completion handler.

### Issues

Completion handler for fetching image data wasn't always called which causes issues for the consumer.


typfel:
I think a test for this would for this is of limited use since it would still pass if the rule for early return changes.

typfel:
@zenkins release patch

PR #530, opened 2018-07-30
Move image request methods to data type protocols

typfel:
### Issues

Image were not being downloaded.

### Causes

The method for requesting images was living on the base message protocol `ConversationMessage` and most of the time the underlying object is implementing the message data type protocol (image, file, text). But this is not always the case for images which resulted in a cast to fail.

### Solutions

Move image download methods to the relevant data type protocols to avoid having to rely on casting.


typfel:
@zenkins release major

PR #533, opened 2018-08-06
Add non-mutable user validators for better Swift import

alexaubry:
### Issues

Validators for `ZMUser` input fields take a reference to an NSString pointer to check the validity of user input, and change the string when it can normalize it.

While this may be useful in some cases, this API can be harder to use when we just want to check if text is valid without checking the updated value; especially in Swift where it is imported with an odd signature that does not return a Boolean value:

~~~swift
class func validatePassword(_ ioEmailAddress: AutoreleasingUnsafeMutablePointer<NSString?>?) throws
~~~

For example, to check if a password is valid, we'd need to write:

~~~swift
func isValidPassword(_ password: String?) -> Bool {
    var validationPassword: NSString? = password as NSString?

    do {
        try ZMUser.validatePassword(& validationPassword)
    } catch {
        return false
    }
}
~~~

### Solutions

We add non-mutable validators for:

- name
- email
- password
- phone number
- phone verification code

New usage:

~~~swift
let nameValid: Bool = ZMUser.isValidName("Alexis")
let passwordValid: Bool = ZMUser.isValidPassword("s0_s3KuRe!")
~~~

## Dependencies

Needs releases with:

- [x] [wire-ios-utilities #54](https://github.com/wireapp/wire-ios-utilities/pull/54)


alexaubry:
@zenkins release minor

PR #534, opened 2018-08-09
Unordered relation Conversation->Messages

mikeger:
### Issues

The performance of adding the message to the conversation was degrading in the bigger (50K messages+) conversations.

### Causes

During the WWDC, we identified that using the ordered relations in CoreData can cause the performance degradation.

### Solutions

1. Migration:
    1.a Rename the relation `conversation.messages` to `conversation.allMessages`.
    1.b Change the relation `conversation.allMessages`, `conversation.hiddenMessages` to be not ordered.

2. (Re-)Introduce property `conversation.messages` backed by the fetch request (with the fetch limit). This would allow to keep the old code working.
3. Remove all code for manually sorting the messages (sortedInsert...).
4. Remove custom merge policy, since it's needed to resort messages and it's loading all messages into memory.
5. Remove `messageWindow`.
6. Check every usage of `conversation.messages` to make sure it's OK to look only on the most recent messages subset. For example: when inserting system message, we are checking if the last message is of the same type, so not necessary to load all messages, we try to find the dependency message for this one, it's also fine not to look on all the messages, etc.


Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -325,20 +327,18 @@ extension ZMConversation {
...
+        return systemMessage
     }
     
-
-    
     /// Returns a timestamp that is shortly (as short as possible) before the given message,
     /// or the last modified date if the message is nil
     fileprivate func timestamp(before: ZMMessage?) -> Date? {
         guard let timestamp = before?.serverTimestamp ?? self.lastModifiedDate else { return nil }
         // this feels a bit hackish, but should work. If two messages are less than 1 milliseconds apart
```
vytis: Maybe comment also needs updating, 1 msec is gone now

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -347,7 +347,7 @@ extension ZMConversation {
         guard let timestamp = after?.serverTimestamp ?? self.lastModifiedDate else { return nil }
         // this feels a bit hackish, but should work. If two messages are less than 1 milliseconds apart
         // then in this case one of them will be out of order
```
vytis: Same here

Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
```diff
@@ -308,9 +308,9 @@ class DatabaseMigrationTests: DatabaseBaseTest {
         }
     }
     
-    func testThatItPerformsMigrationFrom_Between_2_24_1_and_2_50_0_ToCurrentModelVersion() {
+    func testThatItPerformsMigrationFrom_Between_2_24_1_and_2_52_0_ToCurrentModelVersion() {
```
vytis: A nitpick but maybe there is a better name for this test so that we would not need to update it every time...

Tests/Source/Model/Conversation/ZMConversationTests+SecurityLevel.m
```diff
@@ -1086,12 +1086,15 @@ - (void)testThatItDoesNotMoveExistingDegradedMessageWhenRemoteParticpantsAdd_Oth
     XCTAssertEqual(((ZMSystemMessage *)conversation.messages.lastObject).systemMessageType, ZMSystemMessageTypeNewClient);
     XCTAssertTrue([((ZMSystemMessage *)conversation.messages.lastObject).addedUsers isEqualToSet:unverifiedUsers]);
     
+    [NSThread sleepUntilDate:[[NSDate date] dateByAddingTimeInterval:0.2]];
```
vytis: What was happening here that this was needed?

Tests/Source/Utils/FetchedObjectsArrayTests.swift
```diff
@@ -0,0 +1,117 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+import XCTest
+@testable import WireDataModel
+
+public class FetchedObjectsArrayTests: ZMBaseManagedObjectTest {
```
vytis: Maybe worth adding some tests to see what happens when some messages are faulted?

Source/Model/Conversation/ZMConversation+RecentMessages.swift
```diff
@@ -0,0 +1,68 @@
...
+        set {
+            _recentMessagesFetcher = newValue
+        }
+        get {
+            if let currentFetcher = _recentMessagesFetcher as? FetchedObjectsArray<ZMMessage> {
+                return currentFetcher
+            }
+            
+            let fetchRequest = NSFetchRequest<ZMMessage>(entityName: ZMMessage.entityName())
+            fetchRequest.fetchLimit = 100
```
vytis: Why we need to set this value and why it is 100?

Source/Model/Conversation/ZMConversation+RecentMessages.swift
```diff
@@ -0,0 +1,68 @@
...
+            fetchRequest.sortDescriptors = [NSSortDescriptor(key: #keyPath(ZMMessage.serverTimestamp), ascending: false)]
+            
+            let newFetcher = try! FetchedObjectsArray<ZMMessage>(on: self.managedObjectContext!, fetchRequest: fetchRequest)
+            self.recentMessagesFetcher = newFetcher
+            return newFetcher
+        }
+    }
+    
+    @objc public var recentMessages: [ZMConversationMessage] {
+        return recentMessagesFetcher.reversed()
```
vytis: Are there any other places where we use recentMessagesFetcher? Maybe then sort the fetch request differently and there will be no need for reversing? Does the current approach impact performance in any way, i.e. it needs to fetch all to be able to reverse the collection?

Source/Utilis/FetchedObjectsArray.swift
```diff
@@ -0,0 +1,84 @@
...
+
+public final class FetchedObjectsArray<Element: NSFetchRequestResult>: NSObject, Collection, NSFetchedResultsControllerDelegate {
+    
+    public typealias Index = Int
+    
+    public var startIndex: Int {
+        return 0
+    }
+    
+    public var endIndex: Int = 0
```
vytis: Why this is a stored property instead of directly asking fetch controller? Can you elaborate a bit how this mapping from fetch controller to collection works - e.g. what happens if `endIndex` changes while we iterate over a collection? Or this cannot happen because we somehow control when it does the fetch?

mikeger:
@vytis message window is gone and will be replaced with `NSFetchedResultsController` directly in UI, since it's telling the `UITableView` the exact updates. This is how it's working: https://github.com/wireapp/wire-ios/compare/feature/fetch-results-controller?expand=1

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
mikeger: True, not needed any more.

Tests/Source/Model/Conversation/ZMConversationTests+SecurityLevel.m
mikeger: Debug leftover 🐛

Source/Model/Conversation/ZMConversation+RecentMessages.swift
mikeger: If we won't set it, the code that is accessing `.messages` can randomly walk over the huge conversations loading all messages. 100 is the magic number, basically per default we only see up to 100 messages in the past in the conversation with the methods that access `.messages`.

I checked the places where we traverse messages and it should be working, in the places where it's not, it is replaced with another fetch request.

Source/Model/Conversation/ZMConversation+RecentMessages.swift
mikeger: Indeed, can be reversed by design.

Source/Utilis/FetchedObjectsArray.swift
mikeger: It's updated in `updateCount()`: either after the initial fetch or after the update that controller is populating with `controllerDidChangeContent(_:)` callback. It cannot change while we iterate, unless we do something bad like spinning the run loop, since it's limited to run on one thread.

We can also just go to `fetchedObjects?.count` and read the count from there directly, I think it would actually be easier to understand, so I'll change it to that implementation.

Source/Model/Conversation/ZMConversation+RecentMessages.swift
mikeger: Sorry, was confused, actually we cannot, and this is why:

Imagine we have the messages ordered _ascending_:

1
2
3
4
5
6
7
8

In the nutshell the fetch request is like SQL, where you call QUERY ... OFFSET X LIMIT Y. We say please return us 5 messages, ordered ascending, limit 5: it would return 1-5. To get what we want, we'll need to pass in the offset 3 limit 5, and to calculate it we need to do another fetch. Therefore I think it's easier to fetch descending with the limit and reverse the array.

mikeger:
So more fixing needed for `enqueueUpdateLastRead`, since it's still slow to scroll up the convo

Source/Model/Conversation/ZMConversation+RecentMessages.swift
vytis: Makes sense!

mikeger:
After discussion with @marcoconti83 

1. Consider removing `messages` and directly using `recentMessages`.
2. Problem with user going too far in the past in the conversation history: as we are fetching only N last messages in the UI, we can only control the count of messages to fetch. Therefore if user goes all the way to the first message in the conversation then all the messages would be fetched with the fetch results controller. Check
2.a Current memory usage in this scenario.
2.b Same + send memory warning (to check if CoreData would fault the not used messages).
2.c Same + fault manually invisible messages.
3. Refactor `lastUnreadMessage(olderOrEqualThan:)` to do 2 fetches, so it would use the indexes.

Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -210,8 +210,8 @@ extension ZMConversation {
             hasUnreadUnsentMessage = false
         }
         
-        guard let messageTimestamp = message.serverTimestampIncludingChildMessages, let unreadTimestamp = lastUnreadMessage(olderOrEqualThan: messageTimestamp)?.serverTimestamp else { return }
-        
+        guard let messageTimestamp = message.serverTimestampIncludingChildMessages,
+            let unreadTimestamp = lastUnreadMessage(olderOrEqualThan: messageTimestamp)?.serverTimestamp else { return }
```
typfel: So did you try removing the `lastUnreadMessage()`?

mikeger: I tried, it's fast (obviously), but Marco advised to refactor it and keep the logic, partially because we don't know what it is accounting for, partially because it can be implemented using the indexes.

typfel: Did you try running the tests after removing it? Anyway good that we can solve it using the indexes but long term it doesn't seem sustainable to keep code around because we don't know what it does. Maybe we could try to removing it separately after this is merged.

mikeger: I think it's a good idea to remove it and test how the client would be working, but in context of those bigger changes I would be hesitant to also change this part, potentially making the bugs easier to spot. I think it can be hard to isolate changes.

marcoconti83: I think that's a good plan (remove separately to isolate any issue)

vytis:
Codecov was experiencing downtime yesterday, I think it should be fine now - restarted tests.

Source/Model/Conversation/ZMConversation+DeleteOlderMessages.swift
```diff
@@ -0,0 +1,41 @@
...
+            return
+        }
+        
+        let fetchRequest = NSFetchRequest<NSFetchRequestResult>(entityName: ZMMessage.entityName())
+        fetchRequest.predicate = NSPredicate(format: "%K <= %@", #keyPath(ZMMessage.serverTimestamp), clearedTimeStamp as CVarArg)
+        
+        let result = try! managedObjectContext.fetch(fetchRequest) as! [ZMMessage]
+
+        for element in result {
+            managedObjectContext.delete(element)
```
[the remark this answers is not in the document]
mikeger: There is a batch delete request, but it is was not propagating changes from my limited testing in October.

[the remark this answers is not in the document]
mikeger: As I think about it, may be we don't need the change notification in this case...

Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -376,21 +381,64 @@ extension ZMConversation {
     }
     
     fileprivate func lastUnreadMessage(olderOrEqualThan date: Date) -> ZMMessage? {
-        guard let managedObjectContext = managedObjectContext else { return nil }
-        
+        return unreadMessagesIncludingInvisible(fetchLimit: 1, before: date, order: .descending).first
+    }
+    
+    enum Order {
```
typfel: Could be `fileprivate`?

Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -376,21 +381,64 @@ extension ZMConversation {
...
+        }
+        
+        if let limit = fetchLimit, allMessages.count > limit {
+            allMessages.removeLast(allMessages.count - limit)
+        }
+        
+        return allMessages
+    }
+    
+    enum MessageTypeKey: String {
```
typfel: Could be `fileprivate`?

mikeger:
@typfel I will create the migration test for this one.

PR #535, opened 2018-08-10
Treat images larger than 5 MB as files

daehn:
## What's new in this PR?

* Treat images larger than 5 MB as files to avoid auto downloading them.
* The exact threshold is still subject to change.


daehn:
I have to verify this doesn't break old larger image that already where downloaded.

typfel:
I'd say it's fine to potentially break already downloaded images bigger than the limit. They would just be downloaded again when requested and we are anyway storing downloading images/files in the cache directory.

typfel:
@zenkins release minor

PR #536, opened 2018-08-10
Feature: PassKit file support

billypchan:
## What's new in this PR?

Identify PassKit pass file type.


vytis:
A small test would be good here maybe

billypchan:
@zenkins release minor

PR #538, opened 2018-08-16
Show full name for the service user

mikeger:
### Issues

We are not creating the initials for the service user.


mikeger:
@zenkins release patch

PR #539, opened 2018-08-17
Fix memory leak caused by not taking ownership of object when expected

typfel:
### Issues

We are leaking the return value from `UTTypeCreatePreferredIdentifierForTag`

### Causes

the `Create` in `UTTypeCreatePreferredIdentifierForTag` indicate that this function will create an object which the caller is responsible for releasing.

### Solutions

 Take ownership of the return value by calling `takeRetainedValue`.


typfel:
@zenkins release patch

PR #540, opened 2018-08-17
Modernize unregistered user structure + input normalization

alexaubry:
### Issues

The data structures used for the unregistered user, `ZMIncompleteRegistrationUser` and `ZMCompleteRegistrationUser` did not support codes for e-mail users, and did not allow developers to verify the input from Swift code without type casting _pirouettes_.

#### Unregistered User

We create a single class for unregistered users, named `UnregisteredUser`. It is type safe, and only allows to put either e-mail or phone credentials, making it less prone to errors.

Prior to these changes, the migration between and incomplete user handled user input validation. With this version, the burden or verification lies on the object updating the unregistered user properties.

The incomplete → complete validation also checked that all phone or e-mail credentials were set. This problem is fixed with the `UnregisteredUser.Credentials` enum. This enum hold either phone or e-mail credentials, it is not possible to have a phone number and a passcode, for example.

We only need to expose a `isComplete` boolean property, that we check before making the request to register to the server, to make sure all the data is provided.

#### Data Validation

Now that the burden of validation is on the consumers of the unregistered user, we need a way to provide type safety and ease of use to Swift parts of the code using the unregistered user. For that, we create two types:

- `ZMPropertyNormalizationResult` = a generic Objective-C wrapper around the result of the validation API (contains the BOOL result, and the error and validated value passed as inout when doing the validation)
- `NormalizationResult` = a generic Swift state enum, that says if the value is either valid (provides the value), invalid because of a known reason (provides the error code), or invalid for no known reason.

The use of the generic Swift enum makes it easy to check if the value is valid at call site, because we don't need to check for both result, value nullability and error there.

This API will typically be called from view controllers and authentication coordinators.


alexaubry:
@zenkins release major

PR #541, opened 2018-08-19
ZIOS-9535: Received ephemeral message with link is shown without preview

nicorsm:
### Issues

Received ephemeral messages were shown without link previews, even if the preview was visible on other platforms.

### Causes

This was caused by a condition during the update of the message that was considering only "normal" messages, and not ephemeral ones. 

### Solutions

I've added conditions for the ephemeral messages + a test.


mikeger:
@zenkins relese patch

mikeger:
@nicorsm mr. @daehn is on vacation for 3 weeks.

PR #543, opened 2018-08-20
ZIOS-10335: Attachments are not deleted when removed individually

nicorsm:
### Issues

A copy of an unencrypted downloaded file that resides on the `/tmp` folder of the file system is not removed when deleting the message from a conversation.

### Causes

The file was not deleted. The only copy of the (encrypted?) file that got deleted was the one inside the `zm_fileAssetCache` folder (which is in `/Library/Caches`).

### Solutions

I've added some logic (and a test) to delete both copies of the file.


Source/Model/Message/ZMAssetClientMessage+Deletion.swift
```diff
@@ -29,6 +29,12 @@ extension ZMAssetClientMessage {
 
     func deleteContent() {
         self.managedObjectContext?.zm_fileAssetCache.deleteAssetData(self)
+        
+        if let url = temporaryDirectoryURL,
+            FileManager.default.fileExists(atPath: url.path) {
+            try? FileManager.default.removeItem(at: url)
```
mikeger: Should it delete all the assets in that directory when one asset is deleted?

nicorsm: Yes, actually we're creating a folder for each file, based on the `cacheKey`, as you can see here:
screen shot 2018-08-20 at 15 51 00

I thought it was good to clear the folder in order to keep the parent folder clean. Do you know any case where we use those folders for 2+ files?

mikeger: Should be fine, I haven't noticed that `temporaryDirectoryURL` is not the same for all files (looking on the implementation).

nicorsm:
@zenkins release patch

PR #544, opened 2018-08-22
Find conversation with service

vytis:
### Issues

We need a way to find an existing conversation with given service so we would not create duplicates.

### Solutions

The predicate for the matching conversation is pretty complicated, but basically is:
* Group conversation in a team
* You are part of it
* There is only one other participant
* The participant is a service with matching identifier

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Services.swift
+++ b/Source/Model/Conversation/ZMConversation+Services.swift
@@ -24,7 +24,7 @@ extension ZMConversation {
         guard let serviceID = service.serviceIdentifier else { return nil }
         let sameTeam = predicateForConversations(in: team)
         let groupConversation = NSPredicate(format: "%K == %d", ZMConversationConversationTypeKey, ZMConversationType.group.rawValue)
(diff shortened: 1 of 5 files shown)


Source/Model/Conversation/ZMConversation+Services.swift
```diff
@@ -0,0 +1,37 @@
...
+
+import Foundation
+
+extension ZMConversation {
+    public class func existingConversation(in moc: NSManagedObjectContext, service: ServiceUser, team: Team?) -> ZMConversation? {
+        guard let team = team else { return nil }
+        guard let serviceID = service.serviceIdentifier else { return nil }
+        let sameTeam = predicateForConversations(in: team)
+        let groupConversation = NSPredicate(format: "%K == %d", ZMConversationConversationTypeKey, ZMConversationType.group.rawValue)
+        let selfIsActiveMember = NSPredicate(format: "isSelfAnActiveMember == YES")
```
typfel: No key path here?

vytis: Copied from somewhere else, makes sense.

vytis:
@zenkins release minor

PR #545, opened 2018-08-28
Open service conversation: skipping conversations with a name

vytis:
### Issues

The existing logic of opening conversation with a service was not skipping conversations with custom defined name which made it easy to have multiple conversations with a service. "Open conversation" button would then open the most recent of all conversations with a service.

### Solutions

We filter out conversations with user defined name, so guestrooms and regular conversations created from UI are excluded. This (in most circumstances) leaves only a single conversation that we can open.


mikeger:
@zenkins release patch

PR #548, opened 2018-08-29
ZIOS-10445: Adjust VoIP networking session service type

alexaubry:
## What's new in this PR?

iOS 10 provides a new session service type for call signaling. We use it instead of the VoIP type when available, as VoIP service type should only be used for voice calls (which go don't go through this session).

### Dependencies

- [x] https://github.com/wireapp/wire-ios-request-strategy/pull/71


mikeger:
@alexaubry not sure if the upstream PRs are necessary any more, since the API is not changed? I think you could have just bumped the transport in UI directly?

alexaubry:
@mikeger Good call. Though there was a test that broke in SE, so I will keep only that PR.

PR #551, opened 2018-08-30
Resort messages immediately after insertion to avoid saves in between

typfel:
### Issues

The integration test `[SendAndReceiveMessages testThatItAppendsClientMessages]` was failing randomly.

### Causes

In this test two messages are received and then verified to be in the correct order. The messages were sometimes in the wrong order because the the conversation observer fired too early before the the messages had been sorted. This could happen since the SE context was saved in between that the new message was inserted and that the messages were sorted.

### Solutions

Sort messages immediately after inserting the message.


typfel:
@zenkins release patch

PR #555, opened 2018-09-14
Added UserID to mention (needed in UI)

mikeger:
### Issues

User ID is needed to create the clickable link to user profile.

### DIFF
--- a/Source/Model/Message/Mention.swift
+++ b/Source/Model/Message/Mention.swift
@@ -23,6 +23,7 @@ public class Mention: NSObject {
     
     public let range: NSRange
     public let user: UserType
+    public let userId: UUID
     
     init?(_ protobuf: ZMMention, context: NSManagedObjectContext) {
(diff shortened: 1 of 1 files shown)


mikeger:
Hey, in UI I need to create URL that must be clickable in order to open the user profile; I guess it’s the valid case for exposing the user ID? Currently there is no way to identify the UserType object. Also I cannot open the user profile without knowing the ID.

Sent from a Mobile Device

daehn:
Yeah, I understand that. What I meant is that in order to identify users with a link we could also use temporary identifiers on the UI side and have a map `[UUID: UserType]` to get the correct user when a link is tapped. AFAIK the user profile screen can be initialised with a `UserType` directly.

mikeger:
@daehn what I see in `ProfileViewController`:
`initWithUser:(id<UserType, AccentColorProvider>)user`

May be you can explain what i

PR #557, opened 2018-09-17
Refactor add messages

typfel:
## What's new in this PR?

Method for appending messages and creating generic messages have been re-factored to be more composable.


Source/Model/Message/ZMClientMessage.m
```diff
@@ -237,7 +237,7 @@ - (void)expire
 - (void)resend
 {
     if (self.genericMessage.hasEdited) {
-        NOT_USED([ZMMessage edit:self newText:self.textMessageData.messageText]);
+        NOT_USED([ZMMessage edit:self newText:self.textMessageData.messageText mentions:@[] fetchLinkPreview:YES]); // TODO jacob why can't we just resend the message?
```
mikeger: Seems like a todo, or not necessary any more?

Source/Utilis/Protos/ZMGenericMessage+Helper.swift
```diff
@@ -416,6 +384,55 @@ public extension ZMOtrAssetMeta {
...
+    
+    public func setEphemeralContent(on builder: ZMEphemeralBuilder) {
+        builder.setAsset(self)
+    }
+    
+}
+
+extension ZMImageAsset: EphemeralMessageContentType {
+    
+//    public static func imageAsset(mediumProperties: ZMIImageProperties, processedProperties: ZMIImageProperties, encryptionKeys: ZMImageAssetEncryptionKeys, format: ZMImageFormat) -> ZMImageAsset {
```
mikeger: Can be removed?

Tests/Source/Model/Conversation/ZMConversationTests+Messages.swift
```diff
@@ -118,31 +118,13 @@ class ZMConversationMessagesTests: ZMConversationTestsBase {
...
-    {
-        // given
-        let conversation = ZMConversation.insertNewObject(in: self.uiMOC)
-        conversation.remoteIdentifier = UUID()
-        let start = self.uiMOC.insertedObjects
-    
-        // when
-        var message: Any? = nil
-        self.performIgnoringZMLogError {
-            message = conversation.appendMessage(withText: nil)
```
mikeger: Good that we don't need this kind of test any more 👍

Source/Model/Message/ZMClientMessage.m
typfel: Well I still don't why we re-generate the edited when re-sending it?

typfel: Let's try removing it and see it what happens :)

typfel: After talking to @daehn I'll remove this in a separate PR

PR #559, opened 2018-09-18
Store mentions in message draft

daehn:
## What's new in this PR?

* Store mentions in message draft field introduced on `ZMConversation`.

### This PR is still depending on:
- [ ] A migration test for the new model version has to be added.
- [ ] Comments need to be added to the added types.

### DIFF
--- a/Source/Public/DraftMessage.swift
+++ b/Source/Public/DraftMessage.swift
@@ -1,0 +1,105 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


Source/Public/DraftMessage.swift
```diff
@@ -0,0 +1,105 @@
...
+            StorableMention(range: range, userIdentifier: $0)
+        }
+    }
+}
+
+@objc extension ZMConversation {
+    
+    @NSManaged var draftMessageData: Data?
+
+    public var draftMessage: DraftMessage? {
```
vytis: You can do `primitiveDraftMessage` - it's a special prefix for CoreData. An example of the same pattern here - https://github.com/wireapp/wire-ios-data-model/blob/develop/Source/Model/UserClient/UserClient.swift#L88-L110

PR #562, opened 2018-09-25
Fix some link previews not being rendered

typfel:
### Issues

Some Link previews were not rendered when containing a valid link preview

### Causes

We were expecting a link preview protobuf to either to contain a tweet or an article but this is no longer required. A link preview is by default an article and can be extended with more meta data (tweet).

### Solutions

Don't verify that the link preview contains an article.


typfel:
@zenkins release patch

PR #563, opened 2018-09-28
Keep mentions when updating text message after setting image data

typfel:
### Issues

Mention were disappearing when sending mention together with links

### Causes

We didn't copy the mentions over when setting the downloaded link preview image data on a message.

### Solutions

Copy the mention over.


typfel:
@zenkins release patch

PR #564, opened 2018-10-02
Feature/mentions open notification

typfel:
## What's new in this PR?

Add `firstUnreadMessageMentioningSelf`  property on `ZMConversation` which will be used to find the scroll position when opening notification containing a mention.


typfel:
@zenkins release minor

PR #565, opened 2018-10-02
Mute status for mentions: new API

mikeger:
## What's new in this PR?

Initial version, missing the sync logic and massive amount of tests.


Source/Notifications/ObjectObserverTokens/ConversationChangeInfo.swift
```diff
@@ -93,8 +93,8 @@ extension ZMConversation : ObjectInSnapshot {
         return changedKeysContain(keys: #keyPath(ZMConversation.isArchived))
     }
 
-    public var isSilencedChanged : Bool {
-        return changedKeysContain(keys: #keyPath(ZMConversation.isSilenced))
+    public var mutedMessageTypesChanged : Bool {
```
typfel: Doesn't it more sense to name this `mutedStatusChanged` to match `mutedStatus`?

typfel: Ah or `mutedStatus` is private and `mutedMessageTypes` is the public interface.

Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -0,0 +1,53 @@
...
+public struct MutedMessageTypes: OptionSet {
+    public let rawValue: Int32
+    
+    public init(rawValue: Int32) {
+        self.rawValue = rawValue
+    }
+    
+    public static let none     = MutedMessageTypes(rawValue: 0)
+    ///
+    public static let all      = MutedMessageTypes(rawValue: (1 << 0) ^ (1 << 1))
```
vytis: Do we need `all` at all? Seems to be very prone to error when we add another item. Also this could simply be `[.none, .mentions]`

Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -0,0 +1,53 @@
...
+    ///
+    public static let mentions = MutedMessageTypes(rawValue: 1 << 1)
+}
+
+public extension ZMConversation {
+    @NSManaged @objc public var mutedStatus: Int32
```
vytis: This should probably be optional, we need to be able to distinguish between "old" muted conversations and new ones

vytis:
Not sure if I understand how it will deal with conversations which don't have the `otr_muted_status` at all? It will be present only after we change muting from the new client. If that doesn't happen we need to fall back to use the old fields.

Source/Model/Conversation/ZMConversation+Mute.swift
mikeger: It cannot be [.none, .mentions], since none == 0, and mentions == 2, this must be 3, I guess I need to add hidden value for `normal`.

mikeger: Oops, I think some values are confused, let me add more tests.

mikeger: So value for mentions is not correct, will push the fix.

Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -0,0 +1,53 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/// Defines what kind of messages are muted.
+public struct MutedMessageTypes: OptionSet {
+    public let rawValue: Int32
```
alexaubry: Is there a reason why this needs to be an option set?

mikeger: Yes, since the backend value is the bit mask, to better represent the backing value.

vytis:
@mikeger will it work the other way around? e.g. will older clients get all the info needed? We are in a way "loosing" information by removing the old field, but maybe it's not a problem in this case.

mikeger:
@vytis so the new client would set the both fields, where the old field value can always be inferred from the new one, I think in that case we don't need to keep the old value, since it seems redundant. If you see the reason to keep it then please describe it.

vytis:
Actually can't think of any cases where this would go wrong, so it should work!

mikeger:
@zenkins release major

PR #566, opened 2018-10-05
Added logic to check if message is silenced

mikeger:
### Issues

Sometimes it's necessary to check if the message has to generate the notification.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Mute.swift
+++ b/Source/Model/Conversation/ZMConversation+Mute.swift
@@ -34,7 +34,7 @@ public struct MutedMessageTypes: OptionSet {
     
     /// Only non-mentions are muted.
     public static let nonMentions = MutedMessageTypes(rawValue: 1 << 0)
-    private static let mentions = MutedMessageTypes(rawValue: 1 << 1)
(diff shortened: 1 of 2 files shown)


Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -34,7 +34,7 @@ public struct MutedMessageTypes: OptionSet {
     
     /// Only non-mentions are muted.
     public static let nonMentions = MutedMessageTypes(rawValue: 1 << 0)
-    private static let mentions = MutedMessageTypes(rawValue: 1 << 1)
+    public static let mentions = MutedMessageTypes(rawValue: 1 << 1)
```
johnxnguyen: Just out of curiosity, what does the option represent? That only mentions are muted?

mikeger: From the initial perspective does not really make sense to make it public, since we agreed never to use it (and it has no logical sense: it's only blocking mentions), however in the code I found the need to check if the bit is set, then I need this option to check against.

mikeger:
@zenkins release patch

PR #569, opened 2018-10-05
Getter and setter to keep only mentions only for team users

mikeger:
### Issues

The getter would make sure that the new value is only set and correct after the migration for the team users.

### DIFF
--- a/Tests/Source/Model/Conversation/ZMConversationTests+Transport.m
+++ b/Tests/Source/Model/Conversation/ZMConversationTests+Transport.m
@@ -57,7 +57,7 @@
                                        archivedRef:nil
                                         isSilenced:NO
                                        silencedRef:nil
-                                    silencedStatus:@(3)
+                                    silencedStatus:@(MutedMessageOptionValueMentions)
                                             teamID:nil
                                         accessMode:@[]
                                         accessRole:@"non_activated"];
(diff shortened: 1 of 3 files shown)


Tests/Source/Model/Conversation/ZMConversationTests+Transport.m
```diff
@@ -419,7 +422,7 @@ - (void)testThatItUpdatesItselfFromTransportDataForTeamConversation
                                                            archivedRef:nil
                                                             isSilenced:NO
                                                            silencedRef:nil
-                                                        silencedStatus:@(3)
```
typfel: I think `3` the combination of `MutedMessageOptionValueMentions` and MutedMessageOptionValueNonMentions`

mikeger: Yeah, see it now.

mikeger:
@typfel please review

Tests/Source/Model/Conversation/ZMConversationTests+Transport.m
```diff
@@ -57,7 +57,7 @@ - (NSDictionary *)payloadForMetaDataOfConversation:(ZMConversation *)conversatio
                                        archivedRef:nil
                                         isSilenced:NO
                                        silencedRef:nil
-                                    silencedStatus:@(3)
+                                    silencedStatus:@(MutedMessageOptionValueMentions + MutedMessageOptionValueNonMentions)
```
typfel: Should this be `MutedMessageOptionValueMentions | MutedMessageOptionValueNonMentions`?

mikeger: I'll create the new property for it: the text seems too long with either plus or `or`.

PR #570, opened 2018-10-08
Lightweight migration for muted status

mikeger:
### Issues

The migration to version 2.54 of the store was slow.

### Causes

The migration was done using `xcmappingmodel`.

### Solutions

Simply rename the property, since the SQLite is using the INTEGER column type for booleans internally .


WireDataModel.xcodeproj/xcshareddata/xcschemes/WireDataModel.xcscheme
```diff
@@ -89,6 +89,10 @@
             argument = "-com.apple.CoreData.ConcurrencyDebug 1"
             isEnabled = "YES">
          </CommandLineArgument>
+         <CommandLineArgument
```
typfel: Revert?

mikeger: Why? It's useful, and disabled for now with the ☑️

mikeger:
@zenkins release patch

PR #571, opened 2018-10-08
Do not unarchive conversation if fully muted and mentioned

mikeger:
### Issues

We had a bug that the conversation was unarchived when self user is mentioned, even if the conversation is fully muted.


mikeger:
@zenkins release patch

PR #572, opened 2018-10-09
Fix: Team one-to-one show in group section

billypchan:
## What's new in this PR?

This PR is a follow-up of https://github.com/wireapp/wire-ios-sync-engine/pull/873

Filter the conversations that it is a 1 to 1 conversation (lastServerSyncedActiveParticipants == 1 and userDefinedName is nil)

PR #573, opened 2018-10-11
Fix: Search results don't contain group conversations where the current user participates

billypchan:
### Issues

After https://github.com/wireapp/wire-ios-data-model/pull/572, group conversation contains the matching user is not shown.

### Causes

The predicate check for TeamRemoteIdentifierKey, which is not the valid key.

### Solutions

User TeamRemoteIdentifierDataKey instead of TeamRemoteIdentifierKey.


typfel:
Would be nice to have a test case for this in `ZMConversationTests (SearchQuerys)`

mikeger:
@zenkins release patch

PR #578, opened 2018-10-18
ZIOS-10734: Fix user logout after updating to new login stack

alexaubry:
### Issues

When updating to a build that includes the login changes, users get logged out.

### Causes

This happened because we were not using the correct JSON keys for decoding the account in the account store. So the SE tried to login using the legacy store migrator, which couldn't find any account either, which caused the logout.

### Solutions

We fix the JSON Coding Keys for the account model to use the ones that were used in the previous version.

### DIFF
--- a/Source/Model/Account.swift
+++ b/Source/Model/Account.swift
@@ -45,6 +45,16 @@ public final class Account: NSObject, Codable {
         }
     }
 
+    enum CodingKeys: String, CodingKey {
+        case userName = "name"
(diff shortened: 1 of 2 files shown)


alexaubry:
@mikeger I added two test cases for decoding. Let me know if you think they cover enough.

mikeger:
Looking good!

alexaubry:
@zenkins release patch

PR #579, opened 2018-10-22
ZIOS-10707: Subtitle is not updated properly if the message is edited

nicorsm:
### Issues

Subtitles inside the cells of the conversation list was not updated after editing a message and posting a mention.

### Solutions

I've added `messagesChanged` to the whitelist of allowed change types inside of `conversationDidChange()`.


mikeger:
Would be good to have the test for this change!

mikeger:
@typfel can you check this one? I think we are not sending the update intentionally, since the cell is observing the message changes?

typfel:
@mikeger from what I can see is the cell no observing the messages directly.

nicorsm:
@zenkins release minor

PR #580, opened 2018-10-23
Add method for computing a hash of the message content

typfel:
## What's new in this PR?

Add `hashOfContent` property on the message. This returns a SHA256 hash of the message if contains one of following content types: text, location and asset.


Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -64,6 +64,12 @@ import Foundation
         add(ZMGenericMessage.message(content: metadata.asset, nonce: nonce, expiresAfter: timeout))
     }
     
+    public override var hashOfContent: Data! {
```
billypchan: Maybe it is better to return type "Data?" ?

typfel: This `!`  comes from the facts that the property is defined in `ZMOTRMessage.h` which needs be annotated with nullability. Something I'd like to do but preferably not in this PR since it will trigger a cascade of changes.

typfel: @mikeger right that should also work. Will do that.

Tests/Source/Utils/ZMGenericMessageTests+Hashing.swift
```diff
@@ -0,0 +1,141 @@
...
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
+    
+    // MARK: - Text
+
+    func testText1() {
```
vytis: Some more descriptive names of tests would be good

Tests/Source/Utils/ZMGenericMessageTests+Hashing.swift
```diff
@@ -0,0 +1,141 @@
...
+
+class ZMGenericMessageTests_Hashing: XCTestCase {
+
+    override func setUp() {
+        // Put setup code here. This method is called before the invocation of each test method in the class.
+    }
+
+    override func tearDown() {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+    }
```
vytis: Could be removed alltogether

Source/Model/Message/ZMOTRMessage.m
```diff
@@ -49,6 +49,12 @@ - (NSSet *)ignoredKeys;
     return [keys setByAddingObjectsFromArray:@[DeliveredKey, ZMMessageIsExpiredKey]];
 }
 
+//- (NSData *)hashOfContent
```
vytis: Can be deleted probably

mikeger:
@vytis updated

PR #584, opened 2018-10-29
Do not reset company login flag from notification stream

mikeger:
### Issues

SSO user lost the information about it's being an SSO user after receiving any profile update.

### Problem

We use one method to process /self response and any notification stream self update payload.


mikeger:
@zenkins release patch

PR #586, opened 2018-10-30
Data model: fix array vs set and relation cardinality

mikeger:
### Issues

- Changed the relation cardinality from 1-1 to 1-many for the Message.replies-Message.quote relation.
- Fixed type to be Set, no array.
- Moved the replies from ZMTextMessage to ZMClientMessage.


mikeger:
@typfel please re-review

Source/Model/Conversation/ZMConversation+Message.swift
```diff
@@ -56,7 +56,7 @@ extension ZMConversation {
         let message = appendClientMessage(with: clientMessage)!
         message.linkPreviewState = fetchLinkPreview ? .waitingToBeProcessed : .done
         message.quote = replyToMessage
-        replyToMessage?.mutableSetValue(forKey: "replies").add(message)
+        replyToMessage?.mutableSetValue(forKey: ZMMessageRepliesKey).add(message)
```
typfel: This is not necessary since we established the link in line 58

PR #587, opened 2018-10-30
Establish reply/quote relationships when receiving a message

typfel:
## What's new in this PR?

We need to establish a core data relationship when we receive a message containing a quote of another message.

## Notes

- Changed the `ZMText` builder method to take a `ZMMessage` instead of a `id`since we need to compute the hash of the quoted content.
- The case for when a message is edited will come in a separate PR.

### DIFF
--- a/Source/Model/Message/ZMOTRMessage+Replies.swift
+++ b/Source/Model/Message/ZMOTRMessage+Replies.swift
@@ -1,0 +1,41 @@
+//
+// Wire
+// Copyright (C) 2018 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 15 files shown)


Source/Model/Message/ZMOTRMessage+Replies.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+
+
+import Foundation
+
+extension ZMOTRMessage {
+    
+    @objc
+    func updateQuoteRelationships() {
+        // Should be overridden in subclasses
```
alexaubry: Maybe it's worth having an `assert` or `fatalError` here so that we catch when we don't override this method in superclasses?

PR #588, opened 2018-10-31
Edit messages in-place

typfel:
## What's new in this PR?

When editing a message we now update the edit message in-place instead of deleting the original message an inserting a new edited message. This has the benefit that any existing reply relationships will stay intact after a message is edited.

### DIFF
--- a/Source/Model/Message/ConversationMessage+Deletion.swift
+++ b/Source/Model/Message/ConversationMessage+Deletion.swift
@@ -80,9 +80,8 @@ extension ZMClientMessage {
         else {
             return false
         }
-        
-        return genericMessage.hasEdited() ||
-               genericMessage.hasText() && !isEphemeral && (deliveryState == .sent || deliveryState == .delivered)
+                
+        return (genericMessage.hasEdited() || genericMessage.hasText()) && !isEphemeral && (deliveryState == .sent || deliveryState == .delivered)
     }
 }
 
--- a/Source/Model/Message/ConversationMessage+Deletion.swift
+++ b/Source/Model/Message/ConversationMessage+Deletion.swift
@@ -68,30 +68,6 @@ extension ZMMessage {
         return message
     }
     
-    @objc public static func edit(_ message: ZMConversationMessage, newText: String, mentions: [Mention] = [], fetchLinkPreview: Bool = true) -> ZMMessage? {
-        guard let castedMessage = message as? ZMMessage else { return nil }
-        return castedMessage.edit(newText, mentions: mentions, fetchLinkPreview: fetchLinkPreview)
-    }
-        
-    func edit(_ newText: String, mentions: [Mention] = [], fetchLinkPreview: Bool) -> ZMMessage? {
-        guard isEditableMessage else { return nil }
-        guard !isZombieObject, let sender = sender , sender.isSelfUser else { return nil }
-        guard let conversation = conversation, let messageNonce = nonce else { return nil }
-        guard let newMessage = conversation.append(message: ZMMessageEdit.edit(with: ZMText.text(with: newText, mentions: mentions), replacingMessageId: messageNonce), expires: true) else { return nil }
-        
-        newMessage.updatedTimestamp = newMessage.serverTimestamp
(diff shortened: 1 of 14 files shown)


mikeger:
This would allow to observe the message properly, I've seen in my reply view that the message is deleted and no trace of the new message is given in the observation.

typfel:
@mikeger @alexaubry realised I need update the sending edit processing as well. More changes coming.

typfel:
Updated

typfel:
@mikeger @alexaubry please take a look at the changes for local edits

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -68,43 +68,21 @@ extension ZMMessage {
...
-                   (genericMsg.hasEdited() ||
-                       (genericMsg.hasText() && !isEphemeral && (deliveryState == .sent || deliveryState == .delivered)))
+        guard let genericMessage = genericMessage,
+              let sender = sender, sender.isSelfUser
+        else {
+            return false
         }
-        return false
+        
+        return genericMessage.hasEdited() ||
```
alexaubry: Why do we check if `genericMessage.hasEdited()` here?

Source/Model/Message/ZMClientMessage.m
```diff
@@ -229,24 +229,19 @@ - (void)removeMessageClearingSender:(BOOL)clearingSender
...
     }
     [super expire];
 }
 
 - (void)resend
 {
     if (self.genericMessage.hasEdited) {
-        NOT_USED([ZMMessage edit:self newText:self.textMessageData.messageText mentions:self.textMessageData.mentions fetchLinkPreview:YES]); // TODO jacob why can't we just resend the message?
+        // Re-apply the edit since we've restored the orignal nonce when the message expired
+        [self editText:self.textMessageData.messageText mentions:self.textMessageData.mentions fetchLinkPreview:YES];
```
johnxnguyen: Do we want to always fetch the link preview here? Shouldn't we obey the user setting?

Source/Model/Message/ZMClientMessage+Editing.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// - Returns: true if edit was succesfully applied
+    @objc
+    func processMessageEdit(_ messageEdit: ZMMessageEdit, from updateEvent: ZMUpdateEvent) -> Bool {
+        guard let nonce = updateEvent.messageNonce(),
+              let senderUUID = updateEvent.senderUUID(),
+              senderUUID == sender?.remoteIdentifier,
+              messageEdit.hasText()
+        else { return false }
+        
+        add(ZMGenericMessage.message(content: messageEdit.text, nonce: nonce).data())
```
alexaubry: Does that mean that you are inserting a new message for every edit?

Source/Model/Message/ZMClientMessage.m
typfel: Yes seems like we are always fetching the link preview here. This setting is however passed in from the UI when we send a message and we don't store it in SE. We would need to inject this setting into SE to be able to do this properly, that's too big of change to do in this PR though.

Source/Model/Message/ConversationMessage+Deletion.swift
typfel: I only re-factored this code the logic is the same but this is necessary in order to edit an already edited message.

the protobuf structure is like this:

```
genericMessage > text
genericMessage > edited > text
```

An edited message can only contain text at moment which is the reason why we don't check `genericMessage.edited?.hasText()`. We probably should probably always check the delivery state in both cases.

Source/Model/Message/ZMClientMessage+Editing.swift
typfel: This doesn't insert a new message instead it merges the previous protobuf with the new edited one, essentially replacing it.

alexaubry: Ah I see. Thanks for explaining.

PR #590, opened 2018-11-05
ZIOS-10799: Added quote field to DraftMessage

nicorsm:
## What's new in this PR?

Added the `quote` field to the `DraftMessage` class, and also added the supporting class `StorableQuote` and other methods. Updated tests.

### DIFF
--- a/Source/Public/DraftMessage.swift
+++ b/Source/Public/DraftMessage.swift
@@ -26,16 +26,19 @@ import Foundation
     public let text: String
     /// The mentiones contained in the text.
     public let mentions: [Mention]
+    /// The quoted message, if available.
+    public let quote: ZMMessage?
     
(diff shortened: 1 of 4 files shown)


Source/Public/DraftMessage.swift
```diff
@@ -139,11 +163,21 @@ fileprivate extension Mention {
 
 }
 
+fileprivate extension ZMMessage {
+    
+    /// The storable version of the object.
+    var storable: StorableQuote {
+        return StorableQuote(nonce: nonce)
```
alexaubry: Maybe `storable` is too general? If this is only used once, maybe we can just inline it in the initializer.

nicorsm: I agree. I followed the same behavior of mentions, but there it's used together with `map()`.

PR #591, opened 2018-11-06
Feature/direct replies mute logic

johnxnguyen:
## What's new in this PR?

This PR introduces changes for direct replies in the context of muted conversation logic:
- `MutedMessageTypes` options have been renamed to reflect the inclusion of direct replies. The logic here hasn't changed, since replies are treated exactly the same as mentions. I've renamed `nonMentions` to `regular` and `mentions` to `mentionsAndReplies`. The way to understand this is `regular` means messages that are neither a self mention nor a self reply are muted.
- The data model has been updated to include an `internalEstimatedUnreadSelfReplyCount`.
- The unread conversation predicate has been updated to use this new property.
- The unarchiving behaviour has been updated to consider replies to the self user.
- The `ZMConversationListIndicator` has been updated to include a case for unread self replies.


Source/Model/Conversation/ZMConversation+Internal.h
```diff
@@ -73,6 +73,7 @@ extern NSTimeInterval ZMConversationDefaultLastReadTimestampSaveDelay;
 extern NSString *const ZMConversationEstimatedUnreadCountKey;
 
 extern NSString *const ZMConversationInternalEstimatedUnreadSelfMentionCountKey;
+extern NSString *const ZMConversationInternalEstimatedUnreadSelfReplyCountKey;
```
billypchan: I think "Self" is not necessary and just name it "ZMConversationInternalEstimatedUnreadReplyCountKey" ?

mikeger: I think Self is included to underline that only the mentions of self user are counted. So I think it is valid.

johnxnguyen: Also true 👍

Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -45,14 +46,14 @@ public struct MutedMessageTypes: OptionSet {
...
-    public static let all: MutedMessageTypes = [.nonMentions, .mentions]
+    public static let all: MutedMessageTypes = [.regular, .mentionsAndReplies]
     
-    /// Only non-mentions are muted.
-    public static let nonMentions = MutedMessageTypes(rawValue: MutedMessageOptionValue.nonMentions.rawValue)
+    /// Only regular messages (no mentions nor replies) are muted.
+    public static let regular = MutedMessageTypes(rawValue: MutedMessageOptionValue.regular.rawValue)
     
-    /// Only mentions are muted. Only used to check the bits in the bitmask.
+    /// Only mentions and replis are muted. Only used to check the bits in the bitmask.
```
mikeger: Typo: `replis`

PR #594, opened 2018-11-07
Direct replies

typfel:
## What's new in this PR?

Add support for direct replies:

- quote/replies relationship added to Message
- support for quotes in drafts
- support for replies when counting unread messages


typfel:
@zenkins release major

PR #595, opened 2018-11-08
Fix editing message with quote

typfel:
### Issues

We were losing the quote when applying a message edit.

### Causes

According to the spec we do not allow users to edit the quote in a message, only the text can be edited. This means we should copy only text part of an edit but leave any existing quote as they are. We didn't do this but instead copied all the fields from the edit.

### Solutions

Keep the quote from the original message when applying a text edit.


typfel:
@zenkins release patch

PR #597, opened 2018-11-12
Include byte order marker when hashing image asset id

typfel:
### Issues

Images quoted on iOS would show as "not available" on Web and Android.

### Causes

An image hash is based on its asset id which is a string, therefore we must include a byte order marker in the beginning of the data just like we do for text message hashes.

### Solutions

Include byte order marker and update tests.


typfel:
@zenkins release patch

PR #598, opened 2018-11-12
Expose non persisted identifier for message

alexaubry:
### Issues

In UI, we are using the nonce of the message to cache the cell descriptions. This causes a problem when the message is edited, because the nonce changes, which causes race conditions when the edit notification are sent.

### Solutions

- We expose `nonpersistedObjectIdentifer` on the message type protocol to use it in the cache
- We add a test to verify that this identifier does not change on edit


Source/Public/ZMManagedObject.h
```diff
@@ -19,6 +19,8 @@
 
 @import CoreData;
 
+NS_ASSUME_NONNULL_BEGIN
```
typfel: You might need do major release to catch any usage of this in other frameworks.

alexaubry: Ah good point. Should I revert these changes then?

alexaubry: @typfel Reverted. Can you have a look again?

alexaubry:
@zenkins release minor

PR #599, opened 2018-11-12
Read correct message nonce for ephemeral replies

mikeger:
### Issues

Ephemeral reply was not displaying the correct mentioned message.

### Causes

The `genericMessage?.ephemeral?.text?.quote` keypath must be used for the ephemeral messages to get to the text message's quote, since the text message is wrapped inside of the ephemeral container.

### DIFF
--- a/Source/Model/Message/ZMClientMessage+Quotes.swift
+++ b/Source/Model/Message/ZMClientMessage+Quotes.swift
@@ -21,7 +21,16 @@ import Foundation
 extension ZMClientMessage {
     
     override func updateQuoteRelationships() {
-        guard let quote = genericMessage?.text?.quote else { return }
+        let quote: ZMQuote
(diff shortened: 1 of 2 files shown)


Source/Model/Message/ZMClientMessage+Quotes.swift
```diff
@@ -21,7 +21,16 @@ import Foundation
 extension ZMClientMessage {
     
     override func updateQuoteRelationships() {
-        guard let quote = genericMessage?.text?.quote else { return }
```
typfel: we have helper property on `genericMessage` for getting the `ZMText` which takes ephemeral messages into account. 

it's called: `textData`

mikeger: Thanks, fixed!

PR #600, opened 2018-11-12
Fix "objectIdentifier" typo

alexaubry:
objectIdentifer → objectIdentifier


alexaubry:
@zenkins release patch

PR #601, opened 2018-11-13
ZIOS-10799: Moved fetch of quotes to Data Model

nicorsm:
## What's new in this PR?
 
Moved code from https://github.com/wireapp/wire-ios/pull/2899 inside the `ZMConversation` extension available in `DraftMessage`.


Source/Public/DraftMessage.swift
```diff
@@ -132,6 +132,12 @@ fileprivate struct StorableQuote: Codable {
         }
 
     }
+    
+    public func fetchQuoteFromDraftMessage() -> ZMMessage? {
```
typfel: Why is this necessary? Isn't the message is fetched in: 
`func quote(in context: NSManagedObjectContext) -> ZMMessage?`

nicorsm: Conversation is `nil` when fetched from there...

typfel: But that can be fixed.

typfel: You could for example add the conversation parameter to the `draftMessage(in: context)` and then use `ZMMessage.fetch(withNonce: nonce, for: self, in: context)` in `quote(in:)`

nicorsm:
@zenkins release minor

PR #602, opened 2018-11-14
Selectively apply link preview updates

typfel:
### Issues

When applying a link preview update we were merging the whole protobuf which in theory could add quotes. This is not allowed.

### Solutions

When  applying a link preview update we should only pick the link previews from the updated updated protobuf message.


typfel:
@zenkins release patch

PR #603, opened 2018-11-15
Fix for GIFs getting stuck

mikeger:
### Issues

In UI we started to post GIF only after the confirmation controller is dismissed. This caused the image appear while it is still not started to upload. 

We have the code that is requesting the download immediately when it sees the image, the data model was confused and allowed the uploading image to be set to be downloaded. This caused the uploader not to pick up the image.

### Solution

Check in `requestImageDownload` if the image being uploaded now before marking it to be downloaded.


typfel:
@zenkins release patch

PR #604, opened 2018-11-16
Temp: do not assert on multiple delivery receipts

mikeger:
### Issues

In order to un-brick the user's device we need to temporarily disable the assert. We need to run the in-depth investigation to fix it.


mikeger:
@zenkins release patch

PR #605, opened 2018-11-16
Check if link preview exists

mikeger:
### Issues

Guest from ObjC world, `text.linkPreview` can be nil.


mikeger:
@zenkins release patch

PR #606, opened 2018-11-21
Log a warning when a quote is rejected

typfel:
## What's new in this PR?

Log when a quote is rejected because the hash doesn't match. This will make it easier to debug any issues with replies not working.


typfel:
@zenkins release patch

PR #611, opened 2018-11-27
Added self user setting

mikeger:
### Issues

As discussed, I've added the setting to `ZMEditableUser`. I am saving the setting to the context metadata, this seems to be the most appropriate. I could have added the database field to all users, but this seems unnecessary, since we don't know and don't care about the settings of other users.

### DIFF
--- a/Source/Public/ZMEditableUser.h
+++ b/Source/Public/ZMEditableUser.h
@@ -31,5 +31,6 @@
 @property (nonatomic) ZMAccentColor accentColorValue;
 @property (nonatomic, copy, readonly) NSString *emailAddress;
 @property (nonatomic, copy, readonly) NSString *phoneNumber;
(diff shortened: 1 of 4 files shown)


Source/Public/ZMEditableUser.h
```diff
@@ -31,5 +31,6 @@
 @property (nonatomic) ZMAccentColor accentColorValue;
 @property (nonatomic, copy, readonly) NSString *emailAddress;
 @property (nonatomic, copy, readonly) NSString *phoneNumber;
+@property (nonatomic) BOOL enableReadReceipts;
```
typfel: Maybe this should be read only depending on how we implement the server side syncing.

mikeger: @typfel yeah I thought we would sync it with the transcoder, therefore this architecture. We can change it later.

PR #612, opened 2018-11-27
Fix: asset message preview cannot be downloaded

vytis:
### Issues

Some Sync Engine tests started failing after #603 because we have a possibility (not sure if ever used in practice) to download a preview of asset message while it's still being uploaded by the sender. This (sort of) makes sense, when a person sending the message has uploaded a preview and is in the process of uploading the full asset. Then the receiver can look at the preview until an updated message with full asset come

### Solutions

Do not check for transfer state when we are about to download a preview for file message.


vytis:
@zenkins release patch

PR #614, opened 2018-11-28
Allow setting EnableReadReceipts to be synced + rename

mikeger:
### Issues

This sets the modified keys when setting EnableReadReceipts locally.


Source/Model/User/ZMUser+Internal.h
```diff
@@ -84,7 +84,7 @@ extern NSString * __nonnull const EnableReadReceiptsKey;
 @property (nonatomic) ZMAccentColor accentColorValue;
 
 - (void)setHandle:(NSString * __nullable)handle;
-
+- (void)setReadReceiptsEnabled:(BOOL)readReceiptsEnabled remote:(BOOL)isRemote;
```
typfel: In sync engine a parameter like `remote` is is usually named `synchronize`. Would improve readability I think.

mikeger:
@typfel renamed the stuff

PR #617, opened 2018-11-29
Read receipts

typfel:
## What's new in this PR?

- Send read confirmation messages when updating the last read timestamp
- Add stored properties needed for read receipt settings

### ZMEditableUser
* `readReceiptsEnabled` (persistent meta data) (setting for 1:1 read receipts)

### ZMConversation
* `hasReadReceiptsEnabled` - indicate that messages inserted in this conversation needs to be confirmed (only in groups)

### ZMMessage
* `expectsReadConfirmation` - indicate that this message needs to be confirmed when read (only in groups)


typfel:
@zenkins release major

PR #618, opened 2018-11-29
Use both ObjC and Swift protobuf bindings

vytis:
### Issues

We are using ancient Protobuf version to generate Objective-C classes for it. Would be good to migrate to official Protobuf compiler, but the API is vastly different.

@alexaubry has a PR which adds Swift versions of all Protobuf data. It was updated to add back the current ObjC classes. This PR tries to integrate these changes to see if it works.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-protos/pull/27

### Notes

Since Swift Protobuf objects do not have a prefix they do not conflict with ObjC versions of the same thing. They however can conflict with other classes, as in this case LinkPreview.


vytis:
@zenkins release major

PR #619, opened 2018-11-29
Fix nullability of messageUpdateResultFromUpdateEvent

mikeger:
### Issues

Nullability was broken with the recent update (method can return nil).


mikeger:
@zenkins release major

PR #622, opened 2018-12-03
Allow slow sync of user properties

mikeger:
### Issues

This adds one more key that indicates if the user properties must be synced.

It's set to true per default.


mikeger:
@zenkins release minor

PR #623, opened 2018-12-04
Receipt mode system messages

typfel:
## What's new in this PR?

- Split read receipt system message into three types (enabled, disabled, on)
- Insert read receipt system message for new conversation and when the receipt mode is discovered to have changed due to a slow sync.
- Expose methods for appending read receipt system messages to wire-ios-sync-engine.


typfel:
@zenkins release major

PR #624, opened 2018-12-06
Fix broken image when an image upload fails

typfel:
### Issues

If an image upload times out while sending, re-sending it would send a broken image.

### Causes

We do not send placeholders for images while uploading them it therefore also doesn't make sense to send a "failed-to-upload" update for images like we do for files. When an asset message expire we didn't make a distinction between files and images so we were sending a "failed-to-upload" update for images when they expired. Upon successfully sending that update the image was marked as successfully sent since the rest of the image sending code were expecting to only send an update once the full image is uploaded.

### Solutions

Don't add an `NotUploaded` to image protobufs when they expire which will prevent the asset strategy from sending out an update.


typfel:
@zenkins release patch

PR #625, opened 2018-12-07
Fix crash when added to a conversation

typfel:
### Issues

The app would crash when you were added to a conversation with read receipts enabled.

### Causes

When syncing the new conversation we would attempt to add system message using the `serverTimestamp` which is nil, which results in a force unwrap.

### Solutions

Use current date instead.


typfel:
@zenkins release patch

typfel:
@zenkins release patch

PR #626, opened 2018-12-07
Expose needsReadConfirmation in the message protocol

alexaubry:
## What's new in this PR?

We add the `needsReadConfirmation` property to the `ZMConversationMessage` protocol to be able to access it and mock it from UI.


alexaubry:
@zenkins release minor

PR #630, opened 2018-12-10
Keep read expectation when editing a message

typfel:
### Issues

iOS client doesn't send read receipts for edited messages

### Causes

We were losing the `expectsReadConfirmation` flag when applying a message edit.

### Solutions

Make sure to keep the `expectsReadConfirmation` flag.


typfel:
@zenkins release patch

PR #631, opened 2018-12-10
Allow editing read messages

mikeger:
### Issues

It is not possible to commit the edit on the read message.

### Causes

We forgot to update the enum use case where we've been checking if message is possible to edit: before we had sending, failed, sent, delivered states. Now we added the new seen state.

### Solutions

Consider the seen state.


mikeger:
@zenkins release patch

PR #632, opened 2018-12-11
ZIOS-10917: "No read receipts" state description is wrong

nicorsm:
### Issues
Messages sent in a group with read receipts enabled were flagged in the message details view as read receipts were disabled when sending.

### Causes

This is because the `expectsReadConfirmation` was not set when receiving the server timestamp, so there was no visual difference between receipts enabled or not.

### Solutions

I'm setting `expectsReadConfirmation` based on the `hasReadReceiptsEnabled` flag inside of `updateWithPostPayload`.


Source/Model/Message/ZMOTRMessage.m
```diff
@@ -244,4 +244,13 @@ + (MessageUpdateResult *)messageUpdateResultFromUpdateEvent:(ZMUpdateEvent *)upd
     return nil;
 }
 
+-(void)updateWithPostPayload:(NSDictionary *)payload updatedKeys:(NSSet *)updatedKeys {
```
typfel: So you added this method because `expectsReadConfirmation` is only available on `ZMOTRMessage`?

typfel: I think you need to call the super here.

nicorsm: Yes, and the same method inside `ZMMessage` doesn't get called in this case, since the `super` object referenced in `ZMClientMessage` is of `ZMOTRMessage` type

nicorsm: @typfel yes, I need to call `super` because I still need to set the timestamp. Fix is coming.

nicorsm:
@zenkins release patch

PR #633, opened 2018-12-11
ZIOS-10986: Allow liking message with read delivery state

alexaubry:
### Issues

It was not possible to like your own message when it was marked as read.

### Causes

We only allowed liking messages it their delivery state was `delivered` or `sent`.

### Solutions

Add the missing check for `read` state.


alexaubry:
@zenkins release patch

PR #634, opened 2018-12-11
Consider ephemeral location as location message

mikeger:
### Issues

Initial fix: https://github.com/wireapp/wire-ios-data-model/pull/629

Thanks to QA real-life testing found one additional case where the ephemeral message was ignored.


vytis:
A bit unfortunate that we have all this redundant code in different places... Also one thing that maybe in scope for another PR is that at the moment `underlyingMessage` is not cached in any way. Every time we access it we have to merge all protobufs to get the final result of the content. Old style generic message had an intermediate var that would cache this in memory.

mikeger:
@zenkins release patch

PR #635, opened 2018-12-11
Remove implicit save when requesting profile picture

typfel:
### Issues

App would sometimes crash when inserting a new message

### Causes

Root cause for this crash is that we are sometimes saving the context while laying out the subviews in a table view. The save happens because we are implicitly saving the context when we request profile image of a user.

Note that even though we are calling `enqueueDelayedSave()` this doesn't mean that the save will always be delayed. If enough time has passed since the last save this method saves the context immediately.

### Solutions

Remove implicit save and require the UI to wrap the profile image request inside a `enqueueChanges` which will not perform the save immediately.


typfel:
@zenkins release patch

PR #636, opened 2018-12-11
Delete read and delivery receipts when deleting the message

mikeger:
### Issues

We need to remove the confirmations when removing the message. The confirmations must persist when the message is obfuscated.

I added the code to assert the behavior and some tests on top.


mikeger:
@zenkins release patch

PR #637, opened 2018-12-13
ZIOS-11007: Allow deleting read message

alexaubry:
### Issues

Read messages could not be deleted.

### Causes + Solutions

We were not allowing to delete the message if the state is read. We fixed the comparison to include the read state.

### Note

This issue happened a lot during the run, it would be good to look on the issue to identify ways we can catch new delivery state changes in the future.


typfel:
Indeed, perhaps a new property which captures this set of delivery states.

typfel:
@zenkins release patch

PR #638, opened 2018-12-17
Fix not ignoring broken clients when sending as external

typfel:
### Issues

The app would go into a request loop fetching prekeys and attempting to send a message.

### Causes

This could happen in big conversations which contain a client with invalid prekeys. If a client doesn't have valid prekeys we mark the session has broken,  later when we want to send a message to this client we send a 💣 payload and remove the broken session flag so that we will check if the user has uploaded new valid prekeys.

This doesn't work when the message payload is big enough so we decide to re-encrypt the message using an external payload, because the broken session flag is cleared after performing the first encryption. This has the consequence that the app will go into a request loop until the message sending expires.

### Solutions

Clear the broken session flag after potentially re-encrypting the message as an external message.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-images/pull/22

## Notes

This PR also clean us up some encryption method used for creating `ZMOtrAssetMeta` protobufs which was used in the old way of sending asset messages.


typfel:
@zenkins release major

PR #639, opened 2018-12-18
ZIOS-10993: Conversation list secondary line doesn't update with the edited message text

nicorsm:
### Issues

Messages sent inside the conversation list weren't updated after being edited.

### Causes

This is because the conversation list was not observing for changes on `MessageChangeInfo`.

### Solutions

Together with @typfel , we've now updating the list when the generic messages change. This is made possible by creating a new `ConversationChangeInfo.messagesChanged` object after listening for changes with the new keyPath `.genericMessageChanged`.


nicorsm:
@zenkins release patch

PR #640, opened 2018-12-18
Refactor delivery state logic

typfel:
### Issues

In many places we have business logic which verifies that the delivery state is one of the states after a message has been sent. This silently breaks if we add a new delivery state which is easy to miss.

### Solutions

Introduce a new `isSent` property on the message which can replace this logic and is future proof for new delivery states.


typfel:
@zenkins release minor

PR #641, opened 2018-12-19
Add password as a separate property in unregistered user

alexaubry:
### Issues

In the new registration flow, the password will not be provided at the same time as the email. Therefore, we cannot use the previous "credentials" structure as is, because it requires to store the password at the same time as the email.

### Solutions

Add the password as a separate property, and add an helper property to check if the user needs a password.

### Notes

Do not merge this PR yet, as it requires changes in SE and UI.


Source/Model/User/UnregisteredUser.swift
```diff
@@ -71,6 +80,7 @@ extension UnregisteredUser: Equatable {
             && lhs.accentColorValue == rhs.accentColorValue
             && lhs.acceptedTermsOfService == rhs.acceptedTermsOfService
             && lhs.marketingConsent == rhs.marketingConsent
+            && lhs.password == rhs.password
```
vytis: Can we use automatically-derived `Equatable` here?

alexaubry: I think not, because it's a class?

vytis: 😿

alexaubry:
@zenkins release major

PR #642, opened 2018-12-21
Fix observering of read receipts on assets

typfel:
### Issues

List of read receipts didn't live update for assets

### Causes

The `confirmations` property wasn't included in the list observable keys for asset messages.

### Solutions

Make The `confirmations` property observable.


typfel:
@zenkins release patch

PR #643, opened 2018-12-27
Move APS cert name logic to SE

mikeger:
### Issues

See SE PR description. 

### Important changes in the PR.

We had the possibility to keep the application files (database, cryptobox) outside of the shared container.

The necessity of the share container URL is now asserted. Before we had the workaround, but now it's not needed any more, since we have the shared container on simulator and on all sorts of builds. It's better to focus on one approach.

### Dependencies

- [x] https://github.com/wireapp/wire-ios-utilities/pull/70


vytis:
@zenkins release major

PR #648, opened 2019-01-17
Make timers for ephemeral messages optional

vytis:
### Issues

When sending or receiving ephemeral message a timer gets created to fire some time in the future when message should be deleted or obfuscated. The timer was also starting a background task which would make the app stay in the background until the timer expires. We get around 30s of background time so almost all of these background tasks are destined to expire.

### Causes

When a message has destruction date then in various places throughout the code we would try to start the timer. This happens also when app is in the background. It does not make sense to fire these timers while the app is backgrounded.

### Solutions

Made the timers optional and added methods to create these timers instead of implicitly creating them in accessors. This will allow us to avoid starting the timers when they are not present i.e. when app is running in background.

Also removed background task creation in `ZMMessageTimer` because it does not make sense to keep the app running in background because there is some timer active.


vytis:
@zenkins release major

PR #649, opened 2019-01-18
Do not create timers if they already exist

vytis:
### Issues

After #648 we have added API for creating timers explicitly. It was not very robust, because it would re-create timers if they existed.

### Solutions

Create timers only if they don't exist.

## Notes

Tweaked the checks for correct MOC from:

```swift
if(!zm_isUserInterfaceContext) {
   preconditionFailure("must be UI MOC")
}
```

to a simpler

```swift
precondition(zm_isUserInterfaceContext, "must be UI MOC")
```

### DIFF
--- a/Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
+++ b/Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
@@ -33,7 +33,7 @@ public extension NSManagedObjectContext {
     
     @objc public var zm_messageObfuscationTimer : ZMMessageDestructionTimer? {
         precondition(zm_isSyncContext, "MessageObfuscationTimer should be started only on the syncContext")
-
+        
         return userInfo[MessageObfuscationTimerKey] as? ZMMessageDestructionTimer
     }
     
@@ -44,7 +44,7 @@ public extension NSManagedObjectContext {
             log.debug("Obfuscation timer already exists, skipping")
             return 
         }
-        
+
         userInfo[MessageObfuscationTimerKey] = ZMMessageDestructionTimer(managedObjectContext: self)
         log.debug("creating obfuscation timer")
     }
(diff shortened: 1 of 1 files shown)


Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
```diff
@@ -26,33 +26,35 @@ private let log = ZMSLog(tag: "ephemeral")
...
         
         return userInfo[MessageObfuscationTimerKey] as? ZMMessageDestructionTimer
     }
     
     @objc public func zm_createMessageObfuscationTimer() {
-        if !zm_isSyncContext {
-            preconditionFailure("MessageObfuscationTimer should be started only on the syncContext")
+        precondition(zm_isSyncContext, "MessageObfuscationTimer should be started only on the syncContext")
+        
+        guard userInfo[MessageObfuscationTimerKey] == nil else {
```
typfel: This is the real fix?

Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
```diff
@@ -26,33 +26,35 @@ private let log = ZMSLog(tag: "ephemeral")
...
         log.debug("creating obfuscation timer")
     }
     
-    @objc public func zm_createMessageDeletionTimer() {        
-        if !zm_isUserInterfaceContext {
-            preconditionFailure("MessageDeletionTimer should be started only on the uiContext")
+    @objc public func zm_createMessageDeletionTimer() {            
+        precondition(zm_isUserInterfaceContext, "MessageDeletionTimer should be started only on the uiContext")
+        
+        guard userInfo[MessageDeletionTimerKey] == nil else {
```
typfel: And this :)

Source/Model/Message/Ephemeral/ZMMessageDestructionTimer.swift
vytis: Yeah this makes sure to not create it again. If we do that then currently it crashes because timer's `tearDown` was not called before deallocating.

vytis:
@zenkins release patch

PR #651, opened 2019-01-21
Don't insert a text message before all the fields have been populated

typfel:
### Issues

Text message with quotes would in some circumstances be rendered incorrectly.

### Causes

The quote property on a text message is populated after it's been inserted into the conversation. That means if we attempt to render the message immediately we would assume that the quote is invalid.

### Solutions

Populate all fields before inserting the message into the conversation.

## Notes

It would be nice to re-factor the these `appendMessage` methods but I don't wan't to break the API in this PR.


typfel:
@zenkins release minor

PR #656, opened 2019-01-24
Reverting changes from transport

vytis:
## What's new in this PR?

Reverting #653 because it introduced a lot of problems and needs a more careful consideration.


vytis:
A flaky test, all seems fine on transport side

vytis:
@zenkins release major

PR #658, opened 2019-01-28
Avoid unnecessary unfaulting

typfel:
### Issues

We were observing that conversations were being faulted and un-faulted repeatedly during event processing.

### Causes

When fetching the self conversation we will end up calling `fetchObjectWithRemoteIdentifier:inManagedObjectContext:`which would un-fault all conversations when accessing  the `remoteIdentifier` property.

### Solutions

Skip the comparison for faults since we will anyway need to perform a fetch to get the object which is the current fallback if we don't find the object among the registered objects.


typfel:
@zenkins release patch

PR #660, opened 2019-01-29
Data patch for deleting invalid conversations

typfel:
### Issues

As described in  https://github.com/wireapp/wire-ios-sync-engine/pull/953 we've been creating invalid conversations.

### Solutions

Delete all conversations of type `.invalid` in a data patch

## Dependencies

This should be released together or after the fix for not creating invalid conversations.

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-sync-engine/pull/953

### DIFF
--- a/Source/Utilis/PersistedDataPatches+Directory.swift
+++ b/Source/Utilis/PersistedDataPatches+Directory.swift
@@ -29,7 +29,7 @@ extension PersistedDataPatch {
         PersistedDataPatch(version: "78.1.0", block: DuplicatedEntityRemoval.removeDuplicated),
         PersistedDataPatch(version: "81.2.1", block: InvalidClientsRemoval.removeInvalid),
         PersistedDataPatch(version: "103.0.2", block: InvalidGenericMessageDataRemoval.removeInvalid),
-        PersistedDataPatch(version: "145.0.2", block: InvalidConversationRemoval.removeInvalid)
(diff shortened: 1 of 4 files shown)


Source/Utilis/PersistedDataPatches+Directory.swift
```diff
@@ -29,6 +29,7 @@ extension PersistedDataPatch {
         PersistedDataPatch(version: "78.1.0", block: DuplicatedEntityRemoval.removeDuplicated),
         PersistedDataPatch(version: "81.2.1", block: InvalidClientsRemoval.removeInvalid),
         PersistedDataPatch(version: "103.0.2", block: InvalidGenericMessageDataRemoval.removeInvalid),
+        PersistedDataPatch(version: "145.0.2", block: InvalidConversationRemoval.removeInvalid)
     ]
```
johnxnguyen: I'm just curious, how does this work?

typfel: It will extract the version from the data model framework and compare it to what it was running previously and only apply patches associated with versions newer than what it was running before the update.

typfel: I'll change associated version to `145.0.3` so it also applies to internal users which are already running `145.0.2`

typfel:
@zenkins release patch

PR #661, opened 2019-01-29
Don't attempt to avoid refreshing new messages

typfel:
### Issues

We observed that a lot of time was spent in `messagesNotToRefresh` causing the sync thread to stall

### Causes

Since we made `ZMConversdation.messages` an unordered relationship it's not much more expensive to access it since it's backed by a fetch request (fetch results controller).

### Solutions

We decided it's not worth attempting to keep the last three messages from being faulted considering how expensive it is to decide if a message is one of the last three message in a conversation.


typfel:
@zenkins release patch

PR #662, opened 2019-01-29
Revert "Revert "Data patch for deleting invalid conversations (#660)""

typfel:
Restore https://github.com/wireapp/wire-ios-data-model/pull/660


typfel:
@zenkins release patch

PR #663, opened 2019-01-30
Avoid unnecessary fetch request when updating the last read timestamp

typfel:
### Issues

Marking messages as read does a fetch request which is fairly expensive

### Causes

When we mark a message as read we currently search backwards looking for the first unread message and update the last read timestamp using the `serverTimestamp` of that message. This search is done by a fetch request which is fairly expensive.

### Solutions

I believe that this backwards search is a left over from the time when we updated the last read state using event identifiers. Since we nowadays update the last read using a timestamp it should not be any harm in updating the last read state using a newer timestamp.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Timestamps.swift
+++ b/Source/Model/Conversation/ZMConversation+Timestamps.swift
@@ -239,11 +239,8 @@ extension ZMConversation {
     public func markMessagesAsRead(until message: ZMConversationMessage) {
         if let currentTimestamp = lastReadServerTimeStamp,
            let messageTimestamp = message.serverTimestamp,
-           message.sender?.isSelfUser == false,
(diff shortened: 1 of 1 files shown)


typfel:
_Running integration tests:_ ...

typfel:
After further investigation it's still necessary to do a fetch request in one case:
> If the last message in the conversation is a pending/failed to send message we don't want to use its timestamp to update the last read but instead the timestamp of the first unread message above.

This should however be an edge case and in most cases we would update the last read using the given message's timestamp without doing any fetch request.

NOTE: I also unified some code paths for finding the last unread message since from profiling I could not justify having a separate codepath for it.

ping @alexaubry @vytis

typfel:
@zenkins release patch

PR #664, opened 2019-01-30
Fix: destruction timer starts before the file message is sent

billypchan:
### Issues

The ephemeral timer start before the file message is sent. It may expire before the message uploading is done and an error may show in this case.

### Causes

The timer starts and expires before the asset is uploaded

### Solutions

if the message's deliveryState is pending, update destructionDate and do not start destruction.
Updated the tests: set message.delivered = true to let the destruction timer starts.


typfel:
This PR fixes the issue but we should understand the root cause of the issue. When we looked at this code previously we found: https://github.com/wireapp/wire-ios-data-model/blob/d016710096baedad999d2bd20dd64f752c5c6efb/Source/Model/Message/ZMAssetClientMessage%2BEphemeral.swift#L76-L82

It looks like that code is there to solve the issue we are having here. We debugged it and saw that `self.senderClientID` was `nil`  and the if statement was therefore never executed. `senderClientID` is actually never set for messages that you send. That might be something we should fix but I'd argue not in this PR.

The `uploadState` is only modified by sender so we can achieve the behaviour by replacing the if statement with something like:

```
if isSelfUser, !(uploadState == .uploadingFullAsset || uploadState == .done) {
      return false
}
```

typfel:
This will become obsolete when https://github.com/wireapp/wire-ios-data-model/pull/685 is merged

PR #666, opened 2019-02-01
Avoid sorting participants in conversation

vytis:
### Issues

When profiling the app one of the problematic points was using `ZMConversation.activeParticipants` because it sorts them alphabetically. After looking at use cases it seems that most do not need it to be sorted at all.

### Solutions

Added a property `ZMConversation.sortedActiveParticipants` and changes `activeParticipants` to be simple set. Also adding generics annotations saved some casting in call sites.


vytis:
@zenkins release major

PR #667, opened 2019-02-01
Removed unused code for fetching profile pictures

vytis:
## What's new in this PR?

After more thorough inspection it seems that there is no need to nil the identifiers when fetching profile picture. Posting notification is enough to trigger download.


vytis:
@zenkins release patch

PR #669, opened 2019-02-05
Improve conversation status performance

typfel:
## What's new in this PR?

- Introduce `estimatedHasMessages` as a faster way to determine if the conversation has any content than inspecting the `allMessages` relationship.

- Allow `.newConversation` system messages to affect the unread count

- `unreadMessages` now only return messages which return true to `shouldGenerateUnreadCount()`

These changes are done to prepare the conversation status line to only display a secondary line for unread messages. Previously we were also showing an status line for the last message in the conversation if it was a system message.

## Notes

- Removed `unreadTimeStamps` since it's no longer used


typfel:
@zenkins release major

PR #670, opened 2019-02-05
Update quote relationship before updating unread count

typfel:
### Issues

A direct reply would not always have the correct status icon which indicate that's it's a reply to your message.

### Causes

`internalEstimatedUnreadSelfReplyCount` is updated before we establish the quote relationship for a newly inserted message, this means we will count the direct reply as a normal unread message.

### Solutions

We can't call `updateQuoteRelationships` before `updateWithUpdateEvent:forConversation:` since the conversation relationship is established there, which is a prerequisite for `updateQuoteRelationships`. 

To break this circular dependency we move. it inside `updateWithUpdateEvent:forConversation:`


Source/Model/Message/ZMMessage+Quotes.swift
```diff
@@ -0,0 +1,28 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension ZMMessage {
+    
+    @objc
+    func updateQuoteRelationships() {
+        assertionFailure("Subclasses should override this method")
```
typfel: Should verify that we don't call this for system messages now.

vytis:
Is there anything left to do here?

typfel:
@vytis no, I just forgot about this PR

typfel:
@zenkins release patch

PR #671, opened 2019-02-05
Fix downloading images when they are not in cache

vytis:
### Issues

After restoring from backup the images would not load.

### Causes

When requesting image download we were checking that transfer state is not "uploading", but it is always set to uploading by default. This was introduced by #603 and later touched in #612 

### Solutions

Check upload state instead.


vytis:
@zenkins release patch

PR #672, opened 2019-02-07
Add isAccountDeleted field to the user entity

typfel:
## What's new in this PR?

Add`isAccountDeleted` field to the `User` entity. This field will be used to identify users which have deleted their accounts.

## Todo

- [x] Unit test for `removeFromAllConversations()`
- [x] Data model migration test


typfel:
@zenkins release major

PR #673, opened 2019-02-08
Add isAccountDeleted field

typfel:
## What's new in this PR?

Extracted only the data model changes from https://github.com/wireapp/wire-ios-data-model/pull/672


typfel:
@zenkins release minor

PR #674, opened 2019-02-08
Add methods for handling user deletion

typfel:
## What's new in this PR?

Business logic extracted from https://github.com/wireapp/wire-ios-data-model/pull/672


typfel:
@zenkins release major

PR #675, opened 2019-02-11
ZIOS-11203: Added support for new "managed_by" field

nicorsm:
## What's new in this PR?

Added support for the new `managed_by` field.


billypchan:
@nicorsm Ref to https://github.com/wireapp/wire-ios/pull/3184, we should branch a feature branch from release 3.27 branch and then merge to it instead of develop branch.

nicorsm:
@billypchan sure! I'll create the `feature/readonly-profile` (and fix the tests, btw)

alexaubry:
Will this require changes from the sync engine, or break the API on the UI?

nicorsm:
@alexaubry I don't think so, for now, since it's just a value we need to check to enable/disable editing.

WireDataModel.xcodeproj/project.pbxproj
```diff
@@ -45,6 +45,11 @@
 		1689FD462194A63E00A656E2 /* ZMClientMessageTests+Editing.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1689FD452194A63E00A656E2 /* ZMClientMessageTests+Editing.swift */; };
 		16AD86BA1F75426C00E4C797 /* NSManagedObjectContext+NotificationContext.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16AD86B91F75426C00E4C797 /* NSManagedObjectContext+NotificationContext.swift */; };
 		16C391E2214BD438003AB3AD /* MentionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16C391E1214BD437003AB3AD /* MentionTests.swift */; };
+<<<<<<< HEAD
+=======
```
billypchan: @nicorsm merge is in progress.

Source/Model/User/ZMUser+Internal.h
```diff
@@ -50,6 +50,8 @@ extern NSString * __nonnull const ReadReceiptsEnabledKey;
 @property (nullable, nonatomic) NSData *imageMediumData;
 @property (nullable, nonatomic) NSData *imageSmallProfileData;
 
+@property (nullable, nonatomic) NSString *managedBy;
```
typfel: Shouldn't this property be annotated as a `managed` property somewhere?

typfel: I guess `@dynamic managedBy;` since it's Objective-C

Source/Model/User/ZMUser.m
```diff
@@ -563,6 +569,11 @@ - (void)updateWithTransportData:(NSDictionary *)transportData authoritative:(BOO
         self.name = name;
     }
     
+    NSString *managedBy = [transportData optionalStringForKey:@"managed_by"];
```
typfel: A test in `ZMUserTests` checking that we parse this field would be nice.

Source/Model/User/ZMUser.m
```diff
@@ -380,6 +383,7 @@ @implementation ZMUser (Internal)
 @dynamic showingUserAdded;
 @dynamic showingUserRemoved;
 @dynamic createdTeams;
+@dynamic managedBy;
```
nicorsm: @typfel added @dynamic here.

PR #676, opened 2019-02-11
Log which protobuf messages are being processed

typfel:
## What's new in this PR?

Log the protobuf messages (sanitized) we are processing.


typfel:
@zenkins release patch

PR #677, opened 2019-02-12
Old conversations generate a badge count

typfel:
### Issues

It's been observed that internal users get an unread badge count without seemingly having any unread conversations. 

### Causes

We recently started including `.newConversation` system messages in the unread count if the conversation was started by someone else than the self user. This had the side effect of marking all conversations which were slow synced on login as unread. The badge count wouldn't update immediately since we only re-calculate the unread count when something unfaults the conversation.

### Solutions

Distinguish between conversations which were slow synced and conversations which were created from an update event. One way of doing this is by looking at the `lastServerTimestamp` which doesn't exist for slow synced conversations and therefore is set to the UNIX reference. timestamp.


typfel:
@zenkins release patch

PR #679, opened 2019-02-13
Converted public managedBy property to managedByWire

nicorsm:
## What's new in this PR?

Converted public `managedBy` property (string, could be either "wire" or "scim") exposed from Data Model to `managedByWire` (boolean, true if equals to "wire").


Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -1325,6 +1324,11 @@ - (void)testThatCallingRefreshDataMarksItAsToDownload {
     }];
 }
 
+
+-(NSString*)managedByString:(ZMUser*)user {
```
typfel: In the Objective-C its common to add a space between the type and the `*`

```suggestion
-(NSString *)managedByString:(ZMUser *)user {
```

Source/Public/ZMUser.h
```diff
@@ -73,6 +73,8 @@ extern NSString * _Nonnull const ZMPersistedClientIdKey;
 @property (nonatomic, readonly, nullable) NSString *imageSmallProfileIdentifier;
 @property (nonatomic, readonly, nullable) NSString *imageMediumIdentifier;
 
+@property (nonatomic, readonly) BOOL managedByWire;
```
typfel: Maybe we should also add `managedByWire` to the `UserType` protocol.

nicorsm:
@zenkins release major

PR #680, opened 2019-02-14
Create .newConversation system messages with a timestamp

typfel:
### Issues

`.newConversation` system messages can now generate unread dots, this created unwanted unread  conversation for people who upgrade.

### Causes

We previously tried to solve this by introducing `hasBeenModifiedSinceSlowSync` but that solution didn't cover all the edge cases,  a conversation can be modified by for example someone being added or removed causing the same issue to re-appear.

### Solutions

`.newConversation` system message will from now on have a correct `serverTimestamp` associated with them since that timestamp will be visible when you read the message, to enable this  we need to re-factor `appendNewConversationSystemMessageIfNeeded` to take a timestamp parameter.

NOTE: The solution for the unwanted unread conversations is done in sync engine.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Internal.h
+++ b/Source/Model/Conversation/ZMConversation+Internal.h
@@ -106,10 +106,6 @@ NS_ASSUME_NONNULL_END
 
 @property (nonatomic) BOOL internalIsArchived;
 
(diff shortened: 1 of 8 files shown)


typfel:
@zenkins release major

PR #681, opened 2019-02-15
ZIOS-11202: Add property for SCIM extended metadata

alexaubry:
### Issues

The backend will start providing an array of key/value pairs as it rolls out support for SCIM extended profiles. We need to save this field to the user model in Core Data.

### Solutions

- Add the `extendedMetadata` requirement on the `UserType` protocol to allow testing before BE is ready.
- Add a `extendedMetadata` field on the Core Data model for `ZMUser`. It has the "transformable" type, which means Core Data will encode and decode the NSArray directly.
- Add appropriate migration tests.


alexaubry:
@zenkins release major

PR #682, opened 2019-02-18
Bump transport

vytis:
Bump transport after https://github.com/wireapp/wire-ios-transport/pull/130


vytis:
@zenkins release major

PR #683, opened 2019-02-19
Add required SCIM observation and permissions properties

alexaubry:
### Issues

In the UI project, we needed two properties on the user:
- detecting changes in extended metadata
- detecting if two users can see each other's private company information

### Solutions

Add the required properties in the observing system and the `UserType` protocol. To detect if two users have access to their mutual company info, we compare their team ID. If they match, we allow them to see the info.


alexaubry:
@zenkins release major

alexaubry:
@zenkins release major

PR #685, opened 2019-02-21
Introduce abstraction for processing asset uploads

typfel:
## What's new in this PR?

Simplify asset uploading. 

We no longer want to send intermediate update for an asset message before the the final asset has been uploaded to the backend. This allows us to remove some state which currently main for asset messages.

- `uploadState` is removed, was used to keep track of which assets have been uploaded.
- `transferState` is now keeping track of the upload state and the state of incoming assets (like before).
- `downloadState` is a new transient property which keep track of the download state, this was previously the job of the `transferState`.

Two new protocols have been added `Asset` and `AssetMessage`. This is the abstraction on which the asset request strategies will work.

## Todos

- [x] Delete intermediate data after message is sent
- [x] Migration of legacy transfer field
- [x] Update tests
- [x] Improve test coverage


Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -347,12 +292,237 @@ extension ZMAssetClientMessage {
...
-    case uploadingFailed = 4
+    case preprocessing
+    case encrypting // TODO jacob remove this step?
+    case uploading
 }
 
+struct CacheAsset: Asset {
+    
+    enum AssetType {
+        case image, file, thumbnail
```
vytis: What about video? This will need some preprocessing to get the preview

Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -347,12 +292,237 @@ extension ZMAssetClientMessage {
     case thumbnail = 3
 }
 
-@objc public enum AssetUploadState: Int16 {
+@objc public enum AssetDownloadState: Int16 {
+    case unavailable = 0
+    case downloaded
+    case downloading
+}
```
vytis: Maybe there should also be "not downloaded" state?

Source/Model/Message/ZMAssetClientMessage.swift
typfel: The video will have two assets `.file ` and `.thumbnail`.

Source/Model/Message/ZMAssetClientMessage.swift
typfel: I was thinking about it but it's not really useful in practice since this state is only valid when the `transferState` == `.uploaded`. We can add it for completeness but it won't be used in the UI.

vytis: Not sure exactly how it all will work, but I assume we will then always try to download asset when a corresponding notification fires? States will then transition `unavailable -> downloading -> unavailable` each time we display a message with this broken asset?

typfel: I guess `unavailable` is a confusing name, are more appropriate name would be `availableRemotely` or something.

typfel: Renamed `.unavailable` to `.remote`

Source/Model/Message/ZMAssetClientMessage+GenericMessage.swift
```diff
@@ -164,59 +157,21 @@ extension ZMAssetClientMessage {
     
     public override func update(with message: ZMGenericMessage, updateEvent: ZMUpdateEvent, initialUpdate: Bool) {
         self.add(message)
+        self.version = 3 // We only support receiving V3 assets
```
vytis: A nitpick but the comment is a bit confusing - we support downloading v2 assets, but not uploading so there is no way for us to receive an event that v2 was uploaded.

Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -241,49 +223,26 @@ import Foundation
...
-            self.uploadState = .uploadingPlaceholder
+        if transferState != .uploaded {
+            transferState = .uploading
         }
         
-        self.transferState = .uploading
         self.progress = 0
-        self.removeNotUploaded()
-        
+        setLocallyModifiedKeys(Set(arrayLiteral: #keyPath(ZMAssetClientMessage.transferState))) // TODO jacob hacky
```
vytis: Todo left

Source/Model/Message/ZMAssetClientMessage+GenericMessage.swift
typfel: Backend doesn't accept V2 assets anymore therefore I assumed we we also won't get any incoming V2 assets.

Source/Model/Message/ZMAssetClientMessage.swift
typfel: Mm I think I'll just remove this TODO since we don't have better way of singling that we need to sync this message again.

Source/Model/Message/ZMAssetClientMessage+GenericMessage.swift
typfel: V2 assets were also not sent as regular messages instead they were posted to /conversation/<id>/assets which fanned them out to everyone in the conversation. Since this endpoint is gone it should be pretty hard to send a V2 asset today.

vytis: Makes sense, just reading through the comment made me think about downloading an asset. No changes in code needed, just thought the comment could be a bit more clearly worded.

typfel: Ah ok, yes sure I'll improve the comment.

Source/Utilis/TransferStateMigration.swift
```diff
@@ -28,39 +28,37 @@ struct TransferStateMigration {
...
-            switch self {
-            case .uploading: return .uploading
-            case .uploaded, .downloading, .downloaded, .failedDownloaded, .unvailable: return .uploaded
-            case .failedUpload: return .uploadingFailed
-            case .cancelledUpload: return .uploadingCancelled
-            }
-        }
+        case unavailable
+                
+        static var migrationMappings: [(AssetTransferState, [LegacyTransferState])] = [
```
vytis: A bit of a nitpick, but maybe reversing the tuple would be clearer, so that we have them in format `(FROM, TO)`.

Source/Utilis/TransferStateMigration.swift
```diff
@@ -28,39 +28,37 @@ struct TransferStateMigration {
...
         
-        let assetMessages = moc.fetchOrAssert(request: fetchRequest)
+        let transferStateKey = "transferState"
         
-        for assetMessage in assetMessages {
-            guard let rawLegacyTransferState = assetMessage.primitiveValue(forKey: transferStateKey) as? Int,
-                  let legacyTransferState = LegacyTransferState(rawValue: rawLegacyTransferState)
-            else { continue }
+        for (newValue, legacyValues) in LegacyTransferState.migrationMappings {
+            let batchUpdateRequest = NSBatchUpdateRequest(entityName: ZMAssetClientMessage.entityName())
```
vytis: Not sure how to address it, but it's not obvious that there is no overlap between values - e.g. if we migrate from `3` to `1` and then on the next batch update maybe migrate `1` to `4`. Basically it should boil down to avoid migrating same object twice. Maybe we already ensure this by having the right values in `migrationMappings` and it would be enough just to mention it.

typfel: Yeah, good point. I'll add a comment and the order is of course also important.

typfel:
@zenkins release major

PR #686, opened 2019-02-22
Fix: read receipts does not work when sending location

vytis:
### Issues

When sending location message read receipts are ignored and for sender the message is not marked as read.

### Causes

After #618 we have introduced Swift protobufs and methods for reading read receipt flags were not implemented in the code that deals with location.

### Solutions

Add read receipt flags to Swift protobuf protocol.


vytis:
@zenkins release patch

PR #687, opened 2019-02-25
Fields for rich user profile

vytis:
### Issues

We want to display extra info in the user profile that we receive from backend.

### Solutions

Added a flag `needsRichProfileUpdate` that when set to `true` should trigger download of `richProfile` from the server.

## Still need to do

- [x] Migration test


Source/Model/User/ZMUser+RichProfile.swift
```diff
@@ -0,0 +1,60 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum ZMUserKeys {
+    public static let RichProfile = "richProfile"
```
typfel: This is also defined in ZMUser.m:L96 maybe it makes sense to expose through `ZMUser.h`

Source/Model/User/ZMUser.m
```diff
@@ -877,6 +880,7 @@ @implementation ZMUser (Editable)
 @dynamic readReceiptsEnabled;
 @dynamic needsPropertiesUpdate;
 @dynamic readReceiptsEnabledChangedRemotely;
+@dynamic needsRichProfileUpdate;
```
typfel: Shouldn't `needsRichProfileUpdate` also be added to the list of ignored keys?

Source/Model/User/UserType.swift
```diff
@@ -100,7 +100,8 @@ public protocol UserType: NSObjectProtocol {
     var readReceiptsEnabled: Bool { get }
     
     /// The extended metadata for this user, provided by SCIM.
-    var extendedMetadata: [[String: String]]? { get }
+    var richProfile: [UserRichProfileField] { get }
+    var needsRichProfileUpdate: Bool { get set }
```
alexaubry: Can we add a documentation comment for this flag?

Source/Model/User/ZMUser+Predicates.swift
```diff
@@ -0,0 +1,96 @@
...
+fileprivate extension String {
+    
+    func normalizedAndTrimmed() -> String {
+        guard let normalized = self.normalizedForSearch() as String? else { return "" }
+        return normalized.trimmingCharacters(in: .whitespaces)
+    }
+    
+    func strippingLeadingAtSign() -> String {
+        guard hasPrefix("@") else { return self }
+        var copy = self
```
alexaubry: I think `return String(self.dropFirst())` would be more idiomatic and straightforward here.

vytis:
@zenkins release major

PR #689, opened 2019-03-01
Added nil condition to managedByWire

nicorsm:
### Issues

Users are not able to edit profile when the `managedBy` property is `nil`.

### Solutions

Added this case to the`managedByWire` property.

### DIFF
--- a/Tests/Source/Model/User/ZMUserTests.m
+++ b/Tests/Source/Model/User/ZMUserTests.m
@@ -838,6 +838,23 @@ static NSString *const ImageSmallProfileDataKey = @"imageSmallProfileData";
     XCTAssertNil(user.emailAddress);
 }
 
+- (void)testThatItSetsManagedByAsWire
+{
+    // given
+    NSUUID *uuid = [NSUUID createUUID];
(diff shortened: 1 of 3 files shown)


Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -838,21 +838,21 @@ - (void)testThatItSetsEmailToNilIfItIsMissing
     XCTAssertNil(user.emailAddress);
 }
 
-- (void)testThatItSetsManagedByAsWire
+- (void)testThatNilManagedByIsConsideredAsManagedByWire
```
typfel: Shouldn't we add a new test instead of editing the existing one?

nicorsm: Ouch, I overwritten that. I'll restore it.

alexaubry:
@nicorsm You also need to update `ZMSearchUser` to change the default value:

```swift
var managedByWire: Bool {
    return user?.managedByWire != false
```

alexaubry:
@zenkins release patch

PR #690, opened 2019-03-05
Expose activeConversations on the UserType

typfel:
## What's new in this PR?

Expose `activeConversations` on the `UserType`. We need this property to in order to calculate the set of "active" users.


typfel:
@zenkins release major

PR #691, opened 2019-03-06
Add support for additional fields in the Member entity

typfel:
## What's new in this PR?

- Add `createdBy` and `createdAt` fields to `Member` entity
- Parse `created_by` and `created_at` in the team member payload
- Remove query normalisation from search predicate (it's normalised higher up in SE now)

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-utilities/pull/73


typfel:
@zenkins release minor

PR #693, opened 2019-03-07
Crash when sending ephemeral gif

typfel:
### Issues

App crashes after sending ephemeral gif

### Causes

App crashes due to division by zero when computing image constraints because the image size is reported as zero.

The image size is zero because we don't do image preprocessing for GIFs so were relying on the value of `preprocessedSize`, which gets set to zero when the image gets obfuscated.

### Solutions

- Add the image size into the protobuf from the start, we don't need  the`preprocessedSize` property anymore but we have to keep for now to be compatible with older images.
- Make constraint code resilient to zero size images.

### DIFF
--- a/Source/Model/Message/V2Asset.swift
+++ b/Source/Model/Message/V2Asset.swift
@@ -100,14 +100,15 @@ extension String {
     }
 
     public var originalSize: CGSize {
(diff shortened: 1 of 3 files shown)


typfel:
@zenkins release patch

PR #695, opened 2019-03-13
Remove V2 profile images

typfel:
## What's new in this PR?

Remove methods and properties associated with V2 profile images. In data model this involves methods for setting & reading the profile images and  for parsing user-update events.

Fields that was used to track V2 profile images:
* imageSmallProfileIdentifier
* imageMediumIdentifier
* localImageSmallProfileIdentifier
* localImageMediumIdentifier
* imageMediumData;
* imageSmallProfileData;

In addition to removing things I've made two logical changes:
* The self user profile pictures are now stored in the same on disk cache as regular users, instead of persisting them in core data.
* Observing when profile image data change is now done by firing a `nonCoreDataChange` instead of mixture of core data observing and custom observer code in the notification dispatcher.


typfel:
@zenkins release major

PR #696, opened 2019-03-15
Observe asset identifier in addition to the image data

typfel:
### Issues

Your own profile image wouldn't update immediately after changing it.

### Causes

 We observing when the image data changed on disk but at the time when it changes the `previewProfileAssetIdentifier` and `completeProfileAssetIdentifier` were not yet updated on the UI context. This resulted in that we fetched the old profile image from the cache instead of the new one.

### Solutions

Also observe changes to `previewProfileAssetIdentifier` and `completeProfileAssetIdentifier`.


typfel:
@zenkins release patch

PR #697, opened 2019-03-18
Fix: Change search query length limit

billypchan:
## What's new in this PR?

Change search query length limit form 2-character to 1-character.

### DIFF
--- a/Source/Model/Message/TextSearchQuery.swift
+++ b/Source/Model/Message/TextSearchQuery.swift
@@ -183,7 +183,7 @@ public class TextSearchQuery: NSObject {
         self.conversation = conversation
         self.conversationRemoteIdentifier = conversation.remoteIdentifier!
         self.originalQuery = query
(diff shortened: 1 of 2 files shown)


Source/Model/Message/TextSearchQuery.swift
```diff
@@ -179,7 +183,7 @@ private let zmLog = ZMSLog(tag: "text search")
         self.conversation = conversation
         self.conversationRemoteIdentifier = conversation.remoteIdentifier!
         self.originalQuery = query
-        self.queryStrings = query.normalizedForSearch().components(separatedBy: .whitespacesAndNewlines).filter { $0.count > 1 }
+        self.queryStrings = query.normalizedForSearch().components(separatedBy: .whitespacesAndNewlines).filter { TextSearchQuery.isValid(query: $0) }
```
alexaubry: I think it's possible to make it shorter here:

```suggestion
        self.queryStrings = query.normalizedForSearch().components(separatedBy: .whitespacesAndNewlines).filter(TextSearchQuery.isValid)
```

billypchan:
@zenkins release minor

PR #698, opened 2019-03-18
Undo - Change search query length limit

billypchan:
## What's new in this PR?

After discussion with QA team, we decided to keep the current query length limit to align to other platforms at this moment. The changes in https://github.com/wireapp/wire-ios-data-model/pull/697 is keep but set the limit to 1 character.


billypchan:
@zenkins release patch

PR #699, opened 2019-03-19
ZIOS-11281: Delivery receipt - send a cumulative one once the sync is done

nicorsm:
## What's new in this PR?

- Added sending of a single delivery receipt, like it's already happening with reading receipts (it uses the same method)
- Added test
- Fixed a repetition in a method signature


vytis:
I guess it's worth making a playground build with this change to check if other platforms support batching of delivery receipts also.

Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -277,6 +277,7 @@ extension ZMConversation {
     func savePendingLastRead() {
         guard let timestamp = pendingLastReadServerTimestamp else { return }
         confirmUnreadMessagesAsRead(until: timestamp)
+        confirmMessagesAsDelivered()
```
typfel: There's a difference between `read` and `delivery` receipts: delivery receipts should be sent in the background without any action of the receiver. `savePendingLastRead()` is called as a response to opening and reading a conversation.

Source/Model/Conversation/ZMConversation+Confirmations.swift
```diff
@@ -44,6 +44,34 @@ extension ZMConversation {
         return confirmationMessages
     }
     
+    public static func confirmDeliveredMessages(_ messages: [UUID], in conversations: [UUID], with managedObjectContext: NSManagedObjectContext) -> [ZMMessage] {
+        
+        var confirmationMessages: [ZMMessage] = []
+        for conversationID in conversations {
+            guard let convo = ZMConversation(remoteID: conversationID, createIfNeeded: false, in: managedObjectContext),
```
typfel: We could use `ZMManagedObject.fetchObjectsWithRemoteIdentifiers()` to fetch all conversations with one fetch request. Look at how it's done for users in `ZMuser.usersWithRemoteIDs()`

We could the same thing for messages.

nicorsm:
@zenkins release minor

PR #700, opened 2019-03-20
Batch fetch users when updating conversation members

typfel:
### Issues

During profiling it's visible that `updateMembersWithPayload:` take up a significant time if the number of participants is big.

### Causes

Inside `updateMembersWithPayload:` the line which takes the most time is `[ZMUser userWithRemoteID:userId createIfNeeded:YES inContext:self.managedObjectContext]` since it's executed for every participant added to the conversation.

### Solutions

Fetch all participants in one batch.

## Notes

- I replaced `NSOrderedSet` with a plain `NSSet` since we don't rely on the order.


typfel:
@zenkins release major

PR #701, opened 2019-03-20
ZIOS-11296: Persist link attachments

alexaubry:
### Issues

Due to compatibility issues between clients, we need to treat YouTube and SoundCloud links as a separate type of links, since they cannot be sent as link previews.

Their parsing was previously done in the UI and not persisted; which means we needed to recompute them for every rendering of the text message cells.

In addition to this performance bottleneck, we used the YouTube and SoundCloud API to manually fetch the details of the media, which this could have been achieved using the OpenGraph data provided by these websites at faster speeds and lower costs.

### Solutions

In https://github.com/wireapp/wire-ios-link-preview/pull/41, we add types and helpers to detect and represent these link attachments. With this PR, we add a field on the message model in CoreData to store the `ZMLinkAttachment`, and another field to trigger an update in request-strategy.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-link-preview/pull/41

## Notes

Using this branch, we will need to:

- update Request Strategy to create the context observer that will kick off the attachment detection
- update UI to opt-out of the old system (for now only for YouTube links)


vytis:
It's a bit sad to work with NSCoding in Swift when we have Codable. There is a way to mark the field as JSON data and then decode it on the fly. Not sure what are the performance implications of it though, but the code is much nicer 😎 https://github.com/wireapp/wire-ios-data-model/pull/687/files#diff-8c9281077ff32a7c0ecde0b280887dccR48

WireDataModel.xcodeproj/project.pbxproj
```diff
@@ -2343,7 +2353,6 @@
 					F9C9A7F21CAED9510039E10C = {
 						CreatedOnToolsVersion = 7.2;
 						LastSwiftMigration = 0800;
-						ProvisioningStyle = Manual;
```
typfel: Intentional change?

Source/Model/Message/ZMClientMessage+Editing.swift
```diff
@@ -41,6 +41,7 @@ extension ZMClientMessage {
         self.nonce = nonce
         self.updatedTimestamp = updateEvent.timeStamp()
         self.reactions.removeAll()
+        self.needsLinkAttachmentsUpdate = true
```
typfel: Shouldn't we also remove any previously fetched link attachments? Wouldn't we otherwise potentially display links which no longer exists in the message

typfel: Related to this, isn't there also separate code path for when the self user edits a message?

alexaubry: Good call, I will set the property to `nil` here then. I also do it in `editText(_ text: String, mentions: [Mention], fetchLinkPreview: Bool)` for the self user edits

alexaubry:
@typfel I updated the PR to nil out the attachments on edit, and to remove the code to trigger the update as we discussed (will do in UI when message is scrolled into view, to respect the user setting).

WireDataModel.xcodeproj/project.pbxproj
alexaubry: No, I turned signing on and back off to test on device. Should I change it back?

typfel: Ah ok, I guess it doesn't matter.

alexaubry:
@zenkins release major

PR #704, opened 2019-03-28
ZIOS-7347: Do not play audio and video formats we don't understand

alexaubry:
### Issues

When rendering video or audio messages, we would offer the user to play video or audio files with the player, even if the format cannot be played by the system (such as MIDI or MPEG).

### Causes

When checking if an asset is a video or audio, we didn't check if the file type was actually playable by the system, only if it conformed to the audio or video class.

### Solutions

Before flagging a file as audio or video, we query the supported MIME times from `AVFoundation`. We also refactor the type detection by moving away from the previous implementation which was a String extension. We now have an enum that reports the supported types of the message.

This means that when the user receives a video file that is not supported by AVFoundation, it will show up as a file transfer in the conversation list and the user will be able to download it and open it in another app.

We also integrate the new UTI helper, which simplifies some parts of the code.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-utilities/pull/75


alexaubry:
@zenkins release major

PR #705, opened 2019-03-29
Unarchive a conversation when you are added to it

typfel:
### Issues

If you leave a group the conversation will get archived, if you later get added back the conversation will stay in archive.

### Causes

it's unclear what was causing the unarchiving previously. (*checking*)

### Solutions

Unarchive the conversation if it isn't muted.


typfel:
@zenkins release patch

PR #706, opened 2019-04-01
ZIOS-11330: Add blacklist property to link metadata

alexaubry:
### Issues

We do not want to display YouTube/SoundCloud/Giphy link previews in the UI, or make them behave like link previews (i.g. with removing the trailing link from the text).

### Solutions

In the helper property that returns the first valid link preview, return `nil` if the preview is blacklisted and should not be displayed to the user.


alexaubry:
@zenkins release minor

PR #707, opened 2019-04-01
Support disabling notification dispatcher

typfel:
## What's new in this PR?

* Support disabling the notification dispatcher by adding the `isDisabled` property.
* Add support for observing when conversation lists reload, this is necessary because we now re-load the conversation lists when the app enter foreground inside the `ConversationListObserverCenter`.  It used be done from the UI but is now the responsibility of the notification dispatcher.


typfel:
@zenkins release major

PR #708, opened 2019-04-05
Remove recentMessages property

typfel:
## What's new in this PR?

Remove `recentMessages` property. This property was backed by a fetch results controller which is quite expensive to run for each conversation.


typfel:
@zenkins release major

PR #711, opened 2019-04-10
Add mutedMessageTypes property for the self user

typfel:
## What's new in this PR?

This PR adds a the `mutedMessageTypes`  property on `ZMuser`, which returns a option set of muted message types based on your availability status.

The union of this property and the existing `mutedMessageTypes`  property on `ZMConversation` is exposed as `mutedMessageTypesIncludingAvailability`.

The purpose of these new properties is to also take your availability status into account when determining if:
- a notification show be displayed or not
- ~~if we show a badge count~~
- if calls are displayed


typfel:
@zenkins release minor

PR #712, opened 2019-04-11
ZIOS-11329: Improve detection and observing of guest/services presence

alexaubry:
### Issues

The system to detect if the conversation contains guests or services was located in UI, and was not observable. 

While it is fine in most cases, there are some cases where a service can be added remotely and we get the participants changed notification before we fetch that the added user is a service, we will not update the UI to reflect that it's a service once the fetch is done (e.g. we won't show the "services are present banner").

### Solutions

This PR contains two solutions to this problem:

1. We move the logic from `GuestBarState` that's defined in UI into data model. We create the `ZMConversationExternalParticipantsState` enum, and the `externalParticipantsState` property on the conversation. We add the matrix and the tests for this property.

2. We make this property observable as part of `ConversationChangeInfo`. This involved:
    - defining the keys that affect `isServiceUser` on `ZMUser`
    - defining the keys that affect the `externalParticipantsState`
    - updating the side effects store for the user to use the changed and affected keys when notifying conversation of participants changes instead of just the changed keys. We do that so that we can observe `isServiceUser` directly (ex: if `serviceIdentifierChanges` changes, then `isServiceUser` is affected, and we should allow the conversation to observe `lastSyncedActiveParticipants.isServiceUser` directly)

### DIFF
--- a/Source/Model/Conversation/ZMConversation+ExternalParticipant.swift
+++ b/Source/Model/Conversation/ZMConversation+ExternalParticipant.swift
@@ -60,29 +60,22 @@ extension ZMConversation {
 
         // Calculate the external participants state
         let canDisplayGuests = selfUser.team != nil
-
-        var areServicesPresent: Bool = false
-        var areGuestsPresent: Bool = false
+        var state = ExternalParticipantsState()
 
         for user in otherUsers {
             if user.isServiceUser {
-                areServicesPresent = true
+                state.insert(.visibleServices)
             } else if canDisplayGuests && user.isGuest(in: self) {
-                areGuestsPresent = true
+                state.insert(.visibleGuests)
             }
 
             // Early exit to avoid going through all users if we can avoid it
-            if areServicesPresent && (areGuestsPresent || !canDisplayGuests) {
+            if state.contains(.visibleServices) && (state.contains(.visibleGuests) || !canDisplayGuests) {
                 break
             }
         }
 
-        switch (areGuestsPresent, areServicesPresent) {
-        case (false, false): return []
-        case (true, false): return [.visibleGuests]
-        case (false, true): return [.visibleServices]
-        case (true, true): return [.visibleGuests, .visibleServices]
-        }
+        return state
     }
 
     /// Returns whether an services are present, regardless of the display rules.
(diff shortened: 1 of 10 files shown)


Source/Model/Conversation/ZMConversation+ExternalParticipant.swift
```diff
@@ -0,0 +1,105 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+/**
+ * Represents the possible state of external participants in a conversation.
+ */
+
+@objc public enum ZMConversationExternalParticipantsState: Int, CustomStringConvertible {
```
typfel: So this will be accessed by objective-C? Otherwise it could be a good fit for an OptionSet.

alexaubry: I thought about it, but yeah we will need to access it from Objective-C in UI (the conversation view controller). I can always try to refactor that part of UI to Swift though, if this is a better solution. 

Another thing I considered when not making it an option set, is that we have individual copy for each case, and in the future I'm not sure how we would deal with adding more cases. Do you think it's a problem?

typfel: Sure this doesn't scale very well if we add more cases, but we can worry about that when we add more cases. Don't know how big such a re-factoring would be in UI, if it's not small I wouldn't do it now.

alexaubry: Agreed. It's rather straightforward, so I'll do it.

alexaubry: @typfel Updated to use the option set.

Source/Model/Conversation/ZMConversation+ExternalParticipant.swift
```diff
@@ -0,0 +1,98 @@
...
+                areGuestsPresent = true
+            }
+
+            // Early exit to avoid going through all users if we can avoid it
+            if areServicesPresent && (areGuestsPresent || !canDisplayGuests) {
+                break
+            }
+        }
+
+        switch (areGuestsPresent, areServicesPresent) {
```
typfel: Now that this is an option set this could be simplified:

```
var externalParticipants = ExternalParticipantsState()
if areGuestsPresent {
    externalParticipants.insert(.visibleGuests)
}
if areServicesPresent {
    externalParticipants.insert(.visibleServices)
}
```

alexaubry:
@zenkins release minor

PR #713, opened 2019-04-12
Add flag for availability behaviour change

typfel:
## What's new in this PR?

We will alter the behaviour of the availability feature, which we want to inform the user about when they upgrade. The solution is to add the `needsToNotifyAvailabilityBehaviourChange` flag on `ZMUser` which we set to true in a data model patch if the user is `away` or `busy`.

## Todo

- [x] Write tests


typfel:
@zenkins release minor

PR #714, opened 2019-04-16
Make Availability conform to CaseIterable

alexaubry:
## What's new in this PR?

This allows us to remove a helper in UI that implements the behavior of CaseIterable to access all the cases of the enum.


alexaubry:
@zenkins release minor

PR #715, opened 2019-04-24
Simplify return value when creating OTR message from update event

typfel:
## What's new in this PR?

We now create delivery receipts at end of the event processing instead of when inserting the messages, therefore we no longer access any of the properties on `MessageUpdateResult`. This means we can simply the API and just return the created OTR message directly.


typfel:
@zenkins release major

PR #717, opened 2019-04-26
remove libPhoneNumber

billypchan:
## What's new in this PR?

remove unused libPhoneNumber


vytis:
We need to make sure that it is not used anywhere in UI or other frameworks that include data model. I remember at some point it was used for phone registration, but maybe we have refactored it already.

billypchan:
@vytis I will make a PR in UI project and test it.

billypchan:
@zenkins release major

PR #719, opened 2019-05-02
Enqueue a delayed save when requesting to refetch link attachments

typfel:
### Issues

Performing many `refetchLinkAttachmentsIfNeeded` is expensive.

### Solutions

Take over the responsibility of saving the context and do it with a delayed save.

### DIFF
--- a/Source/Model/Message/ConversationMessage+Attachments.swift
+++ b/Source/Model/Message/ConversationMessage+Attachments.swift
@@ -18,21 +18,18 @@
 
 import Foundation
 
-extension ZMConversationMessage {
+extension ZMClientMessage {
 
     /**
      * Requests to refetch the link attachments of messages received prior to
      * the persistent link attachments update.
      */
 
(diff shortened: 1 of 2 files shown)


Source/Model/Message/ConversationMessage+Attachments.swift
```diff
@@ -18,21 +18,18 @@
...
-        }
-
-        if linkAttachments == nil {
-            needsLinkAttachmentsUpdate = true
-        }
+    public override func refetchLinkAttachmentsIfNeeded() {
+        guard !needsLinkAttachmentsUpdate &&  linkAttachments == nil && textMessageData != nil else { return }
+        
+        needsLinkAttachmentsUpdate = true
+        managedObjectContext?.enqueueDelayedSave()
```
typfel: Noticed that this can be lead to save within a save if the UI calls this method in response to save.

PR #722, opened 2019-05-14
ZIOS-11422: Make the conversation not secure when ignoring the new client alert

alexaubry:
### Issues

When a conversation becomes unverified after discovering a client when sending a message, we show an alert to the user asking them whether we want to send the message anyway, see the device or cancel.

If they cancel, the conversation remains degraded, and the next time the user sends a message, we show the alert again.

### Causes

We do not decrease the security of the conversation when the user cancels the alert.

### Solutions

Mark the conversation as not secure when cancelling the alert, so that we don't ask the user again. That way, we align the behavior with calls and the upcoming legal hold feature.


typfel:
I'd run the integration tests in `wire-ios-sync-engine` against this PR.

alexaubry:
@zenkins release major

PR #723, opened 2019-05-16
Convert deviceClass and type into typed enums

typfel:
## What's new in this PR?

- Turn `UserClient.type` and `UserClient.deviceClass` from `String` into extensible typed enums.
- Annotate `UserClientType` with nullability
- Add `update(with payload: [String: Any])` in order to move responsibility for parsing user client payload from request strategy to the data model.

### DIFF
--- a/Source/Model/UserClient/UserClient.swift
+++ b/Source/Model/UserClient/UserClient.swift
@@ -342,14 +342,11 @@ public extension UserClient {
         client.type = DeviceType(rawValue: type)
         client.activationAddress = activationAddress
         client.model = model
+        client.deviceClass = deviceClass.map({ DeviceClass(rawValue: $0) })
         client.activationDate = activationDate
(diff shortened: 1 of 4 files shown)


Source/Model/UserClient/UserClient.swift
```diff
@@ -315,15 +339,18 @@ public extension UserClient {
...
         client.activationAddress = activationAddress
         client.model = model
-        client.deviceClass = deviceClass
         client.activationDate = activationDate
         client.activationLocationLatitude = latitude
         client.activationLocationLongitude = longitude
         client.remoteIdentifier = id
         
+        if let deviceClass = deviceClass {
+            client.deviceClass = DeviceClass(rawValue: deviceClass)
```
alexaubry: Maybe you can shorten this to `client.deviceClass = deviceClass.map(DeviceClass.init)`, if you want to keep it on one line.

PR #724, opened 2019-05-16
Implement detection of legal hold in users and conversation

alexaubry:
### Issues

We want to be able to detect the legal hold user clients when they are removed and added to a conversation, and notify the UI of this change.

### Solutions

- Update the security level handling to be able to take legal hold into account
- Add properties on data model and change observers
- Add the mechanism to insert system messages

### Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-data-model/pull/723

### To-Do

- [x] Finish system message implementation
- [x] Add test cases
- [x] Wait for detection and alert conditions meeting decisions


Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -284,49 +384,69 @@ extension ZMConversation {
...
+                            sender: ZMUser.selfUser(in: self.managedObjectContext!),
+                            users: users,
+                            clients: clients,
+                            timestamp: timestampAfterLastMessage())
     }
     
-    fileprivate enum DiscoveryCause {
-        case message(ZMOTRMessage?)
+    fileprivate enum SecurityChangeCause {
+        case message(ZMOTRMessage?, Set<UserClient>)
```
typfel: Would it make sense to call this `addedClients(_ clients: Set<UserClient>, _ message: ZMOTRMessage?)`

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -16,58 +16,146 @@
...
+        applySecurityChanges(cause: .ignoredClients(clients))
+    }
+
+    /// Applies the security changes for the set of users.
+    private func applySecurityChanges(cause: SecurityChangeCause) {
+        // Check legal hold changes
+        updateLegalHoldState(cause: cause)
+
+        // Check verification
+        if securityLevel == .secure {
```
typfel: Could we extract this `if else` into a `updateSecurityLevel(cause: cause)`?

typfel:
Looks good,

I'm thinking about the degradation scenario for legal hold. How would we know that we need to display the alert in the UI? 

`isUnderLegalHold == true AND !messagesThatCausedSecurityLevelDegradationByConversation.isEmpty`

But `messagesThatCausedSecurityLevelDegradationByConversation` is transient so that knowledge would be gone on the launch. That isn't so bad but it also means that we would silently start sending read and delivery receipts in the conversation.

But maybe that's the OK? I'll bring that up in the next meeting.

alexaubry:
@typfel Would having something similar to the `secureWithIgnored` security level for legal hold (e.g. `legalHoldWithIgnoredClients`) to save the legal hold status instead of just a boolean?

typfel:
@alexaubry yes, but let's wait for the answer to my question about this in group.

alexaubry:
@typfel Updated. Maybe you could have a look at the new tests too?

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -16,58 +16,184 @@
...
+
+    @NSManaged private var primitiveLegalHoldStatus: Int
+
+    /// Whether the conversation is under legal hold.
+    @objc public internal(set) var legalHoldStatus: ZMConversationLegalHoldStatus {
+        get {
+            willAccessValue(forKey: #keyPath(legalHoldStatus))
+            defer { didAccessValue(forKey: #keyPath(legalHoldStatus)) }
+
+            if let status = (primitiveValue(forKey: #keyPath(legalHoldStatus)) as? Int).flatMap(ZMConversationLegalHoldStatus.init) {
```
typfel: You could use the `primitiveLegalHoldStatus` here?

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -16,58 +16,184 @@
...
+                return status
+            } else {
+                legalHoldStatus = .disabled
+                return .disabled
+            }
+        }
+        set {
+            willChangeValue(forKey: #keyPath(legalHoldStatus))
+            setPrimitiveValue(newValue.rawValue, forKey: #keyPath(legalHoldStatus))
```
typfel: Also here

WireDataModel.xcodeproj/project.pbxproj
```diff
@@ -128,7 +129,9 @@
 		54FB03AA1E41F204000E13DC /* PersistedDataPatches+Directory.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54FB03A81E41F1B6000E13DC /* PersistedDataPatches+Directory.swift */; };
 		54FB03AD1E41F6C2000E13DC /* PersistedDataPatchesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54FB03AC1E41F6C2000E13DC /* PersistedDataPatchesTests.swift */; };
 		54FB03AF1E41FC86000E13DC /* NSManagedObjectContext+Patches.swift in Sources */ = {isa = PBXBuildFile; fileRef = 54FB03AE1E41FC86000E13DC /* NSManagedObjectContext+Patches.swift */; };
-		5E0FB2012051493700FD9867 /* Set+Filter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5E0FB2002051493700FD9867 /* Set+Filter.swift */; };
+		5E055B2C228C736500E91314 /* (null) in Resources */ = {isa = PBXBuildFile; };
+		5E055B2E228C7F1C00E91314 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
billypchan: Some garbage added by Xcode.

Source/Model/UserClient/UserClient.swift
```diff
@@ -58,7 +58,7 @@ private let zmLog = ZMSLog(tag: "UserClient")
         }
     }
     
-    @NSManaged public var type: DeviceType
+    @NSManaged public var type: DeviceType?
```
typfel: The DeviceType is not optional,  It's `"permanent"` by default.

Source/Model/UserClient/UserClientType.swift
```diff
@@ -0,0 +1,50 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+protocol UserClientType {
```
typfel: `fetchFingerprintOrPrekeys` got lost in translation.

Source/Public/ZMMessage.h
```diff
@@ -62,30 +62,30 @@
...
-    ZMSystemMessageTypeMessageDeletedForEveryone,
-    ZMSystemMessageTypePerformedCall,
-    ZMSystemMessageTypeTeamMemberLeave,
-    ZMSystemMessageTypeMessageTimerUpdate,
-    ZMSystemMessageTypeReadReceiptsEnabled,
-    ZMSystemMessageTypeReadReceiptsDisabled,
-    ZMSystemMessageTypeReadReceiptsOn,
-    ZMSystemMessageTypeLegalHoldEnabled,
-    ZMSystemMessageTypeLegalHoldDisabled
+    ZMSystemMessageTypeParticipantsAdded = 1,
```
typfel: Why explicitly state the value?

Tests/Source/Model/Conversation/ZMConversationTests+Legalhold.swift
```diff
@@ -0,0 +1,341 @@
+//
+//  ZMConversationTests+Legalhold.swift
```
typfel: Wrong header

Source/Model/Message/ZMClientMessage+TextMessageData.swift
```diff
@@ -48,12 +48,12 @@ extension ZMClientMessage: ZMTextMessageData {
         
         let mentions = Array(protoBuffers.compactMap({ Mention($0, context: managedObjectContext) }).prefix(500))
         var mentionRanges = IndexSet()
-        let messageRange = messageText.startIndex.encodedOffset...messageText.endIndex.encodedOffset
-        
+        let messageRange = NSRange(messageText.startIndex ..< messageText.endIndex, in: messageText)
```
billypchan: It is doing the same thing here(https://github.com/wireapp/wire-ios/pull/3396/files), we can move the logic to Utilities.

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
alexaubry: Now it works, I needed to use a NSNumber as the primitive type, because it expects an object as a primitive value. That's why setting 0 works: it was treated as a nil object and therefore didn't crash.

Source/Public/ZMMessage.h
alexaubry: I added the explicit values because it was a pain to debug: lldb only prints the raw value or `__C.ZMSystemMessageType` and I needed to count which one was which. I can remove the explicit values, but I'm wondering if there's a better way to solve the debugging issue that I'm missing.

Source/Model/UserClient/UserClient.swift
alexaubry: It's optional in the protocol though, that's why it prompted me to fix the issue. I also thought that the device type is only for the self clients?

Source/Public/ZMMessage.h
typfel: Right, like would it be possible to implement `CustomDebugStringConvertible` for the enum?

Source/Model/UserClient/UserClient.swift
typfel: Right I wrote an incorrect comment. I think you translated the `UserClientType` from an outdated version because I later changed it to `nonnull`

All non self clients will always be of permanent type since that's the default in the data model.

Source/Public/ZMMessage.h
typfel: My concern with writing out the value explicitly is that it increases the chance for typos by duplicating or skipping a number.

alexaubry: It's not really an issue, I think I will revert this and think about it in another PR.

alexaubry:
@typfel Is it good to go in the feature branch now?

PR #726, opened 2019-05-22
[Feature] Legal Hold

alexaubry:
## What's new in this PR?

Implementing the legal hold feature.

### Solutions

We reuse the code paths used by conversation verification when the devices change. We add new system messages and refactor the logic to include both verification and legal hold.

We also update the data model to persist the legal hold state to the conversation.


alexaubry:
@zenkins release major

PR #729, opened 2019-05-24
Fix system message issues

alexaubry:
### Issues

After merging the logic for degrading and legal hold, tests started failing in sync engine.

### Causes

Some logic got lost in translation:

- We did not insert the list of users in the system message when a client was added.
- It was possible to move to the `notSecure` verification level from the `secureWithIgnored` state.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -153,7 +153,7 @@ extension ZMConversation {
                 securityLevel = .secure
                 appendNewIsSecureSystemMessage(cause: cause)
                 notifyOnUI(name: ZMConversation.isVerifiedNotificationName)
-            } else {
+            } else if securityLevel != .secureWithIgnored {
                 if case .ignoredClients = cause {
                     securityLevel = .secureWithIgnored
                 } else {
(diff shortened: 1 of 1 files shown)


Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -153,7 +153,7 @@ extension ZMConversation {
                 securityLevel = .secure
                 appendNewIsSecureSystemMessage(cause: cause)
                 notifyOnUI(name: ZMConversation.isVerifiedNotificationName)
-            } else {
+            } else if securityLevel != .secureWithIgnored {
                 if case .ignoredClients = cause {
                     securityLevel = .secureWithIgnored
                 } else {
```
billypchan: It seems more make sence to add the else case after else if and move `securityLevel = .notSecure` to there?

alexaubry: It doesn't make much sense to set the security level to ignored if it's already ignored, so I figured I would put it higher up to check both cases in the if that comes after.

alexaubry: @billypchan I refactored this method to be more clear, as we discussed. Could you review again?

typfel: But then `securityLevel = .notSecure` becomes redundant because inside this block the securityLevel is always `.notSecure`

PR #730, opened 2019-06-03
Add support for syncing user clients on demand

typfel:
## What's new in this PR?

We now need to fetch the user client details before we calculate the new security level in the conversations where the client is present.

In order to support this we will add the `needsToBeUpdatedFromBackend` flag which will indicate the the client needs to be fetch from the backend. A newly discovered client will always have the `needsToBeUpdatedFromBackend` flag set to `true`.

Since it's important to know which message discovered a new client in order to insert system messages at the right position we also a `dicoveredByMessage` property on the `UserClient` entity.


typfel:
@zenkins release major

PR #731, opened 2019-06-03
ZIOS-11451: Set legal hold status when sending message

alexaubry:
### Issues

We want to send a hint to other users about legal hold presence when sending a message.

### Solutions

- Add a mechanism on the message to allow setting the flag
- Update the content protobuf objects to allow setting the flag


alexaubry:
@zenkins release major

PR #732, opened 2019-06-05
fix: GIF form file app can not finish uploading

billypchan:
### Issues

When uploading a GIF file from file picker, the upload does not finish.

### Causes

The thumbnail is not cached and the app tries to get it but never succeeded.

### Solutions

Allow preprocessing for all kind of image and thumbnail. (Then GIF thumbnail cache can be generated.)


Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -345,9 +345,9 @@ struct CacheAsset: Asset {
         switch type {
         case .file:
             return false
-        case .image, .thumbnail:
-            return owner.genericMessage?.assetData?.original.mimeType
-                .flatMap(UTType.init(mimeType:))?.isGIF == false
```
typfel: I think the reason why we excluded gifs from preprocessing is because during resizing the gif will be turned it into a jpeg.

typfel: But if this was only issue with thumbnails I guess we can do the check only for `.image`.

billypchan: @typfel I tested sending both animated/one-frame GIF and the thumbnails seem resized correctly.

typfel:
I think we should add a test for this.

Source/Model/Message/ZMAssetClientMessage.swift
typfel: Aha :) I see know that the downscaling is only applied to GIFs larger than 5 MBs

https://github.com/wireapp/wire-ios-images/blob/f6f57f7ccbce6b03fb24cbb86266cb5ec3be886d/Sources/Image%20Processing/ZMImageDownsampleOperation.m#L31

billypchan:
@typfel updated exist testing about GIF asset

billypchan:
@zenkins release minor

PR #733, opened 2019-06-06
Fix legal hold system message position

typfel:
### Issues

When we discover a legal hold client we are currently inserting the system message after the message which discovered it.

### Solutions

Use the message associated with the discovered client to insert the system message before the which discovered it.


typfel:
@zenkins release patch

PR #734, opened 2019-06-07
ZIOS-11452: Process legal hold hint flag when receiving message

alexaubry:
### Issues

We added support for legal hold flags in messages in #731. We now need to process this flag when receiving a message to update the legal hold status in the conversation.

### Solutions

- Update the message content protocol to conditionally return the status, only if it's set in the protobuf
- Add a method on the conversation to update the status based on the hint and current status
- Update the OTR message downstream creation method to call the method mentioned above before inserting the message

### DIFF
--- a/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -73,6 +73,14 @@ extension ZMConversation {
 
     // MARK: - Events
 
+    /// Should be called when a message is received.
+    /// If the legal hold status hint inside the received message is different than the local status,
+    /// we update the local version to match the remote one.
+    @objc(updateSecurityLevelIfNeededAfterReceiving:)
+    public func updateSecurityLevelIfNeededAfterReceiving(message: ZMOTRMessage) {
+        updateLegalHoldIfNeededWithHint(from: message)
+    }
+
     /// Should be called when client is trusted.
     /// If the conversation became trusted, it will trigger UI notification and add system message for all devices verified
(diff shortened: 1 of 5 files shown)


Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -291,6 +318,14 @@ extension ZMConversation {
                             timestamp: timestamp ?? timestampAfterLastMessage())
     }
 
+    private func appendLegalHoldEnabledSystemMessageForConversation(afterReceiving message: ZMGenericMessage, at timestamp: Date) {
```
typfel: `message` parameter is unused

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -299,6 +334,14 @@ extension ZMConversation {
                             timestamp: timestampAfterLastMessage())
     }
 
+    private func appendLegalHoldDisabledSystemMessageForConversation(afterReceiving message: ZMGenericMessage, at timestamp: Date) {
```
typfel: `message` parameter is unused

alexaubry: Will remove, thanks

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
alexaubry: Removed 👍

alexaubry:
@zenkins release major

PR #735, opened 2019-06-11
Fix: new clients are reported as new users joining the conversation

typfel:
### Issues

New clients are reported as new users joining the conversation.

### Causes

The new clients system message included a non-empty set of  `addedUsers` which causes the UI report it as new users joining the conversation.

### Solutions

Don't set `addedUsers` when discovering a new clients.


typfel:
@zenkins release patch

PR #736, opened 2019-06-11
Add hasLegalHoldRequest attribute to the User entity

billypchan:
## What's new in this PR?

Add hasLegalHoldRequest attribute to the User entity.


billypchan:
@zenkins release major

PR #737, opened 2019-06-12
ZIOS-11474: Persist legal hold request

alexaubry:
### Issues

We want to persist the legal hold request after receiving it from the backend through an update event.

### Solutions

- Add a `legalHoldRequest` property on the User entity in Core Data, as a JSON Data blob
- Create the Codable structure that will contain the data from the request payload
- Add logic to handle receiving and accepting requests
- Add `legalHoldStatus` computed property on the user

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-transport/pull/137


Source/Model/UserClient/UserClient.swift
```diff
@@ -361,6 +362,7 @@ public extension UserClient {
         
         let selfUser = ZMUser.selfUser(in: context)
         client.user = client.user ?? selfUser
+        selfUser.needsToAcknowledgeLegalHoldStatus = client.isLegalHoldDevice
```
typfel: This will reset `needsToAcknowledgeLegalHoldStatus` to false if another self client is created after the legal hold client.

Source/Model/UserClient/UserClient.swift
```diff
@@ -247,6 +247,7 @@ private let zmLog = ZMSLog(tag: "UserClient")
 
         // increase securityLevel of affected conversations
         if let previousUser = user {
+            previousUser.needsToAcknowledgeLegalHoldStatus = isLegalHoldDevice && previousUser.isSelfUser
```
typfel: This similar issue here.

Source/Model/UserClient/UserClient.swift
alexaubry: Ah I didn't think of this, thanks

alexaubry:
@typfel I fixed the reset issue and added tests for it here: https://github.com/wireapp/wire-ios-data-model/pull/737/commits/da98173a42d7233a88090103806a72945f8d6f47

alexaubry:
@zenkins release major

PR #738, opened 2019-06-13
Fix .newClient system message not having any users set

typfel:
### Issues

Tests were failing because `.newClient` system message didn't have any users.

### Causes

This broke in https://github.com/wireapp/wire-ios-data-model/pull/735

### Solutions

Restore previous behaviour but keep the distinction between `users` and `addedUsers`


typfel:
@zenkins release patch

PR #739, opened 2019-06-13
Add needsToVerifyLegalHold flag on conversation entity

typfel:
## What's new in this PR?

Add `needsToVerifyLegalHold` flag on the conversation entity. This flag gets set to `true` when the legal hold status get updated by the `legalHoldStatus` flag on a message. 

## Notes

I will merge this after https://github.com/wireapp/wire-ios-data-model/pull/737 and then also update data model and tests.


vytis:
Migration tests fail because current version is not added to the array. Maybe we can think how to automate this test even more since we always increment the version by 1.

typfel:
> Migration tests fail because current version is not added to the array. Maybe we can think how to automate this test even more since we always increment the version by 1.

Right, but as I will merge this PR after https://github.com/wireapp/wire-ios-data-model/pull/737 I would need to redo the data model tests anyway.

typfel:
@zenkins release minor

PR #741, opened 2019-06-17
API change for safe logging

vytis:
### Issues

`PrivateStringConvertible` no longer exists and we need to conform to `SafeForLoggingStringConvertible` instead


vytis:
@zenkins release major

PR #742, opened 2019-06-17
ZIOS-11483: Add logic to create legal hold client from request

alexaubry:
### Issues

We want to create a legal hold client when the user accepts a legal hold.

### Solutions

Create an entity when the user accepts the request, and establish a session from the data in the request.


Source/Model/User/ZMUser+LegalHoldRequest.swift
```diff
@@ -18,6 +18,8 @@
 
 import Foundation
 
+private let log = ZMSLog(tag: "legal-hold")
```
typfel: How about using the already existing `UserClient` tag?

Tests/Source/Model/User/ZMUserLegalHoldTests.swift
```diff
@@ -29,36 +29,34 @@ class ZMUserLegalHoldTests: ModelObjectsTests {
         XCTAssertEqual(selfUser.legalHoldStatus, .disabled)
     }
 
-    func testThatLegalHoldStatusIsPending_AfterReceivingRequest() {
+    func testThatLegalHoldStatusIsEnabled_AfterReceivingRequest() {
```
typfel: Why did you rename this test method? Previous title seems to match what we are checking now.

alexaubry: Right, I got confused, will revert.

alexaubry:
@zenkins release major

PR #743, opened 2019-06-21
Fix legal hold disclosure issues

alexaubry:
## What's new in this PR?

In this PR, we fix bugs related to the disclosure of legal hold to the user.

### ZIOS-11493: Wrong position of legal hold system message when creating a group

When a user under legal hold created a group conversation, the system message indicating that legal hold is enabled was inserted before the "new conversation" system messages. The expected behavior is the opposite.

#### Causes

The method `+[ZMConversation insertGroupConversationIntoManagedObjectContext:withParticipants:]` previously inserted the users in the conversation before inserting the new conversation message. However, the call to insert users (`-[ZMConversation internalParticipants:]`) also called the method to degrade the conversation if needed.

While this didn't have any effect on verification, because the conversation isn't verified when created, it actually caused the legal hold to be enabled, because the conversation is legal-hold-disabled by default and the degradation enables it.

#### Solutions

We refactor this method to insert the users after inserting the new conversation message, thus avoiding the degradation before inserting the message. To support this, we update the `appendNewConversationSystemMessage` method to pass in the users that were initially added.

Another change we did is updating the `internalAddParticipants` to take an array instead of a set. The reason behind this is that previously, the insert conversation method, which takes an array of users, iterated over the users and called `internalAddParticipants` with a set of one user at a time, to retain the order inside the mutableLastServerSyncedActiveParticipants, which is an ordered set. To make it more efficient, we now pass an array to this method and convert it to an ordered set.

### ZIOS-11498: Legal hold status isn't displayed in any conversation on the device where it was accepted

There was a bug where the device where legal hold is accepted doesn't display any indication that legal hold is enabled in the conversations that existed prior to the approval.

#### Causes

Legal hold is enabled for conversations in a call for `decreaseSecurityLevelIfNeededAfterDiscoveringClients`, which is called when a client is added remotely, not locally, as it's the case for the legal hold client on the client that accepts the hold.

#### Solutions

Call the degradation flow via `decreaseSecurityLevelIfNeededAfterDiscoveringClients` in the block called by the request strategy when the user accepts the legal hold request.

### DIFF
--- a/Source/Model/Conversation/ZMConversation.m
+++ b/Source/Model/Conversation/ZMConversation.m
@@ -783,8 +783,10 @@ const NSUInteger ZMConversationMaxTextMessageLength = ZMConversationMaxEncodedTe
     conversation.creator = selfUser;
     conversation.team = team;
 
-    [conversation internalAddParticipants:[NSSet setWithObject:participant]];
-    [conversation appendNewConversationSystemMessageAtTimestamp:[NSDate date]];
+    NSSet<ZMUser *> *participants = [NSSet setWithObject:participant];
+
+    [conversation appendNewConversationSystemMessageAtTimestamp:[NSDate date] users:participants];
+    [conversation internalAddParticipants:@[participant]];
 
     // We need to check if we should add a 'secure' system message in case all participants are trusted
     [conversation increaseSecurityLevelIfNeededAfterTrustingClients:participant.clients];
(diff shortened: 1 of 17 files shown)


Source/Model/Conversation/ZMConversation.m
```diff
@@ -873,24 +875,29 @@ + (nullable instancetype)insertGroupConversationIntoManagedObjectContext:(nonnul
...
     NSMutableSet *allClients = [NSMutableSet set];
     for (ZMUser *user in conversation.activeParticipants) {
         [allClients unionSet:user.clients];
     }
-    
-    [conversation appendNewConversationSystemMessageAtTimestamp:[NSDate date]];
-    
-    // We need to check if we should add a 'secure' system message in case all participants are trusted
+
     [conversation increaseSecurityLevelIfNeededAfterTrustingClients:allClients];
```
typfel: Can't this line be deleted now since this is already done inside `internalAddParticipants`?

Source/Model/Conversation/ZMConversation.m
```diff
@@ -783,8 +783,10 @@ + (instancetype)fetchOrCreateTeamConversationInManagedObjectContext:(NSManagedOb
...
 
-    [conversation internalAddParticipants:[NSSet setWithObject:participant]];
-    [conversation appendNewConversationSystemMessageAtTimestamp:[NSDate date]];
+    NSSet<ZMUser *> *participants = [NSSet setWithObject:participant];
+
+    [conversation appendNewConversationSystemMessageAtTimestamp:[NSDate date] users:participants];
+    [conversation internalAddParticipants:@[participant]];
 
     // We need to check if we should add a 'secure' system message in case all participants are trusted
     [conversation increaseSecurityLevelIfNeededAfterTrustingClients:participant.clients];
```
typfel: Same thing here, this can be deleted?

alexaubry: Resolving this, as we discussed in person.

alexaubry:
@zenkins release major

PR #744, opened 2019-06-24
ZIOS-11499: Fix redundant legal hold update messages

alexaubry:
### Issues

When receiving a message with a legal hold hint, we would insert too many system messages:

```
<legal hold activated>
<message content>
<legal hold deactivated>
<legal hold activated>
```

### Causes

The reason behind this is that after we finish the request to the BE to verify the legal hold hint, we'd call `updateSecurityLevelIfNeededAfterFetchingClients`, which performs an unconditional legal hold state update before we actually fetch the legal hold client. This caused a legal hold disabled system message to be inserted, and a hold-enabled message to be inserted later after we actually fetch the device. 

### Solutions

- Create a `ZMConversationRemoteClientChangeSet` option set to pass the type of changes that occurred to `updateSecurityLevelIfNeededAfterFetchingClients`
- If there are no changes, that means that the local state is correct, and that therefore we can update the legal hold state.
- If there are changes, we need to wait for them to process (eg client metadata to be fetched), which will in turn cause a legal hold status update through existing flows (adding a client, removing a client, etc).

We also stop assuming which type of event enable or disable legal hold, and we always check for both scenarios in `updateLegalHoldState` (reduces complexity and possible errors).

### DIFF
--- a/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -34,6 +34,28 @@ import WireCryptobox
     }
 }
 
+/**
+ * Represents a set of client changes in a conversation.
+ */
+
+public struct ZMConversationRemoteClientChangeSet: OptionSet {
+    public let rawValue: Int
+
+    public init(rawValue: Int) {
+        self.rawValue = rawValue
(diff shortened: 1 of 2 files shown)


alexaubry:
@zenkins release major

PR #747, opened 2019-06-28
Message stuck in sending

typfel:
### Issues

Messages fail to send certain conversations

### Causes

The message depends on a client to be updated but it never updates because the server response doesn't contain the expected `class` field

### Solutions

The  `class` field is optional so we should not require it to be present.


typfel:
@zenkins release patch

PR #749, opened 2019-07-01
Fix: Conversation list not sorted by datetime after restoring from backup

billypchan:
### Issues

After restoring from backup, the conversation's sorting order is lost.

### Causes

The sorting key `lastModifiedDate` is overridden to the date of restoration. 

### Solutions

Do not update the `lastModifiedDate` when it is not nil.


Source/Model/Conversation/ZMConversation+Transport.m
```diff
@@ -80,7 +80,10 @@ - (void)updateWithTransportData:(NSDictionary *)transportData serverTimeStamp:(N
     self.conversationType = [self conversationTypeFromTransportData:[transportData numberForKey:ConversationInfoTypeKey]];
     
     if (serverTimeStamp != nil) {
-        [self updateLastModified:serverTimeStamp];
+         // If the lastModifiedDate is non-nil, e.g. restore from backup, do not update the lastModifiedDate
+        if ([self lastModifiedDate] == nil) {
```
typfel:
```suggestion
        if (self.lastModifiedDate == nil) {
```

typfel:
Let's rename `ZMConversationTransportTests.swift` to `ZMConversation+TransportTests.swift` so that it's more clear that they belong together.

Tests/Source/Model/Conversation/ZMConversationTransportTests.swift
```diff
@@ -0,0 +1,51 @@
...
+            let conversation = ZMConversation.insertNewObject(in: self.syncMOC)
+            let uuid = UUID.create()
+            conversation.remoteIdentifier = uuid
+            let currentTime = Date()
+
+            // assume that the backup date is one day before
+            let lastModifiedDate = currentTime.addingTimeInterval(86400)
+            conversation.lastModifiedDate = lastModifiedDate
+            let serverTimestamp = currentTime
+            let archivedDate = currentTime
```
typfel: I assume `archivedDate` and `silencedDate` is not important for this test? Could we remove them and just pass `currentTime` instead?

billypchan:
@zenkins release patch

PR #751, opened 2019-07-03
Update LegalHoldRequest to parse new event format

typfel:
## What's new in this PR?

Legal hold request event format has changed. This PR updates the `LegalHoldRequest` to parse the new structure.

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-transport/pull/143

## Notes

~~We also re-introduce `target` and `requester` since that information is now available also on the legal hold status endpoint.~~

They are still not available on the legal hold  endpoint so we add them as optional fields.


typfel:
@zenkins release major

PR #752, opened 2019-07-04
Expose isSelfClient on the UserClientType

typfel:
## What's new in this PR?

Expose `isSelfClient()` on the `UserClientType`


typfel:
@zenkins release patch

PR #753, opened 2019-07-04
Legal hold was flipped even if the legal hold client was already known

typfel:
### Issues

Legal hold is active would re-appear on every app launch

### Causes

We re-fetch the self clients on every launch and we would reset the `needsToAcknowledgeLegalHoldStatus` to `true` even if we already knew the client since before.

### Solutions

Only reset `needsToAcknowledgeLegalHoldStatus` to `true`if it's a new client.


typfel:
@zenkins release patch

PR #754, opened 2019-07-08
Fix issues with legal hold being discovered by a message edit

typfel:
### Issues

@VictorWissink discovered several issues with edited messages when discovering legal hold.

1. An edited message doesn't trigger the legal hold discovered alert
2. An edited message which is expired can't be resent by tapping the resend button.
3. The `legal hold discovered` system message is inserted above the edited message which would falsely indicate that all the following messages are also under legal hold.

#### 1.  An edited message doesn't trigger the legal hold discovered alert

This happens when the original message which was edited message is older than the message timeout limit (60 seconds) because the fetch request which collects the undelivered messages was only looking for messages with a `serverTimestamp` not older than the message timeout limit. Edited messages don't modify the `serverTimestamp` instead they assign the `updatedTimestamp`.

#### 2.  An edited message which is expired can't be resent by tapping the resend button.

When resending an edited message we were re-applying the edit but not resetting the `isExpired` flag by calling `[super resend]`.

#### 3. The `legal hold discovered` system message is inserted above the edited message

If a legal hold was discovered by a message we were always inserting the system above the message which discovered it.

#### 1. An edited message doesn't trigger the legal hold discovered alert

Also include messages with an `updatedTimestamp` newer than the timeout limit.

#### 2. An edited message which is expired can't be resent by tapping the resend button.

Call the `[super resend]` method also for edited messages in order to reset the `isExpired` flag.

#### 3. The `legal hold discovered` system message is inserted above the edited message

Insert the legal hold enabled system message at the bottom if it was discovered by an edited message (or reaction).

### DIFF
--- a/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -456,11 +456,10 @@ extension ZMConversation {
         
         let timeoutLimit = Date().addingTimeInterval(-ZMMessage.defaultExpirationTime())
         let selfUser = ZMUser.selfUser(in: managedObjectContext)
-        let undeliveredMessagesPredicate = NSPredicate(format: "%K == %@ AND %K == %@ AND %K == NO AND %K > %@",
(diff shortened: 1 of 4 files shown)


typfel:
@zenkins release patch

PR #756, opened 2019-07-09
Feature: Add support for the team logo in the file cache

billypchan:
## What's new in this PR?

Support team logo image in file asset cache

### TODO
- [x] tests
- [x] `V3Asset` update
- [ ] add get logo image method to `Team` class


Source/Model/Message/V3Asset.swift
```diff
@@ -50,8 +50,8 @@ private let zmLog = ZMSLog(tag: "AssetV3")
 }
 
 
-@objcMembers public class V3Asset: NSObject, ZMImageMessageData {
-    
+@objcMembers public class V3Asset: NSObject, ZMImageMessageData {///TODO: Team protocol?
```
typfel: I would not add any API for the team logo on the `V3Asset` since that's currently tied to asset messages in conversations. 

User profile pictures also are not handled by the `V3Asset`.

typfel: I'd suggest adding a property for setting/getting the team logo on the `Team` entity in a similar fashion as we do for `ZMUser`.

billypchan: @typfel I will add the `logoImage` property in next PR.

Tests/Source/Model/TeamTests.swift
```diff
@@ -330,5 +338,94 @@ class TeamTests: BaseTeamTests {
         XCTAssertFalse(user2CanSeeUser1)
     }
 
-    
+    // MARK: - FileAssetCache
```
typfel: I'd like to move these tests to `FileAssetCacheTests`

Tests/Source/Model/TeamTests.swift
```diff
@@ -20,8 +20,16 @@
 import WireTesting
 @testable import WireDataModel
 
+extension Team {
```
typfel: Let's move this extension to `Model/Utils` test directory.

Source/Model/Message/AssetEncryption.swift
```diff
@@ -43,20 +43,28 @@ extension Cache {
         self.deleteAssetData(encryptedEntryKey)
         return true
     }
+
+    //TODO: test with no sha256Digest
```
typfel: This test exists now?

billypchan: tested together with encryption

PR #757, opened 2019-07-09
Feature: team logo property

billypchan:
## What's new in this PR?

Add `imageData` property to `Team`


Source/Model/Conversation/Team.swift
```diff
@@ -93,3 +93,21 @@ extension Team {
...
+
+// MARK: - Logo Image
+extension Team {
+    public var imageData: Data? {
+        get {
+            return managedObjectContext?.zm_fileAssetCache.assetData(for: self, format: .original, encrypted: false)
+        }
+
+        set {
+            guard let newValue = newValue else {
```
typfel: Why disallow deleting the team logo?

billypchan: added delete asset code for newValue is nil case.

Source/Model/Conversation/Team.swift
```diff
@@ -93,3 +93,22 @@ extension Team {
     }
     
 }
+
+
+// MARK: - Logo Image
+extension Team {
+    public var imageData: Data? {
+        get {
+            return managedObjectContext?.zm_fileAssetCache.assetData(for: self, format: .original, encrypted: false)
```
typfel: I think it makes more sense to use the `.medium` format. We use`.original` when we store an unprocessed image before it's been resized and uploaded to the backend.

PR #762, opened 2019-07-12
Support the unknown legalHoldStatus case

typfel:
## What's new in this PR?

https://github.com/wireapp/wire-ios-protos/pull/35 introduced the `.unknown` case for the `legalHoldStatus`.

The `legalHoldStatus` was previously defaulting to `.disabled` but now it defaults to `.unknown`. This PR updates the code path for updating the legal hold status from a message given this new behaviour.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-protos/pull/35


typfel:
@zenkins release major

PR #763, opened 2019-07-16
Fix: change default legal hold status from .disable to .unknown

billypchan:
## What's new in this PR?

After https://github.com/wireapp/wire-ios-protos/pull/35, update the default value of `MessageContentType` form `.DISABLED` to `.UNKNOWN`.


billypchan:
@zenkins release patch

PR #764, opened 2019-07-17
remove a line break

billypchan:


billypchan:
@zenkins release patch

PR #765, opened 2019-07-19
Fix: zios 11589 app crashes on iOS13 beta 4

billypchan:
### Issues

App crash on iOS 13 when accessing `ZMUpdateEvent.timeStamp()` 

### Causes

The method converting string to Date returns nil on iOS13.

### Solutions

Add nil handling when `ZMUpdateEvent.timeStamp()` is nil. I create a simple string to date method to find out other causes of the crash.

## Todo

Review `+ (nullable instancetype)dateWithTransportString:(nonnull NSString *)transportString;` in `WireTransport` to see needs update or not.


billypchan:
close after https://github.com/wireapp/wire-ios-data-model/pull/766 is approved.

PR #767, opened 2019-07-29
Feature: teamlogo

billypchan:
## What's new in this PR?

Merge team logo feature.


Source/Model/Message/AssetEncryption.swift
```diff
@@ -43,20 +43,26 @@ extension Cache {
...
             return false
         }
         
         let sha256 = encryptedData!.zmSHA256Digest()
-        if sha256 != sha256Digest {
-            self.deleteAssetData(encryptedEntryKey)
-            return false
+
+        if let sha256Digest = sha256Digest {
+            if sha256 != sha256Digest {
```
nicorsm: This could be merged into a unique `if` statement.

billypchan: We changed the requirement for the team image data and encryption is not required finally. I commented on the method for this case.

billypchan:
@zenkins release major

PR #768, opened 2019-08-09
Fix: app crashes when calling addParticipantIfMissing

billypchan:
### Issues

On iOS 13, App crashes at the line 
```
conversation addParticipantIfMissing:[ZMUser userWithRemoteID:updateEvent.senderUUID createIfNeeded:YES inContext:moc] date: [updateEvent.timeStamp dateByAddingTimeInterval:-0.01]];
```

### Causes

`addParticipantIfMissing` does not accept optional value but it has a default value for `date`.  A mehtod with default value should not exposed to objc callers.

### Solutions

Refactor the method to Swift and make sure all arguments are non nil.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -316,12 +316,7 @@ extension ZMConversation {
     ///   - dateOptional: if provide a nil, current date will be used
     @objc(addParticipantIfMissing:date:)
     public func addParticipantIfMissing(_ user: ZMUser, date dateOptional: Date?) {
-        let date: Date
-        if let dataOptional = dateOptional{
-            date = dataOptional
-        } else {
-            date = Date()
-        }
+        let date = dateOptional ?? Date()
 
         guard !activeParticipants.contains(user) else { return }
         
--- a/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
+++ b/Source/Model/Conversation/ZMConversation+SecurityLevel.swift
@@ -311,6 +311,10 @@ extension ZMConversation {
 
     /// Adds the user to the list of participants if not already present and inserts a .participantsAdded system message
     @objc(addParticipantIfMissing:date:)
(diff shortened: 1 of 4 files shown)


Source/Model/Message/ZMOTRMessage+VerifySender.swift
```diff
@@ -0,0 +1,30 @@
...
+import Foundation
+
+extension ZMConversation {
+    @objc
+    public func verifySender(of updateEvent: ZMUpdateEvent, moc: NSManagedObjectContext) {
+
+        guard let senderUUID = updateEvent.senderUUID(),
+              let user = ZMUser(remoteID: senderUUID, createIfNeeded: true, in: moc) else { return }
+
+        addParticipantIfMissing(user, date: updateEvent.timeStamp()?.addingTimeInterval(-0.01))
```
typfel: Let's add `?? Date()` here so that we can avoid the additional `addParticipantIfMissing()` mehod

typfel: And can we then stop exposing to to objective-c?

billypchan: There are some objc callers still need `addParticipantIfMissing:date:` interface.

I refactor the interface to simulate the Swift default value.

Source/Model/Conversation/ZMConversation+SecurityLevel.swift
```diff
@@ -310,8 +310,19 @@ extension ZMConversation {
...
 
     /// Adds the user to the list of participants if not already present and inserts a .participantsAdded system message
+    ///
+    /// - Parameters:
+    ///   - user: the participant to add
+    ///   - dateOptional: if provide a nil, current date will be used
     @objc(addParticipantIfMissing:date:)
-    public func addParticipantIfMissing(_ user: ZMUser, at date: Date = Date()) {
+    public func addParticipantIfMissing(_ user: ZMUser, date dateOptional: Date?) {
+        let date: Date
```
typfel: Can't this be reduced to `let date = dateOptional ?? Date()`

billypchan:
@zenkins release patch

PR #769, opened 2019-08-12
Fix: iOS 13 crash when calling serverTimestampIncludingChildMessages

billypchan:
### Issues

On iOS13, app crashes when calling `serverTimestampIncludingChildMessages`

### Causes

Unknown, but found some ambiguous nullable definitions in its callee `lastChildMessageDate`.  When calling `NSDate.compare`, passing a null argument when causing unexpected result.   
`lastChildMessageDate` is defined as `nonnull` in the header but it may return null.

### Solutions

Refactor `lastChildMessageDate` in Swift and unwrap optional valuables before access.


billypchan:
@zenkins release patch

typfel:
@billypchan seems like we missed deleting the old objective-c implementation here?

PR #770, opened 2019-08-15
Fix team logo not updating

typfel:
### Issues

If the team logo was already assigned it would not update if it was changed.

### Causes

We would not download the new team logo because we would think that we already have the image data locally.

### Solutions

Include the team logo asset id in the cache key to differentiate between old and new team logos.


typfel:
@zenkins release patch

PR #771, opened 2019-08-21
Delete unused code

typfel:
## What's new in this PR?

This method has been converted to Swift but the objective-c version was not deleted.


typfel:
@zenkins release patch

PR #772, opened 2019-08-22
Fix: support for new validators in Swift [DM]

nicorsm:
## What's new in this PR?

Did mostly refactoring of old classes in Objective-C to Swift, in order to support `inout` functions.

## Dependencies

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-utilities/pull/84


Source/Model/User/ZMUser+Validation.swift
```diff
@@ -0,0 +1,133 @@
...
+            throw error
+        }
+    }
+    
+    // E-mail address
+    
+    static func validate(emailAddress: inout String?) throws -> Bool {
+        var mutableEmailAddress: Any? = emailAddress
+        do {
+            return try ZMEmailAddressValidator.validateValue(&mutableEmailAddress)
```
typfel: These validators all take `Any?` which seems silly since we always know what the type will be, `String` in this case.

Why not add the real type to the signature or am I missing something?

typfel:
I see now that we rely on validation feature from core data.

https://developer.apple.com/documentation/coredata/nsmanagedobject

```
Validation Methods
NSManagedObject provides consistent hooks for validating property and inter-property values. You typically shouldn’t override validateValue(_:forKey:). Instead implement methods of the form validate<Key>:error:, as defined by the NSKeyValueCoding protocol. If you want to validate inter-property values, you can override validateForUpdate() and/or related validation methods.
Don’t call validateValue:forKey:error: within custom property validation methods—if you do, you create an infinite loop when validateValue:forKey:error: is invoked at runtime. If you do implement custom validation methods, don’t call them directly. Instead, call validateValue:forKey:error: with the appropriate key. This ensures that any constraints defined in the managed object model are applied.
If you implement custom inter-property validation methods like validateForUpdate(), call the superclass’s implementation first. This ensures that individual property validation methods are also invoked. If there are multiple validation failures in one operation, collect them in an array and add the array—using the key NSDetailedErrorsKey—to the userInfo dictionary in the NSError object you return. For an example, see Managed Object Validation.
```

This means we would need implemented validator methods like described here
https://developer.apple.com/documentation/objectivec/nsobject/1416754-validatevalue

nicorsm:
@typfel I also see that

> You typically use the validation described here only in Objective-C. In Swift, property validation is more idiomatically handled by relying on compiler support for optionals and strong type checking, while using the built-in willSet and didSet property observers to test any run-time API contracts, as described in the Property Observers section of The Swift Programming Language (Swift 4.1).

Might be our case?

typfel:
The problem is that our model classes are not all in Swift so we don't get the compiler validation today. So yeah we could definitely switch that style of validation but not before we've migrated our model classes to Swift.

I'm already worried about this change snow balling in scope, let's not make it even bigger. For now I think we should keep validating using the old style:

```
@objc func validateSomeString(_ ioValue: AutoreleasingUnsafeMutablePointer<AnyObject?>) throws {
    // Test, and possibly replace the value here; or throw an error
}

**Comment by typfel** [comment]
`

File: Tests/Source/Model/ManagedObjectValidationTests.m [hunk 2]
```diff
@@ -63,17 +57,11 @@ - (void)testThatValidationOnNonUIContextAlwaysPass
...
         id value = user.name;
         
-        id validator = [OCMockObject mockForClass:[StringLengthValidator class]];
-        [[[validator reject] andForwardToRealObject] validateValue:[OCMArg anyObjectRef]
-                                               minimumStringLength:2
-                                               maximumStringLength:64
-                                                 maximumByteLength:256
-                                                             error:[OCMArg anyObjectRef]];
-        
+        value = [ManagedObjectValidationTestsUtility validateStringLength:value minimum:2 maximum:64 byteLength:256];
```
typfel: This test is testing that validation always passes when performed on a non UI managed object context. It was previously doing this using `OCMock` by failing the test if the validation method was called. the `ManagedObjectValidationTestsUtility` does not replace that though so we need to re-think this test.

One approach could we test assign the name to an invalid value and see that it's still reported as valid.

https://github.com/wireapp/wire-ios-data-model/blob/40d7e1c943a1a107b0ec663c9b3cb2eefa0d846b/Source/Model/ZMManagedObject.m#L725

nicorsm: As the method is back and available from Objective-C, we could simply restore the tests at the previous state and remove `ManagedObjectValidationTestsUtility`.

typfel: Ok, up to you depending how much work it is. In the long term we would like to get rid of `OCMock` though.

Tests/Source/Model/ManagedObjectValidationTests.m
```diff
@@ -72,8 +74,6 @@ - (void)testThatValidationOnNonUIContextAlwaysPass
         
         BOOL result = [user validateValue:&value forKey:@"name" error:NULL];
         XCTAssertTrue(result);
-        [validator verify];
```
typfel: I think these deleted lines needs to be restored?

nicorsm: Sure!

nicorsm:
@zenkins release major

PR #774, opened 2019-08-27
Add method for deleting messages older than given date

typfel:
## What's new in this PR?

Add method for deleting messages older than given date in an efficient manner.

## TODO

- [x] Delete any cached assets associated with the deleted messages.


typfel:
@zenkins release minor

PR #775, opened 2019-08-29
Add missing methods for conversation permissions

typfel:
## What's new in this PR?

Add missing methods for checking if the self user is allowed to perform certain actions on conversations. This business logic used be implemented in UI where it was much harder to test using unit tests.

## Notes

I've also fixed a bug in `isGuest(in: ZMConversation)` where it would consider a team user in another team's conversation as a guest._


typfel:
@zenkins release minor

PR #776, opened 2019-09-02
fix: make savePendingLastRead public

billypchan:
## What's new in this PR?

`wire-ios` project access `savePendingLastRead` but it is not public. After refactoring the UI code it is not able to access `savePendingLastRead`.


billypchan:
@zenkins release patch

PR #779, opened 2019-09-06
Feature: Support extracting conversationId from .teamConversationDelete event

typfel:
## What's new in this PR?

Support extracting the conversation uuid from `.teamConversationDelete` event.


typfel:
@zenkins release patch

PR #785, opened 2019-09-25
Fix: partner not being able to to create or fetch a team 1:1 conversation

typfel:
### Issues

A team member with the partner role can't open team 1:1 conversation

### Causes

The method for fetching a team 1:1 conversation was checking that the self user has the permission to create conversations, which partners don't have.

### Solutions

Remove incorrect permission check.


typfel:
@zenkins release patch

PR #788, opened 2019-09-30
Fix: crash on canManageAccess

nicorsm:
### Issues

It seems that in `Conversation+AccessMode.swift`, declared in wire-ios-sync-engine, the `canManageAccess` flag is `false`. This means that the "Guests and services" row is shown without meeting the proper conditions.

### Solutions

I've added more conditions inside the logic that's responsible for displaying the Guests row.


typfel:
So we have two methods for determining if a user can modify "Guests and services"? This is indeed problematic and we should remove one of them. Let's keep the one in data model, we can keep the method on `ZMConversation` not break the UI and just forward the call.

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -75,7 +75,8 @@ public extension ZMUser {
     
     @objc(canModifyAccessControlSettingsInConversation:)
     func canModifyAccessControlSettings(in conversation: ZMConversation) -> Bool {
-        guard !isGuest(in: conversation), conversation.isSelfAnActiveMember else { return false }
+        guard !isGuest(in: conversation), conversation.teamRemoteIdentifier != nil,
+            isTeamMember, conversation.isSelfAnActiveMember else { return false }
```
billypchan: I think these checkers are similar to other siblings function e.g. `canModifyEphemeralSettings` & `canAddUser`.
We can group these 4 checkers together as a new function.

PR #789, opened 2019-10-01
Feature: support for marking conversations as favorites 

typfel:
## What's new in this PR?

Add the `isFavorite` property to `ZMConversation` which allows you mark/unmark a conversation as a favourite.

The list of favourite conversations can be accessed & observed via the `ConversationDirectory`.


typfel:
@zenkins release minor

PR #790, opened 2019-10-02
Fix: compare changeInfo.conversationList pointer with conversations

billypchan:
## What's new in this PR?

Compare `changeInfo.conversationList` with conversation list's address instead of value to prevent difference list may have the same values issue.


billypchan:
@zenkins release patch

billypchan:
@zenkins release patch

PR #792, opened 2019-10-08
Feature: add properties necessary for syncing folders

typfel:
## What's new in this PR?

Add `modifiedKeys` property to the `Label` entity to keep track of which properties have been modified locally and need to be synced.

Add `markedForDeletion` property to the `Label` entity to keep track of Labels which have been deleted which needs to be synced. A label with `markedForDeletion = true` is filtered out from the list of available folders.


typfel:
@zenkins release major

PR #793, opened 2019-10-09
fix: expose ZMConversation.folder

billypchan:
## What's new in this PR?

Allow UI to get Conversation's folder label


billypchan:
@zenkins release minor

PR #794, opened 2019-10-09
Fix: conversations in folder also appearing under groups & people

typfel:
### Issues

If you add a conversation to a folder is would still be visible under the groups or people section.

### Causes

We were not filtering out conversations in folders from the groups and people sections.

### Solutions

Update conversation list predicates to filter out conversations in folders.


typfel:
@zenkins release patch

PR #796, opened 2019-10-10
Fix: folder created while in the background does not appear until restart

typfel:
### Issues

If you create folder on a second device while the app is in the background these would not appear until the app is restarted.

### Causes

When the app is in the background we do not observe any changes so we need re-fetch all conversations & folders which were potentially created in the background. 

We were not re-fetching the folders.

### Solutions

Also re-fetch the folders when re-starting the observation system.


typfel:
@zenkins release patch

PR #797, opened 2019-10-10
Fix: guard against moving a conversation to a deleted label

typfel:
### Issues

It's was observed that the app crashed when trying to move a conversation to a folder without a name.

### Causes

The folder without a name is most likely a deleted label which UI for some reason still have a reference to, attempting to establish a relationship between a conversation and a the deleted label will trigger an exception in core data.

### Solutions

Guard against deleted folders.


typfel:
@zenkins release patch

PR #798, opened 2019-10-10
Feature: add flag for re-fetching labels

typfel:
## What's new in this PR?

Add `needsToRefetchLabels` flag on the user to signal that we should re-fetch the labels.


typfel:
@zenkins release minor

PR #799, opened 2019-10-11
Fix: service conversations appearing under groups

typfel:
### Issues

Service conversation would appear under the group section

### Causes

The Group section predicate was not filtering out fake 1:1 conversations

### Solutions

Filter out fake 1:1 conversations.


typfel:
@zenkins release patch

PR #800, opened 2019-10-11
Fix: the initial sorting order folders is inverted

typfel:
### Issues

The initial order of the folder is inverted.

### Causes

Comparator was flipped which was compensated for in the insertion sort method.

### Solutions

Correct the Comparator and undo the correction in the insertion sort method.


typfel:
@zenkins release patch

PR #801, opened 2019-10-14
Fix: labels sorting being case sensitive

typfel:
### Issues

Folders with capital letters would be sorted before folders starting with a lower case letter.

### Causes

The default `NSSortDescriptor` compare method is case sensitive.

### Solutions

Use `NSString.localizedStandardCompare(_:)` as the comparison method for case insensitive sorting.


typfel:
@zenkins release patch

PR #804, opened 2019-10-29
fix: app crash when init a V2Asset

billypchan:
## What's new in this PR?

add a guard to prevent crashing when the message has nil `managedObjectContext` (It is cause when Collection screen want to access an image asset)


billypchan:
It should show a empty collection view cell instead of crashing

billypchan:
@zenkins release patch

PR #805, opened 2019-10-29
Fix: ZMManagedObjectContextProvider optional properties

billypchan:
## What's new in this PR?

Translate `ZMManagedObjectContextProvider` to Swift and make its properties as optional.


Source/Model/User/ZMSearchUser.swift
```diff
@@ -511,9 +513,9 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
                 context.saveOrRollback()
                 
                 let objectId = user.objectID
-                self.contextProvider?.managedObjectContext.performGroupedBlock {
-                    self.user = self.contextProvider?.managedObjectContext.object(with: objectId) as? ZMUser
-                    self.contextProvider?.managedObjectContext.searchUserObserverCenter.notifyUpdatedSearchUser(self)
+                self.contextProvider?.managedObjectContext?.performGroupedBlock {
+                    self.user = self.contextProvider?.managedObjectContext?.object(with: objectId) as? ZMUser
+                    self.contextProvider?.managedObjectContext?.searchUserObserverCenter.notifyUpdatedSearchUser(self)
```
johnxnguyen: This could be cleaned up a lot with some `if let`s or `guard let`s.

billypchan: refactored for the 2 contexts unwrapping

johnxnguyen: I think this can still be refactored to remove a lot of the optionals. For instance, we need both the sync and ui contexts, so we can just unwrap them in a `guard` first....

```swift
guard 
  let syncMoc = contextProvider?.syncManagedObjectContext,
  let uiMoc = contextProvider?.managedObjectContext
  else { return }

syncMoc.performGroupedBlock {
    ...
  uiMoc.performGroupedBlock {
    ...
  }
}
```

billypchan:
@zenkins release major

PR #822, opened 2019-12-12
Parse the role of users from conversation metadata

marcoconti83:
## What's new in this PR?

The role of users is parsed from the conversation metadata. This is achieved by adding the ability to add participant by specifying the role. The same is now possible at conversation creation.

Additionally, the naming of some methods was changed as per https://github.com/wireapp/ios-architecture/issues/88

**NOTE**: this does not include parsing the role self user. This will be addressed in a separate PR.


Source/Model/Conversation/ZMConversation+Participants.swift
```diff
@@ -90,78 +90,187 @@ extension ZMConversation {
...
     ///
-    /// - Parameter users: users to union
+    /// The method will handle the case when the participant is already there, so it's safe to call
+    /// it multiple time for the same user. It will update the role if the user is already there with
+    /// a different role.
+    ///
+    /// The method will also check if the addition of the users will change the verification status, the archive
+    /// status, etc.
+    @objc
+    public func addParticipantAndUpdateConversationState(user: ZMUser, role: Role?) {
```
billypchan: we may add a Swift only interface with `role: Role? = nil` to save some typing.

Tests/Source/Model/Conversation/ZMConversationExternalParticipantsStateTests.swift
```diff
@@ -120,16 +120,16 @@ class ZMConversationExternalParticipantsStateTests: ZMConversationTestsBase {
...
 
             case .personal:
                 let otherUser = createUser(in: uiMOC)
-                conversation.internalAddParticipants([otherUser])
+                conversation.addParticipantAndUpdateConversationState(user: otherUser, role: nil)
 
             case .service:
                 let service = createService(in: uiMOC, named: "Bob the Robot")
-                conversation.internalAddParticipants([service as! ZMUser])
+                conversation.addParticipantAndUpdateConversationState(user: service as! ZMUser, role: nil)
```
billypchan: maybe we can change the user param to `UserType` to prevent casting.

Source/Model/Conversation/ZMConversation+Participants.swift
marcoconti83: I compiled this all the way to sync engine.

Tests/Source/Model/Conversation/ZMConversationExternalParticipantsStateTests.swift
marcoconti83: Good point

marcoconti83: I can change that during integration

PR #823, opened 2019-12-12
Fix/ui tests issues

billypchan:
## What's new in this PR?

Create `ZMConnection.add(user: ZMUser)` method to add a user to the connection's conversaton.

### DIFF
--- a/Source/Model/Connection/ZMConnection.m
+++ b/Source/Model/Connection/ZMConnection.m
@@ -72,12 +72,13 @@ static NSString * const LastUpdateDateInGMTKey = @"lastUpdateDateInGMT";
     if (conversation == nil) {
         connection.conversation = [ZMConversation insertNewObjectInManagedObjectContext:user.managedObjectContext];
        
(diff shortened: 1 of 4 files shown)


Source/Model/Connection/ZMConnection.m
```diff
@@ -72,12 +72,13 @@ + (instancetype)insertNewSentConnectionToUser:(ZMUser *)user existingConversatio
     if (conversation == nil) {
         connection.conversation = [ZMConversation insertNewObjectInManagedObjectContext:user.managedObjectContext];
        
-        ParticipantRole * participantRole = [ParticipantRole createWithManagedObjectContext:user.managedObjectContext user:user conversation:connection.conversation];
-        [connection.conversation.participantRoles setByAddingObject:participantRole];
+        [connection addWithUser:user];
```
marcoconti83: Why not use:
```
[conversation addParticipantAndUpdateConversationStateWithUser:otherUser role:nil]
```
It will create the participant role only if the user is not there already.

marcoconti83: I see, it can not be there because we just created the conversation!

PR #825, opened 2019-12-13
Parse the self user role from conversation metadata

marcoconti83:
## What's new in this PR?

Parse the self user role from conversation metadata. 

This has been tested with integration with upstream frameworks, and it also addresses some of the changes needed for sync engine.


marcoconti83:
Opened on the wrong branch!

PR #827, opened 2019-12-13
Fix comments on connection add

marcoconti83:
## What's new in this PR?

Updated some comment (will use this to trigger a release)


marcoconti83:
@zenkins release major

PR #829, opened 2019-12-13
Test release scripts

marcoconti83:
This commit is really a test for the release process


marcoconti83:
@zenkins release patch

PR #831, opened 2019-12-15
Fix/sync engine fixes

marcoconti83:
Fix all tests from sync engine and upstream


marcoconti83:
@zenkins release patch

PR #832, opened 2019-12-15
Add flags to download roles for conversation and teams

marcoconti83:
## What's new in this PR?

Add flags to mark teams and conversation as needing to download roles, and make sure those flags are set then migrating to this version of data model


marcoconti83:
@zenkins release minor

PR #834, opened 2019-12-16
Feature/add new group permissions

KaterinaWire:
## What's new in this PR?

Add new permission logic for group conversations.


zenkins:

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -41,18 +67,40 @@ public extension ZMUser {
...
+            guard conversation.teamRemoteIdentifier == nil || !isGuest(in: conversation), conversation.isSelfAnActiveMember else { return false }
+            return permissions?.contains(.addRemoveConversationMember) ?? true
+        }
     }
     
     @objc(canRemoveUserFromConversation:)
     func canRemoveUser(from conversation: ZMConversation) -> Bool {
-        return canAddUser(to: conversation)
+        if conversation.conversationType == .group {
+            let canModifyConversationName = self.participantRoles.filter({$0.conversation == conversation}).first?.role?.actions.contains(where: {$0.name == ActionNames.removeConversationMember.name})
```
marcoconti83: I would extract the following:
```
self.participantRoles.filter({$0.conversation == conversation}).first?.role?.actions.contains(where: {$0.name == xxx})
```

To a method, something like `ZMUser.hasRoleWithAction(action: ActionNames, conversation: ZMConversation)`

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -41,18 +67,40 @@ public extension ZMUser {
     
     @objc(canAddUserToConversation:)
     func canAddUser(to conversation: ZMConversation) -> Bool {
-        guard conversation.teamRemoteIdentifier == nil || !isGuest(in: conversation), conversation.isSelfAnActiveMember else { return false }
-        return permissions?.contains(.addRemoveConversationMember) ?? true
+        if conversation.conversationType == .group {
+            let canModifyConversationName = self.participantRoles.filter({$0.conversation == conversation}).first?.role?.actions.contains(where: {$0.name == ActionNames.addConversationMember.name})
```
billypchan: We can create a helper method to search for a action name of a user, e.g.
```

File: Tests/Source/Model/User/ZMUserTests+Permissions.swift [hunk 3]
```diff
@@ -509,4 +511,90 @@ class ZMUserTests_Permissions: ModelObjectsTests {
         // then
         XCTAssertFalse(ZMUser.selfUser(in: uiMOC).canModifyTitle(in: conversation))
     }
+    
+    func testThatConversationTitleCanBeModified_ByGroupParticipant() {
```
marcoconti83: I would also add opposite tests, tests that the user can not modify conversation if the role action is not there

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -18,6 +18,32 @@
 
 import Foundation
 
+enum ActionNames {
```
billypchan: I would suggest use a singlar from `ConversationAction`

KaterinaWire: Fixed

Source/Model/User/ZMUser+Permissions.swift
KaterinaWire: Added a helper method `hasRoleWithAction`

PR #835, opened 2019-12-16
[HACK] Disable code optimization

marcoconti83:
Currently our tests are crashing when the optimization is enabled. We'll temporarily disable it to unlock the team from writing more tests. We will revert this before the release.


marcoconti83:
@zenkins release patch

PR #839, opened 2019-12-17
Do not mark to download roles for team conversation

marcoconti83:
## What's new in this PR?

Do not mark conversations that are part of the team as needing to download roles. 

These roles will be downloaded with the team, there is no need to download them individually per conversation.


Source/Model/Conversation/ZMConversation+Patches.swift
```diff
@@ -106,7 +106,9 @@ extension ZMConversation {
         (moc.executeFetchRequestOrAssert(groupConversationsFetch) as! [ZMConversation]).forEach {
             guard $0.isSelfAnActiveMember else { return }
             $0.needsToBeUpdatedFromBackend = true
-            $0.needsToDownloadRoles = true
+            if $0.team == nil {
+                $0.needsToDownloadRoles = true
+            }
```
johnxnguyen: You could inline this:

```swift
$0.needsToDownloadRoles = $0.team == nil
```

billypchan:
@zenkins release patch

PR #840, opened 2019-12-17
Expose default member role name

johnxnguyen:
## What's new in this PR?

Add and expose the default conversation role "member" for use in upstream projects and expose for objective c.


johnxnguyen:
@zenkins release minor

PR #841, opened 2019-12-17
fix: role change not trigger conversaton change

billypchan:
### Issues

When the role of a user changes, the `conversationDidChange` event is not observed.

### Causes

The conversation is not affected by role change.

### Solutions

add `#keyPath(ZMConversation.participantRoles.role` to the `keyPathsForValuesAffectingLocalParticipantRoles`


Source/Notifications/DependencyKeyStore.swift
```diff
@@ -188,7 +188,12 @@ class DependencyKeyStore {
     }
     
     /// Returns the inverse of keyPathsForValuesAffectingValueForKey, all observable keys that are affected by `key`
-    func observableKeysAffectedByValue(_ classIdentifier: String, key: String) -> Set<String>{
+    ///
+    /// - Parameters:
+    ///   - classIdentifier: the class's id, e.g. "Conversation"
+    ///   - key: the key, e.g. "participantRoles.role"
+    /// - Returns: the inverse of keyPathsForValuesAffectingValueForKey, all observable keys that are affected by `key`
+    func observableKeysAffectedByValue(_ classIdentifier: String, key: String) -> Set<String> { ///TODO: "participantRoles.role" is not affecting convo??
```
marcoconti83: Is the TODO still valid?

PR #842, opened 2019-12-17
Update UserType

johnxnguyen:
## What's new in this PR?

This PR adds a new protocol method to `UserType` to return the role in a given conversation. Some functionality has also been moved from the UI project.

This will help facilitate mocking the role for the `MockUser`.

### DIFF
--- a/Source/Model/User/UserType+ZMUser.swift
+++ b/Source/Model/User/UserType+ZMUser.swift
@@ -1,0 +1,35 @@
+//
+// Wire
+// Copyright (C) 2019 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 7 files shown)


Source/Model/User/UserType+ZMUser.swift
```diff
@@ -0,0 +1,35 @@
...
+    /// Return the ZMUser associated with the generic user, if available.
+    var zmUser: ZMUser? {
+        if let searchUser = self as? ZMSearchUser {
+            return searchUser.user
+        } else if let zmUser = self as? ZMUser {
+            return zmUser
+        } else {
+            return nil
+        }
+    }
```
billypchan: We can just define the interface and put the implement in `ZMSearchUser` and `ZMUser`

johnxnguyen: Actually, does it make sense the `UserType` protocol declare a requirement to return a concrete `ZMUser` ? Perhaps it's better to leave it as it is?

johnxnguyen:
@zenkins release major

PR #843, opened 2019-12-17
Make method public

marcoconti83:
## What's new in this PR?

Expose method for sync engine


marcoconti83:
@zenkins release minor

PR #845, opened 2019-12-17
Parse self user role in conversation JSON fragment for self

marcoconti83:
## What's new in this PR?

Previously we were parsing the self role in the entire JSON payload of a conversation. 

However, we use the `updateSelfStatus` method also to parse member update events JSON fragments, and that method does not parse the self role, relying on the outer method to parse the entire conversation JSON.

This PR moves the self role parsing inside that method, so that it can be used by sync engine to parse the self role also for events.


marcoconti83:
@zenkins release minor

PR #848, opened 2019-12-18
Fix: Missing UI elements (avatars, participants list, text input bar) after updating to conversation role dev build

billypchan:
## What's new in this PR?
refetch all conversation after migration.


marcoconti83:
I have an idea on how to mitigate this one

PR #849, opened 2019-12-18
Fix optional field for core data

marcoconti83:
## What's new in this PR?

Fix crash when optimization is turned on. This was caused by `ParticipantRole.conversation` being optional in the Core Data schema but not in the Swift class.


Source/Model/ConversationRole/ParticipantRole.swift
```diff
@@ -26,7 +26,7 @@ let ZMParticipantRoleModificationToSyncKey  = #keyPath(ParticipantRole.rawOperat
 final public class ParticipantRole: ZMManagedObject {
     
     @objc @NSManaged var rawOperationToSync: Int16
-    @NSManaged public var conversation: ZMConversation
+    @NSManaged public var conversation: ZMConversation?
```
marcoconti83: This line is the root cause

marcoconti83:
@zenkins release patch

PR #854, opened 2019-12-19
Use primitive value `isSelfAnActiveMember` for migration

marcoconti83:
## What's new in this PR?

With version 2.78 of the data model, the `isSelfAnActiveMember` properties is now a computed property based on the `participantRoles`. In 2.77, `isSelfAnActiveMember`  was an actual core data (boolean) property.

There is a migration step that reads the old value for `isSelfAnActiveMember`  and update `participantRole`. However the migration step uses the new computed property (based on `participantRoles` rather than reading from the Core Data property. Therefore it would never migrate property, as there is never any `participantRoles` at the beginning of the migration.

This PR reads from the proper Core Data value during migration.


marcoconti83:
@zenkins release patch

PR #855, opened 2019-12-19
Fix/observe role change

marcoconti83:
### Issues

It is not possible to observe changes in one-to-one relationships via snapshots, as the snapshot only stores whether the relationship is set to something not `nil`.

Therefore if a relationship changes from object A to object B (as opposed to from `nil` to object B), the snapshot will not report a change.

### Solutions

Store the `objectID` of the `NSManagedObject` that the relationship is pointing to, instead of a boolean. This will allow to detect that the value changed when going from object A to object B.


marcoconti83:
@zenkins release minor

PR #856, opened 2019-12-19
Fix: Personal group creator (and Admin) should not see Delete Group option

billypchan:
## What's new in this PR?

Only allow user has team to delete a conversation


marcoconti83:
@zenkins release patch

PR #859, opened 2019-12-20
Fix: allow partner to create 1-on-1 conversaton but not group conversation

billypchan:
## What's new in this PR?

add `type` parameter to `canCreateConversation` checker. Partner team role can only create 1-on-1 conversation but not other kinds.

add `type` parameter to `insertGroupConversation`.


billypchan:
@zenkins release major

PR #860, opened 2019-12-23
Refetch roles when needed

marcoconti83:
## What's new in this PR?

Add a method to make sure conversation roles are refetched when needed. This method is called every time a conversation is updated from payload.

This solves the bug when roles are not fetched when a user is added to a conversation. Previously, the check on whether roles should be fetched was done at the sync engine level, when the conversation payload was not yet parsed. At the time of the check, we would not know if the conversation was a group or in a team, and the check would always fail due to incomplete information.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Participants.swift
+++ b/Source/Model/Conversation/ZMConversation+Participants.swift
@@ -155,6 +155,7 @@ extension ZMConversation {
         
         let addedSelfUser = doesExistsOnBackend && addedRoles.contains(where: {$0.user.isSelfUser})
         if addedSelfUser {
+            self.markToDownloadRolesIfNeeded()
             self.needsToBeUpdatedFromBackend = true
         }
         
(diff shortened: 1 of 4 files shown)


Source/Model/Conversation/ZMConversation+Participants.swift
```diff
@@ -293,6 +294,23 @@ extension ZMConversation {
...
     }
+    
+    /// Check if roles are missing, and mark them to download if needed
+    @objc public func markToDownloadRolesIfNeeded() {
+        guard self.conversationType == .group else { return }
+        
+        let selfUser = ZMUser.selfUser(in: self.managedObjectContext!)
+        let notInMyTeam = self.teamRemoteIdentifier == nil ||
+            selfUser.team?.remoteIdentifier != self.teamRemoteIdentifier
+
```
billypchan: TODO: clean up with `UserType.isOnSameTeam` method

marcoconti83:
@zenkins release minor

PR #861, opened 2019-12-23
Make conversation permission based on conversation roles

marcoconti83:
## What's new in this PR?

Conversation action permissions were based on a mix of conversation roles and team roles. However, according to the new conversation roles specs, the actions should be based exclusively on the conversation role, with a few exceptions:

- deletion -> user is creator && conversation admin
- add/remove service -> user is in the team && conversation admin

Everything else is just based on conversation role.

I updated the code to reflect this, and removed a lot of old tests that were checking these permissions based on the user type or team membership.

### Note 

This PR includes the deletion of a lot of tests. Unfortunately, those tests were testing the wrong behaviour.


marcoconti83:
@zenkins release minor

PR #862, opened 2019-12-27
Using correct version number in migration

marcoconti83:
## What's new in this PR?

The migration for conversation roles was started from the wrong version. The previous released data model version was 197. The migration was set to run if the old version was before 197, which was causing it not to be run at all.


marcoconti83:
@zenkins release patch

PR #863, opened 2019-12-27
Fix: add properties to UserType and create SystemMessageNewConversationProperties

billypchan:
## What's new in this PR?

Change `ZMConversationMessage.sender`'s type from `ZMUser` to `UserType` to allow mocking.
Create `SystemMessageNewConversationProperties` for additional properties of `ZMSystemMessage` to allow mocking. 
Add `remoteIdentifier` and `func displayName(in conversation: ZMConversation?) -> String` to `UserType` protocol.


Source/Model/Message/ConversationMessage.swift
```diff
@@ -235,7 +236,7 @@ extension ZMMessage : ZMConversationMessage {
 
 extension ZMMessage {
     
-    @NSManaged public var sender : ZMUser?
+    @NSManaged public var sender : UserType?
```
johnxnguyen: This is a managed property, can you use the `UserType` here?

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -169,14 +169,14 @@ extension ZMGenericMessage {
...
-            guard let sender = message.sender else {
+            guard let sender = message.sender as? ZMUser else {
                 zmLog.error("sender of deleted ephemeral message \(String(describing: self.deleted.messageId)) is already cleared \n ConvID: \(String(describing: conversation.remoteIdentifier)) ConvType: \(conversation.conversationType.rawValue)")
                 return Set(arrayLiteral: selfUser)
             }
             
             // if self deletes their own message, we want to send delete msg
             // for everyone, so return nil.
-            guard !sender.isSelfUser else { return nil }
+            guard false == sender.isSelfUser else { return nil }
```
johnxnguyen: `sender` is already unwrapped, you can still use `!`

Source/Notifications/ObjectObserverTokens/MessageChangeInfo.swift
```diff
@@ -187,7 +187,7 @@ extension ZMSystemMessage {
     }
 
     public var senderChanged : Bool {
-        if self.usersChanged && (self.userChangeInfo?.user as? ZMUser ==  self.message.sender){
+        if usersChanged && userChangeInfo?.user as? ZMUser ==  message.sender as? ZMUser {
```
johnxnguyen: `userChangeInfo.user` and `message.sender` are both `UserType`, do you need to cast them?

Source/Model/User/UserType.swift
```diff
@@ -26,7 +26,10 @@ public protocol UserConnectionType: NSObjectProtocol {
 }
 
 @objc
-public protocol UserType: NSObjectProtocol {
+public protocol UserType: class, NSObjectProtocol {
+    
+    /// The remote identifier
+    var remoteIdentifier: UUID? { get set }
```
johnxnguyen: I don't think this a suitable property to expose on the `UserType`, we really only want to expose properties for the UI and this is more business logic related.

PR #864, opened 2020-01-02
fixed searching issue

billypchan:
### Issues

Searching has no return after search with a query ended with space

### Causes

The searching does not finish when a empty query

### Solutions

Trim the query string before break down it into array.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Predicates.swift
+++ b/Source/Model/Conversation/ZMConversation+Predicates.swift
@@ -66,8 +66,9 @@ extension ZMConversation {
     
     @objc
     public class func normalize(_ string: String) -> [String] {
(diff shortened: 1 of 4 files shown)


marcoconti83:
I don't understand how the code was broken. Can you explain why it was freezing?

billypchan:
@marcoconti83 When executing a query with precidate `NSPredicate(format: "ANY %K.user != %@ AND ANY %K.user.normalizedName MATCHES %@", ZMConversationParticipantRolesKey, selfUser, ZMConversationParticipantRolesKey, strSearchQuery)`, if the later parameter is empty string, the query is never done.

marcoconti83:
Thanks for the explanation @billypchan !

billypchan:
@zenkins release patch

PR #865, opened 2020-01-02
Fix: migration to roles failing due to duplicate Roles being inserted

typfel:
### Issues

Group participants are incorrectly migrated as members instead of admins.

### Causes

During the migration of team group conversations a new `Role` was created for each conversation instead them all sharing the the same `Role` as expected. The duplicate roles would later be deleted when the `TeamRolesDownloadRequestStrategy` syncs the team roles, leaving the participants without a role and thus default to be "members".

### Solutions

Fetch any existing `Role` before creating a new `Role` for a team conversation.


marcoconti83:
@zenkins release patch

PR #866, opened 2020-01-02
Fix: Search never completes if it includes a space

marcoconti83:
## Issue
A search for conversation never completes if the query contains a space, due to the complexity of the generated fetch predicate. 

In particular, I suspect that using multiple`ANY x.y == z` with the same `x` in the same predicate will cause this issue. 

## Solution
Instead of using multiple `ANY x.y == z`, I used a `SUBQUERY` with the same requirements in `AND`.


marcoconti83:
@zenkins release minor

PR #869, opened 2020-01-03
[Fix] Extend UserType

johnxnguyen:
### Issues

The `UserType` protocol is designed to abstract user related properties and methods away from the several possible concrete user classes (such as `ZMUser`, `ZMSearchUser` and `MockUser`). Unfortunately this abstraction is being degraded because it leaks conforming types.

Accessing this `zmUser` property is bad practice because it defeats the purpose of the `UserType` abstraction and makes it hard to mock code since the production code may use code not defined in `UserType`.

### Causes

There exists a `zmUser` property in a protocol extension which returns a concrete `ZMUser` instance if it exists. This property is used in several places in the UI project, where there is a `UserType` instance available but the code demands `ZMUser` only implementations.

### Solutions

The `UserType` protocol should be extended (more requirements) so the `zmUser` property is redundant and can be removed.


WireDataModel.xcodeproj/project.pbxproj
```diff
@@ -2161,7 +2158,9 @@
 				F9C9A4FD1CAD5DF10039E10C /* Products */,
 				F9C9A6701CAD779E0039E10C /* Frameworks */,
 			);
+			indentWidth = 4;
 			sourceTree = "<group>";
+			tabWidth = 4;
```
billypchan: it should not included in the PR.

johnxnguyen: Removed 👍

johnxnguyen:
@zenkins release major

PR #872, opened 2020-01-09
[Fix] crash by unwrapping optional values before attempting to use them

typfel:
### Issues

We've seen reports that the app crashes while migrating to the new user roles

### Causes

Optional values are force unwrapped

### Solutions

Unwrap optional values before attempting to use them.


typfel:
> The managed object subclass `ZMConnection` doesn't specify nullability. Probably others are the same. Do you think it's worth going through (in a separate PR) to correct these?

Yes, not specifying nullability is very dangerous since the compiler doesn't warn you.

typfel:
@zenkins release patch

PR #873, opened 2020-01-10
[Refactor] Send regular messages using Swift Protobuf API

typfel:
## What's new in this PR?

Refactor the message sending methods to use the the Swift Protobuf API. This is compatible with the existing Objective-C implementation since we serialize and store the protobufs as a data blob.


typfel:
@zenkins release patch

PR #874, opened 2020-01-14
Fix: Application crashes when searching for contacts with symbol

billypchan:
### Issues

Application crashes when searching for contacts with symbol

### Causes

`String.words` methods returns empty string array when the query string is "@"

### Solutions

Append the whole string as a word when the return array is empty. Escape the string for regex.

### DIFF
--- a/Tests/Source/Model/Utils/String+WordTests.swift
+++ b/Tests/Source/Model/Utils/String+WordTests.swift
@@ -21,7 +21,7 @@ import XCTest
 
 final class StringWordsTests: XCTestCase {
     
-    func testThatSetenceIsSplited() {
+    func testThatSentenceIsSplitted() {
         //given
         let sut = "once upon a time"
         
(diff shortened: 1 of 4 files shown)


Tests/Source/Model/Utils/String+WordTests.swift
```diff
@@ -0,0 +1,45 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import WireDataModel
+
+final class StringWordsTests: XCTestCase {
+    
+    func testThatSetenceIsSplited() {
```
typfel:
```suggestion
    func testThatSentenceIsSplitted() {
```

Tests/Source/Model/Utils/String+WordTests.swift
```diff
@@ -0,0 +1,45 @@
...
+        let sut = "once upon a time"
+        
+        //when
+        let words = sut.words
+        
+        //then
+        XCTAssertEqual(words, ["once", "upon", "a", "time"])
+    }
+    
+    func testThatSingleSymbolIsSplitedAsAWord() {
```
typfel:
```suggestion
    func testThatSingleSymbolIsSplittedAsAWord() {
```

billypchan:
@zenkins release patch

PR #876, opened 2020-01-15
[Fix] Update UserType to hide Role

johnxnguyen:
### Issues

The `UserType` abstraction exposes the CoreData class `Role`. This makes it hard to mock on the UI layer.

### Solutions

The requirement `func role(in conversation: ZMConversation) -> Role?` is not really used on the UI layer, so it doesn't make sense to include it in the `UserType`. We can remove it if we replace it with `func isGroupAdmin(in conversation: ZMConversation) -> Bool`, which is all the UI layer really cares about.


johnxnguyen:
@zenkins release major

PR #878, opened 2020-01-16
[Fix] image quotes are getting lost

typfel:
### Issues

If you reply to an image it's not included as quote.

### Causes

The quoted image message was casted to a `ZMClientMessage`, which fails since image messages are of type `ZMAssetClientMessage`.

### Solutions

Cast quoted messages the base class `ZMOTRMessage` instead.


typfel:
@zenkins release patch

PR #880, opened 2020-01-21
Fix/make as a public appendSelfConversationWithLastReadOf method

KaterinaWire:
## What's new in this PR?

Fix a mistake from the previous PR. 
- appendSelfConversationWithLastReadOf() method should be public.
- change a parameter name for appendSelfConversationWithLastReadOf method.


KaterinaWire:
@zenkins release patch

PR #881, opened 2020-01-22
[Refactor] Hide Concrete Users in User Classes

johnxnguyen:
## What's new in this PR?

This PR is part of an ongoing effort to remove all concrete user objects from the UI project. See the architecture issues for more info: https://github.com/wireapp/ios-architecture/issues/89

### Changes

`ZMUser` and `ZMSearchUser` references in are now hidden in `UserChangeInfo.swift`. There were three methods to add user observers, one to add an observer for a conforming `UserType`, and one each for adding observers specific to `ZMUser` and `ZMSearchUser`. The latter two methods allowed to observe a single instance or all concrete instances.

Since we can rely on the first method so observe changes in a specific `ZMUser` or `ZMSearchUser` instance, I've removed them. In their place are now two methods to add observers that are interested in changes in all concrete instances, one for each concrete type. In this way, we can avoid exposing `ZMUser` and `ZMSearchUser`.

## Notes

The changes also include minor clean ups and argument label changes.

### DIFF
--- a/Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
+++ b/Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
@@ -180,26 +180,20 @@ extension ZMUser : ObjectInSnapshot {
 
 extension UserChangeInfo {
 
-    // MARK: Registering UserObservers
+    // MARK: Registering UserType
 
-    /// Adds an observer for all ZMUsers in the given context. You must hold on to the token and use it to unregister.
+    /// Adds an observer for the ZMUser or ZMSearchUser. You must hold on to the token and use it to unregister.
     ///
-    public static func add(userObserver observer: ZMUserObserver, in managedObjectContext: NSManagedObjectContext) -> NSObjectProtocol {
-        return add(userObserver: observer, for: nil, managedObjectContext: managedObjectContext)
-    }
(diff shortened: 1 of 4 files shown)


Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
```diff
@@ -180,54 +180,68 @@ extension ZMUser : ObjectInSnapshot {
...
-    public static func add(searchUserObserver observer: ZMUserObserver,
-                           for user: ZMSearchUser?,
-                           managedObjectContext: NSManagedObjectContext
-                           ) -> NSObjectProtocol
-    {
-        return ManagedObjectObserverToken(name: .SearchUserChange, managedObjectContext: managedObjectContext, object: user)
-        { [weak observer] (note) in
-            guard let `observer` = observer,
+
+    /// Adds an observer for all ZMUsers in the given context. You must hold on to the token and use it to unregister.
```
typfel: Should this say: `for all ZMSearchUsers in the given context`?

johnxnguyen: Fixed

johnxnguyen:
@zenkins release major

PR #883, opened 2020-01-28
[Fix] Add the verified property on the UserType.

KaterinaWire:
## What's new in this PR?

Since the rules for displaying the verified shield have changed, we decided to create a new `verified: Bool` property in` UserType`.`verified` property shows whether the user verified own and other devices.

PR also contains the convert of trusted() and untrusted() methods from Objective C to Swift.

### DIFF
--- a/Source/Model/User/UserType.swift
+++ b/Source/Model/User/UserType.swift
@@ -126,6 +126,9 @@ public protocol UserType: NSObjectProtocol {
     /// All clients belonging to the user
     var allClients: [UserClientType] { get }
     
+    /// Whether the user verified all own devices plus others
+    var verified: Bool { get }
+    
     func requestPreviewProfileImage()
     func requestCompleteProfileImage()
     
--- a/Source/Model/User/UserType.swift
+++ b/Source/Model/User/UserType.swift
@@ -127,7 +127,7 @@ public protocol UserType: NSObjectProtocol {
     var allClients: [UserClientType] { get }
     
     /// Whether the user verified all own devices plus others
-    var verified: Bool { get }
+    var isVerified: Bool { get }
     
     func requestPreviewProfileImage()
     func requestCompleteProfileImage()
--- a/Source/Model/User/ZMSearchUser.swift
+++ b/Source/Model/User/ZMSearchUser.swift
@@ -217,7 +217,7 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
     }
     
     public var isVerified: Bool {
-        return user?.isVerified != false
+        return user?.isVerified ?? false
     }
     
     public var managedByWire: Bool {
--- a/Source/Model/User/ZMSearchUser.swift
+++ b/Source/Model/User/ZMSearchUser.swift
@@ -216,6 +216,15 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
         return user?.allClients ?? []
     }
     
+    public var verified: Bool {
+        guard let user = user,
+            let selfUser = contextProvider?.managedObjectContext.map(ZMUser.selfUser) else {
+                return false
+        }
+        
+        return user.trusted() && selfUser.trusted()
+    }
+    
     public var managedByWire: Bool {
         return user?.managedByWire != false
     }
--- a/Source/Model/User/ZMSearchUser.swift
+++ b/Source/Model/User/ZMSearchUser.swift
@@ -217,12 +217,7 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
     }
     
     public var isVerified: Bool {
-        guard let user = user,
-            let selfUser = contextProvider?.managedObjectContext.map(ZMUser.selfUser) else {
-                return false
-        }
-        
-        return user.trusted() && selfUser.trusted()
+        return user?.isVerified != false
     }
     
     public var managedByWire: Bool {
--- a/Source/Model/User/ZMSearchUser.swift
+++ b/Source/Model/User/ZMSearchUser.swift
@@ -216,7 +216,7 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
         return user?.allClients ?? []
     }
     
-    public var verified: Bool {
+    public var isVerified: Bool {
         guard let user = user,
             let selfUser = contextProvider?.managedObjectContext.map(ZMUser.selfUser) else {
                 return false
--- a/Source/Model/User/ZMUser.swift
+++ b/Source/Model/User/ZMUser.swift
@@ -327,7 +327,7 @@ extension ZMUser {
     
     @objc
     public func trusted() -> Bool {
-        if self.clients.count == 0 {
+        if self.clients.isEmpty {
             return false
         }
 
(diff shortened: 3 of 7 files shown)


Source/Model/User/ZMUser.swift
```diff
@@ -315,3 +322,31 @@ extension NSManagedObject: SafeForLoggingStringConvertible {
         return "\(type(of: self)) \(Unmanaged.passUnretained(self).toOpaque()): moc=\(moc) objectID=\(self.objectID)"
     }
 }
+
+extension ZMUser {
+    
+    @objc
+    public func trusted() -> Bool {
+        if self.clients.count == 0 {
```
billypchan:
```suggestion
        if self.clients.isEmpty {
```

Source/Model/User/UserType.swift
```diff
@@ -126,6 +126,9 @@ public protocol UserType: NSObjectProtocol {
     /// All clients belonging to the user
     var allClients: [UserClientType] { get }
     
+    /// Whether the user verified all own devices plus others
+    var verified: Bool { get }
```
johnxnguyen: I'd suggest to name this `isVerified`, to be consistent with the other `"is" properties.

Source/Model/User/ZMUser.swift
```diff
@@ -315,3 +322,31 @@ extension NSManagedObject: SafeForLoggingStringConvertible {
...
+            return false
+        }
+
+        let selfUser = managedObjectContext.map(ZMUser.selfUser)
+        let selfClient = selfUser?.selfClient()
+        let untrustedClient = self.clients.first(where: { ($0 != selfClient) && !(selfClient?.trustedClients.contains($0) ?? false) })
+        let hasOnlyTrustedClients = (untrustedClient == nil)
```
billypchan: We can `contains(where:)` to do these in one line

Source/Model/User/ZMSearchUser.swift
```diff
@@ -216,6 +216,15 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
         return user?.allClients ?? []
     }
     
+    public var verified: Bool {
+        guard let user = user,
```
typfel: A search user can't be verified since it doesn't have any clients. It enough to delegate to the optional user, look how it's done in `var managedByWire: Bool`.

`return user?.verified == true`

Source/Model/User/ZMUser.swift
```diff
@@ -315,3 +322,31 @@ extension NSManagedObject: SafeForLoggingStringConvertible {
...
+        
+        return hasOnlyTrustedClients
+    }
+    
+    @objc
+    public func untrusted() -> Bool {
```
billypchan: we can simply return `!trusted()`

Source/Model/User/ZMUser.swift
```diff
@@ -315,3 +322,31 @@ extension NSManagedObject: SafeForLoggingStringConvertible {
         return "\(type(of: self)) \(Unmanaged.passUnretained(self).toOpaque()): moc=\(moc) objectID=\(self.objectID)"
     }
 }
+
+extension ZMUser {
+    
+    @objc
+    public func trusted() -> Bool {
```
billypchan:
```suggestion
    public var trusted: Bool {
```

johnxnguyen: `isTrusted` is a more appropriate name for a boolean property.

typfel: Be aware that changing this to a `var` is a breaking change.

KaterinaWire: trusted() and untrusted()  methods were converted from Objective C code. If I change the name and make it a property, this will require refactoring in the data model and UI project.What do you think is it good to do that in this PR?

Tests/Source/Model/User/ZMUserTests.swift
```diff
@@ -718,3 +718,55 @@ extension ZMUserTests {
...
+// MARK: - Verifying user
+extension ZMUserTests {
+    
+    func testThatUserIsVerified() {
+        // GIVEN
+        let user: ZMUser = self.userWithClients(count: 2, trusted: true)
+        let selfUser = ZMUser.selfUser(in: uiMOC)
+        
+        // WHEN
+        XCTAssert(user.trusted())
```
typfel: Use `XCTAssertTrue` or `XCTAssertFalse` when asserting boolean properties.

Tests/Source/Model/User/ZMUserTests.swift
```diff
@@ -718,3 +718,55 @@ extension ZMUserTests {
     }
 }
 
+// MARK: - Verifying user
+extension ZMUserTests {
+    
+    func testThatUserIsVerified() {
```
typfel: Let's be more specific when naming the tests

```suggestion
    func testThatUserIsVerified_WhenBothSelfUserAndOtherUserIsTrusted() {
```

Tests/Source/Model/User/ZMUserTests.swift
```diff
@@ -718,3 +718,55 @@ extension ZMUserTests {
...
+        
+        // WHEN
+        XCTAssert(user.trusted())
+        XCTAssert(selfUser.trusted())
+        
+        // THEN
+        XCTAssert(user.verified)
+    }
+    
+    func testThatUserIsNotVerified() {
```
typfel:
```suggestion
    func testThatUserIsNotVerified_WhenSelfUserIsNotTrusted() {
```

Source/Model/User/ZMUser.swift
typfel: I suggest we do it separately to keep the size of this PR down.

Tests/Source/Model/User/ZMUserTests.swift
KaterinaWire: Fixed

Tests/Source/Model/User/ZMUserTests.swift
KaterinaWire: Fixed

Source/Model/User/ZMUser.swift
KaterinaWire: Fixed

Source/Model/User/UserType.swift
KaterinaWire: Fixed

Source/Model/User/ZMSearchUser.swift
KaterinaWire:
```
public var managedByWire: Bool {
        return user?.managedByWire != false
    }
```
If we have `user == nil`, then our property will be `true`.
Isn't it better to do like?
```
public var isVerified: Bool {
        return user?.isVerified ?? false
    }
```

Source/Model/User/ZMUser.swift
KaterinaWire: I'll create a new PR with refactoring the trusted() method.

Source/Model/User/ZMUser.swift
KaterinaWire: I'll create a new PR with refactoring the trusted() method, and the untrusted() method will be removed.

KaterinaWire:
@zenkins release minor

PR #885, opened 2020-01-29
Fix: crash when calling connectionsInMangedObjectContext

billypchan:
### Issues
App crash when calling `ZMConnection.connections()`

### Causes
Unknown, but objc version `executeFetchRequestOrAssert` may cause a crash when the result is empty.

### Solutions

Refactor the method to Swift and use Swift version `fetchOrAssert` instead.


billypchan:
@zenkins release patch

PR #890, opened 2020-02-04
[Refactor] Updates assets using Swift Protobuf API 

KaterinaWire:
## What's new in this PR?

Refactoring assets update methods to use the Swift Protobuf API.


Source/Model/Message/ZMAssetClientMessage+GenericMessage.swift
```diff
@@ -60,6 +64,17 @@ extension ZMAssetClientMessage {
         return self.cachedGenericAssetMessage
     }
     
+    public var underlyingAssetMessage: GenericMessage? {
```
typfel: We can rename this to `underlyingMessage`, we separated the two in the past for historical reasons.

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -401,3 +416,22 @@ public extension LinkPreview {
         }
     }
 }
+
+extension GenericMessage {
+    
+    public mutating func updatedPreview(withAssetId assetId: String, token: String?) {
+        self.asset.preview.remote.assetID = assetId
```
typfel: Don't we need to handle the ephemeral case here as well?

typfel:
I would run the sync-engine and request-strategy tests with these changes.

KaterinaWire:
> I would run the sync-engine and request-strategy tests with these changes.

There are many failed tests in the sync-engine.

KaterinaWire:
@typfel All tests in SE and request-strategy passed.

typfel:
@KaterinaWire I think we should add some tests for `updatedPreview(withAssetId:token)` and `updatedUploaded(withAssetId:token:)`.

KaterinaWire:
> @KaterinaWire I think we should add some tests for `updatedPreview(withAssetId:token)` and `updatedUploaded(withAssetId:token:)`.

Done

KaterinaWire:
@zenkins release patch

PR #892, opened 2020-02-12
[Fix] The missed calls count doesn't clear.

KaterinaWire:
### Issues
After 3 or more missed calls the missed calls badge doesn't clear.

### Causes
We use `serverTimestamp` for system messages to compare it with `lastReadServerTimeStamp` and mark it as a read messages. If we have, for example, 4 missed messages it takes a `timestamp` from the first missed message.

### Solutions
Use `serverTimestampIncludingChildMessages` property for system messages.


typfel:
Why does issue happen on the third call? Should already start happening on the second missed call?

KaterinaWire:
> Why does the issue happen on the third call? Should already start happening on the second missed call?
@typfel 
I will show it with an example.

`lastReadServerTimeStamp` = 16:10:00
FirstMissedCall: `messageTimestamp` = 16:11:00
SecondMissedCall: `messageTimestamp` = 16:12:00
ThirdMissedCall: `messageTimestamp` = 16:13:00

Steps to reproduce:
**1. We made a first missed call.**
`lastReadServerTimeStamp` = 16:10:00
`messageTimestamp` = 16:11:00

We are checking:
`lastReadServerTimeStamp.compare(messageTimestamp) == .orderedDescending`
`messageTimestamp` is newer than `lastReadServerTimeStamp`, so we change the `lastReadServerTimeStamp`. 

**2. Now `lastReadServerTimeStamp` = 16:11:00**

**3. We made a second missed call.**
`lastReadServerTimeStamp` = 16:11:00
`messageTimestamp` = 16:12:00

We are checking:
**_But in this moment we use the time of a first` child message` (16:11:00)_**
`lastReadServerTimeStamp.compare(messageTimestamp) == .orderedDescending`
In this moment `lastReadServerTimeStamp` and `messageTimestamp` are equal and we are allowed to change the `lastReadServerTimeStamp`.

**4. Now `lastReadServerTimeStamp` = 16:12:00**

**5. We made a third missed call.**
`lastReadServerTimeStamp` = 16:12:00
`messageTimestamp` = 16:13:00

We are checking:
**_But in this moment we use the time of a first` child message` (16:11:00)_**
`lastReadServerTimeStamp.compare(messageTimestamp) == .orderedDescending`
In this moment `lastReadServerTimeStamp` is newer than `messageTimestamp` and we do not change  `lastReadServerTimeStamp`.

Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -235,7 +235,7 @@ extension ZMConversation {
     @objc(markMessagesAsReadUntil:)
     public func markMessagesAsRead(until message: ZMConversationMessage) {
         if let currentTimestamp = lastReadServerTimeStamp,
-           let messageTimestamp = message.serverTimestamp,
+           let messageTimestamp = message.serverTimestampIncludingChildMessages,
```
typfel: Looks like we can remove the second unwrapping of `message.serverTimestampIncludingChildMessages` on line 248 now.

KaterinaWire: Why? 
We compare currentTimestamp and messageTimestamp on the 239 line. `message.serverTimestampIncludingChildMessages` is an optional.

KaterinaWire: Yes, I agree.
I put in the separate guard let statement.

KaterinaWire:
@zenkins release patch

PR #895, opened 2020-02-17
[Refactor] Support sending calls with GenericMessage

David-Henner:
## What's new in this PR?

- Add support for sending calls to GenericMessage
- Additional interfaces to GenericMessage which will be needed in the future


David-Henner:
@zenkins release major

PR #896, opened 2020-02-18
[Refactor] Set message flags using Swift Protobuf API

KaterinaWire:
## What's new in this PR?

Refactor the setting message flags (legal hold status, needs read confirmation) methods to use the Swift Protobuf API.


Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -22,12 +22,21 @@ import WireProtos
 public protocol MessageCapable {
     func setContent(on message: inout GenericMessage)
     var expectsReadConfirmation: Bool { get set }
+    var legalHoldStatus: LegalHoldStatus { get }
+    mutating func updateLegalHoldStatus(_ status: LegalHoldStatus)
```
typfel: Do we need have these methods on the `MessageCapable` protocol in practice or are we always updating the legal hold status on the  `GenericMessage`?

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -148,6 +157,56 @@ extension Ephemeral: MessageCapable {
...
+            }
+        }
+    }
+    
+    public mutating func updateLegalHoldStatus(_ status: LegalHoldStatus) {
+        guard let content = content else { return }
+        switch content {
+        case .text:
+            self.text.legalHoldStatus = status
+        case .image:
```
typfel: Why include the `.image` case when we are not doing anything?

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -148,6 +157,56 @@ extension Ephemeral: MessageCapable {
...
+                return value.legalHoldStatus
+            case let .location(value):
+                return value.legalHoldStatus
+            }
+        }
+    }
+    
+    public mutating func updateLegalHoldStatus(_ status: LegalHoldStatus) {
+        guard let content = content else { return }
+        switch content {
```
typfel: We are missing the `.reaction` case.

typfel: This is ephemeral content so it shouldn't be here.

Source/Utilis/Protos/GenericMessage+Helper.swift
typfel: Again this is ephemeral content so the switch is simply including all cases.

Source/Utilis/Protos/GenericMessage+Helper.swift
KaterinaWire: We don't need it. I removed it.

Tests/Source/Model/Utils/ProtobufUtilitiesTests.swift
```diff
@@ -346,3 +348,86 @@ extension ProtobufUtilitiesTests {
         XCTAssertEqual(sut.ephemeral.asset.preview.remote.sha256, sha)
      }
 }
+
+
+// MARK:-  Using Swift protobuf API, set message flags (legal hold status, needs read confirmation)
+
+extension ProtobufUtilitiesTests {
+
+    func testThatMessageNeedsReadConfirmation_InAOneToOne_WhenSelfUserHasReadReceiptsEnabled() {
```
typfel: I think this is already tested in `ZMMessageTests+Confirmation.swift`

KaterinaWire: Yes, but now we have new `setExpectsReadConfirmation` method in `GenericMessage` struct. That's why I added these tests to check new method.

typfel: I'd like to see a unit test, testing that all expected content types can update the legal hold status. This test doesn't need to involve any `ZMConversation` or other entity objects.

Tests/Source/Model/Messages/ZMMessageTests+LegalHoldStatus.swift
```diff
@@ -0,0 +1,144 @@
...
+import WireTesting
+
+@testable import WireDataModel
+
+class ZMMessageTests_LegalHoldStatus: BaseZMClientMessageTests {
+
+    func testThatItUpdatesLegalHoldStatusFlagForTextMessage() {
+        
+        // given
+        let message = ZMClientMessage(nonce: UUID.create(), managedObjectContext: uiMOC)
```
typfel: Looks like the test would still pass if we delete the `let message`. If that's true let's delete it and rename this test class to be a test on `GenericMessage`

Source/Model/Message/ZMOTRMessage+Confirmations.swift
```diff
@@ -53,4 +53,33 @@ extension ZMOTRMessage {
         return daysElapsed <= ZMOTRMessage.deliveryConfirmationDayThreshold
     }
     
+    func needsReadConfirmation(_ genericMessage: GenericMessage) -> Bool {
```
typfel: I'd like that we delete the old `needsReadConfirmation` property so that we are sure that we are calling the implementation in the tests.

KaterinaWire: Do you mean to delete the `needsReadConfirmation` property at all or delete its implementation in the `ZMOTRMessage` extension?

typfel: The old implementation in `ZMOTRMessage+Confirmations.swift` which uses the Objective-C protobuf methods.

Tests/Source/Model/Messages/ZMMessageTests+LegalHoldStatus.swift
KaterinaWire: Fixed

Source/Model/Message/ZMOTRMessage+Confirmations.swift
KaterinaWire: Done

typfel:
@KaterinaWire there will also come a PR for the request-strategy?

KaterinaWire:
@zenkins release major

PR #898, opened 2020-02-24
[Feature] Add (restore) initials property to ZMUser

KaterinaWire:
## What's new in this PR?

Add (restore) the `initials` property to the ZMUser since we need it in the UI project.

### DIFF
--- a/Source/Model/User/ZMUser.swift
+++ b/Source/Model/User/ZMUser.swift
@@ -333,16 +333,7 @@ extension ZMUser {
     
     /// The initials e.g. "JS" for "John Smith"
     @objc public var initials: String? {
-        return personName(for: self).initials
-    }
-    
-    private func personName(for user: ZMUser) -> PersonName {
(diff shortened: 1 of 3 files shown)


Source/Model/User/ZMUser.swift
```diff
@@ -328,3 +328,21 @@ extension ZMUser {
...
+extension ZMUser {
+    
+    /// The initials e.g. "JS" for "John Smith"
+    @objc public var initials: String? {
+        return personName(for: self).initials
+    }
+    
+    private func personName(for user: ZMUser) -> PersonName {
+        if user.objectID.isTemporaryID {
+            try! managedObjectContext!.obtainPermanentIDs(for: [user])
```
typfel: Why do we need a permanent ID when we only access the name?

KaterinaWire: We don't need it in this case

KaterinaWire:
@zenkins release minor

PR #901, opened 2020-03-02
[Fix] The voice message problem

KaterinaWire:
### Issues
A voice message from the android device can not be played back.

### Solutions
Check if we have a file extension for the audio file, and add it if we did not receive it.


Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
```diff
@@ -150,6 +159,15 @@ extension ZMAssetClientMessage: ZMFileMessageData {
         return temporaryFileURL
     }
     
+    private var fileExtension: String? {
```
billypchan: I found there is a similar code in `FilePreviewGenerator.swift` (`URL.UTI()`), may be we can reuse it.

KaterinaWire: But this code we have in UI project and I need to find a file extension based on mimeType

Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
```diff
@@ -132,9 +133,17 @@ extension ZMAssetClientMessage: ZMFileMessageData {
...
+            let filename = filename,
+            let temporaryDirectoryURL = temporaryDirectoryURL else { return nil }
         
-        let temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(filename)
+        var temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(filename)
+        
+        if let fileExtension = fileExtension,
+            richAssetType == .audio,
+            temporaryFileURL.pathExtension != fileExtension {
+            temporaryFileURL = temporaryFileURL.appendingPathExtension(fileExtension)
```
typfel: You can use the mutating version `appendPathExtension()` then you need to re-assign `temporaryFileURL`.

KaterinaWire: Fixed

KaterinaWire:
@zenkins release patch

PR #902, opened 2020-03-02
[Feature] Introduce new types for composite messages

David-Henner:
## What's new in this PR?

- Introduced the types needed for the composite messages
- Conformed `ZMMessage` to a new protocol to expose the composite message to the UI


Source/Model/Message/ConversationMessage.swift
```diff
@@ -177,7 +182,7 @@ extension ZMMessage {
 
 // MARK:- Conversation Message protocol implementation
 
-extension ZMMessage : ZMConversationMessage {
+extension ZMMessage : ZMConversationMessage, ConversationCompositeMessage {
```
David-Henner: Using a protocol separate from `ZMConversationMessage` because `CompositeMessageData` cannot be exposed to objective-c

Source/Model/Message/ConversationMessage.swift
```diff
@@ -262,6 +267,10 @@ extension ZMMessage {
         return nil
     }
     
+    public var compositeMessageData: CompositeMessageData? {
+        return nil
```
David-Henner: Will be implemented later on

Source/Model/Message/ZMClientMessage+CompositeMessageData.swift
```diff
@@ -0,0 +1,47 @@
...
+
+public enum ButtonMessageState {
+    case notSelected
+    case selected
+    case confirmed
+}
+
+extension ZMClientMessage: CompositeMessageData {
+    public var items: [CompositeMessageItem] {
+        return []
```
David-Henner: Will be implemented later on

Source/Model/Message/ZMClientMessage+CompositeMessageData.swift
```diff
@@ -0,0 +1,53 @@
...
+    var items: [CompositeMessageItem] { get }
+}
+
+public enum CompositeMessageItem {
+    case text(ZMTextMessageData)
+    case button(ButtonMessageData)
+}
+
+public protocol ButtonMessageData {
+    var id: UUID { get }
```
typfel: Why do we need to expose the id of the button to the UI?

Source/Model/Message/ZMClientMessage+CompositeMessageData.swift
```diff
@@ -0,0 +1,53 @@
...
+    var title: String { get }
+    var state: ButtonMessageState { get }
+    func touchAction()
+}
+
+public enum ButtonMessageState {
+    case notSelected
```
typfel: `unselected`?

Source/Model/Message/ZMClientMessage+CompositeMessageData.swift
David-Henner: good catch, it's unnecessary

Source/Model/Message/ZMClientMessage+CompositeMessageData.swift
David-Henner: updated

PR #903, opened 2020-03-03
[Fix] Improve logging when removing participants from a group

typfel:
## What's new in this PR?

In order to investigate issues with system messages about participants being re-added to group, we will add more debug logging.


typfel:
@zenkins release patch

PR #904, opened 2020-03-04
[Refactor] Add method to determine if user type is sender

johnxnguyen:
### Issues

Currently the only way to check if a particular user is the sender of the message is to do a comparison with the `sender: ZMUser` property of `ZMConversationMessage`. Equation with `sender` therefore requires an instance of `ZMUser`.

In order to remove some usages of `ZMUser` from the UI project, we also need a way to check if a given `UserType` is the sender of a message.

### Solutions

Provide a method on `ZMConversationMessage` that checks if a `UserType` instance is the sender of the message.

### DIFF
--- a/Source/Model/Message/ConversationMessage.swift
+++ b/Source/Model/Message/ConversationMessage.swift
@@ -144,6 +144,17 @@ public protocol ZMConversationMessage : NSObjectProtocol {
     var needsLinkAttachmentsUpdate: Bool { get set }
 }
 
+public extension ZMConversationMessage {
+
+    /// Whether the given user is the sender of the message.
+
+    func isUserSender(_ user: UserType) -> Bool {
+        guard let zmUser = user as? ZMUser else { return false }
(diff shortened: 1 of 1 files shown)


johnxnguyen:
@zenkins release minor

Source/Model/Message/ConversationMessage.swift
```diff
@@ -144,6 +144,17 @@ public protocol ZMConversationMessage : NSObjectProtocol {
     var needsLinkAttachmentsUpdate: Bool { get set }
 }
 
+public extension ZMConversationMessage {
+
+    /// Whether the given user is the sender of the message.
+
+    func isUserSender(_ user: UserType) -> Bool {
+        guard let zmUser = user as? ZMUser else { return false }
+        return zmUser == sender
```
marcomaddy1982: If you wanna check whether two instances of a class point to the same memory you should use the operator ===. :-)

PR #905, opened 2020-03-04
[Fix] Retain cycle in ConversationListObserverProxy

typfel:
### Issues

`ConversationListObserverProxy` creates a retain cycle which prevents the UserSession from deallocating when switching accounts.

### Causes

`ConversationListObserverProxy` holds a strong reference to its observer.

### Solutions

Hold a weak reference to the `observer`


typfel:
@zenkins release major

PR #907, opened 2020-03-05
[Feature] Add `ButtonState` entity to data model

David-Henner:
## What's new in this PR?

- Adds the ButtonState entity to datamodel


Source/Model/ButtonState/ButtonState.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public class ButtonState: ZMManagedObject {
+    @NSManaged public var identifier: UUID?
+    @NSManaged public var stateValue: Int16
```
johnxnguyen: The `stateValue` here is declared non nil, but the data model file states it is `optional`. You should synchronize these two definitions, I'd opt to make it non optional because it makes it easier to use in `NSPredicate` queries.

Source/Model/ButtonState/ButtonState.swift
```diff
@@ -0,0 +1,40 @@
...
+    @NSManaged public var message: ZMMessage?
```
johnxnguyen: Make this `private` because we're exposing it through the strongly typed `state` property instead.

Source/Model/ButtonState/ButtonState.swift
```diff
@@ -0,0 +1,40 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public class ButtonState: ZMManagedObject {
+    @NSManaged public var identifier: UUID?
```
johnxnguyen: In the model file, this is marked as transient, meaning that it's not persisted to the database (in memory only). I think you forgot to add the `identifier_data` attribute that is of type Binary Data and is not transient. Take a look at `ZMUser.m` where the UUID is constructed by calling `transientUUIDForKey:`

Source/Model/Message/ZMOTRMessage.h
```diff
@@ -21,13 +21,15 @@
...
 @class MessageUpdateResult;
+@class ButtonState;
 
 NS_ASSUME_NONNULL_BEGIN
 
 extern NSString * const DeliveredKey;
 
 @interface ZMOTRMessage : ZMMessage
 
+@property (nullable, nonatomic, retain) NSSet<ButtonState *> *buttonStates;
```
typfel:
```suggestion
@property (nonatomic, nullable) NSSet<ButtonState *> *buttonStates;
```

- Fix order of annotations
- `retain` is the default and therefore be left out.

Source/Model/ButtonState/ButtonState.swift
johnxnguyen: Basically the idea there is that when we set the (strongly typed, but transient) remote identifier, we transform it to binary data and set the (non transient) remote identifier data property. Getting the transient remote identifier is the reverse process.

David-Henner: Thanks for the explanation! I've updated the implementation in alignment with what you said, and taking example from the `Label` entity

Source/Model/ButtonState/ButtonState.swift
David-Henner: updated in a5ade24

Source/Model/ButtonState/ButtonState.swift
David-Henner: updated in a5ade24

johnxnguyen:
Two more things came to mind:

1. To be consistent with other types, it'd be good to rename `identifier` to `remoteIdentifier` (also the data version too).
2. It'd be nice to define a `ButtonStateType` protocol abstracting over relevant properties, just like `Label` does.

johnxnguyen:
Sorry, I also noticed that `ButtonState` is internal. Did you forget to make it public?

David-Henner:
Regarding  
>`ButtonState` is internal. Did you forget to make it public?

In the current state, `ButtonState` doesn't have to be exposed outside of its module.
It's meant to be used internally and [`ButtonMessageState`](https://github.com/wireapp/wire-ios-data-model/blob/3c29e4ddb78f627eb3536cb7572bf7e2329c1cbc/Source/Model/Message/ZMClientMessage%2BComposite.swift#L47) would represent the states of the button.

However, I think it would make sense to eliminate the redundancy of the button state enum by making the `ButtonState.State` public

> It'd be nice to define a `ButtonStateType` protocol abstracting over relevant properties, just like `Label` does.

Would it still make sense if only the `ButtonState.State` is exposed publicly ?

johnxnguyen:
> In the current state, ButtonState doesn't have to be exposed outside of its module.
It's meant to be used internally in the implementation of ButtonMessageData (which in turn is public) and ButtonMessageState would represent the states of the button.

Great, better not to make it public then 😄

> However, I think it would make sense to eliminate the redundancy of the button state enum by making the ButtonState.State public

You couldn't achieve this without making `ButtonState` public. Which... I think we should avoid.

>> It'd be nice to define a ButtonStateType protocol abstracting over relevant properties, just like Label does.
>
> Would it still make sense if only the ButtonState.State is exposed publicly ?

I don't think it makes much sense.

PR #911, opened 2020-03-10
[Feature] Alarming response - send button action

David-Henner:
## What's new in this PR?

- Implement `ButtonMessageData.touchAction()` to send a `ButtonAction` message in the conversation


Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -280,6 +280,28 @@ public extension NewOtrMessage {
     }
 }
 
+extension ButtonAction: MessageCapable {
+    init(buttonId: String, referenceMessageId: String) {
```
typfel: We can assume that `referenceMessageId` will always be of type `UUID`.

typfel: Similarly is specified what type `buttonId` will have?

Source/Model/Conversation/ZMConversation+Message.swift
```diff
@@ -23,6 +23,12 @@ private let log = ZMSLog(tag: "Conversations")
 @objc 
 extension ZMConversation {
     
+    @discardableResult
+    public func appendButtonAction(with id: String, referenceMessageId: String, nonce: UUID = UUID()) -> ZMConversationMessage? {
```
typfel: I guess this method doesn't need to be public?

typfel: Also here we could be more strict with the types.

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -145,14 +158,45 @@ extension CompositeMessageItemContent: ButtonMessageData {
...
+            !hasSelectedButton else { return }
+
+        moc.performGroupedBlock { [weak self] in
+            guard let `self` = self else { return }
+            var buttonState: ButtonState
+            if let state = self.buttonState {
+                buttonState = state
+            }
+            else {
+                buttonState = ButtonState.insertNewObject(in: moc)
```
typfel: This constructor lines code be a method on the `ButtonState` entity

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -145,14 +158,45 @@ extension CompositeMessageItemContent: ButtonMessageData {
...
+                buttonState.remoteIdentifier = UUID(uuidString: button.id)
+                buttonState.message = self.parentMessage
+            }
+            buttonState.state = .selected
+
+            // TODO: Figure out how to send to only one recipient
+            self.parentMessage.conversation?.appendButtonAction(with: button.id, referenceMessageId: message.messageID)
```
typfel: `message.messageID` is also available via the `nonce` property directly on the `ZMClientMessage`. That would allow you remove the message variable from this method.

Source/Utilis/Protos/GenericMessage+Helper.swift
David-Henner: we can't safely assume that `buttonId` will be `UUID`, at the moment it's an `Int`

Source/Model/ButtonState/ButtonState.swift
```diff
@@ -23,19 +23,26 @@ class ButtonState: ZMManagedObject {
...
         case selected
         case confirmed
     }
     
-    var remoteIdentifier: UUID? {
+    var remoteIdentifier: String? {
         get {
             guard let data = remoteIdentifier_data else { return nil }
-            return UUID(data: data)
+            return String(data: data, encoding: .utf8)
```
typfel: If identifier of the button is `String` we can get rid of the `remoteIdentifier_data` field.

David-Henner: Ok. Was the reason for having it in the first place because `UUID` isn't supported?

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -48,6 +48,19 @@ public enum ButtonMessageState {
...
+        case .some(.unselected):
+            self = .unselected
+        case .some(.selected):
+            self = .selected
+        case .some(.confirmed):
+            self = .confirmed
+        default:
+            self = .unselected
+        }
+    }
```
johnxnguyen: You could make this a bit more readable by creating an initializer that expects a non optional, then a convenience initializer that accepts an optional.

Source/Model/ButtonState/ButtonState.swift
typfel: UUID support in core data came with iOS 11 or later I think so we can't support use it yet.

Source/Model/Conversation/ZMConversation+Message.swift
```diff
@@ -23,6 +23,12 @@ private let log = ZMSLog(tag: "Conversations")
 @objc 
 extension ZMConversation {
     
+    @discardableResult
+    func appendButtonAction(with id: String, referenceMessageId: UUID, nonce: UUID = UUID()) -> ZMClientMessage? {
```
typfel:
```suggestion
    func append(buttonActionWithId id: String, referenceMessageId: UUID, nonce: UUID = UUID()) -> ZMClientMessage? {
```

I think this more follows the naming convention of the other `append` methods

PR #913, opened 2020-03-12
[Feature] Add errorMessage property to ButtonState

David-Henner:
## What's new in this PR?

- Added `errorMessage` property to `ButtonSate`


typfel:
Why do we represent the error with a `String` and not some structured type like an enum?

PR #915, opened 2020-03-13
[Feature] Receive button action confirmation

David-Henner:
## What's new in this PR?

- Implemented the reception of the button action confirmation
- Added `isExpired` property to `ButtonState` to define if the message timed out 

### Note

Work still in progress. Missing some tests


Tests/Source/Model/Messages/ZMClientMessageTests+Composite.swift
```diff
@@ -114,4 +113,53 @@ class ZMClientMessageTests_Composite: BaseZMClientMessageTests {
...
+        var buttonState1: WireDataModel.ButtonState!
+        var buttonState2: WireDataModel.ButtonState!
+        var buttonState3: WireDataModel.ButtonState!
+        var buttonState4: WireDataModel.ButtonState!
+
+        uiMOC.performAndWait { [uiMOC] in
+            buttonState1 = WireDataModel.ButtonState.insert(with: buttonItem1.button.id, message: message, inContext: uiMOC)
+            buttonState2 = WireDataModel.ButtonState.insert(with: buttonItem2.button.id, message: message, inContext: uiMOC)
+            buttonState3 = WireDataModel.ButtonState.insert(with: buttonItem3.button.id, message: message, inContext: uiMOC)
+            buttonState4 = WireDataModel.ButtonState.insert(with: buttonItem4.button.id, message: message, inContext: uiMOC)
```
billypchan: We can refactor to `[buttonItem1, buttonItem2, ..].map()...` to create an array of `ButtonState`

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -92,6 +93,43 @@ extension ZMClientMessage: ConversationCompositeMessage {
...
+        message?.updateButtonStates(withConfirmation: confirmation)
+    }
+    
+    private func updateButtonStates(withConfirmation confirmation: ZMButtonActionConfirmation) {
+        guard let states = buttonStates, states.contains(where: { $0.remoteIdentifier == confirmation.buttonId }) else {
+            return
+        }
+        managedObjectContext?.performGroupedBlock { [managedObjectContext] in
+            states.confirmButtonState(withId: confirmation.buttonId)
+            managedObjectContext?.saveOrRollback()
```
typfel: No need to save manually here, it will be done after all events have been processed (see https://github.com/wireapp/wire-ios-sync-engine/blob/17cdc73bc699432b94d40716eabf72460349d2e1/Source/Synchronization/ZMSyncStrategy%2BEventProcessing.swift#L69)

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -92,6 +93,43 @@ extension ZMClientMessage: ConversationCompositeMessage {
...
+        let nonce = UUID(uuidString: confirmation.referenceMessageId)
+        let message = ZMClientMessage.fetch(withNonce: nonce, for: conversation, in: moc)
+        message?.updateButtonStates(withConfirmation: confirmation)
+    }
+    
+    private func updateButtonStates(withConfirmation confirmation: ZMButtonActionConfirmation) {
+        guard let states = buttonStates, states.contains(where: { $0.remoteIdentifier == confirmation.buttonId }) else {
+            return
+        }
+        managedObjectContext?.performGroupedBlock { [managedObjectContext] in
```
typfel: I think there's no need to dispatch onto `managedObjectContext` since we are already on the sync moc queue.

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -92,6 +93,43 @@ extension ZMClientMessage: ConversationCompositeMessage {
...
+                                        forConversation conversation: ZMConversation,
+                                        inContext moc: NSManagedObjectContext) {
+        let nonce = UUID(uuidString: buttonAction.referenceMessageId)
+        let message = ZMClientMessage.fetch(withNonce: nonce, for: conversation, in: moc)
+        message?.expireButtonState(withButtonAction: buttonAction)
+    }
+    
+    private func expireButtonState(withButtonAction buttonAction: ZMButtonAction) {
+        let state = buttonStates?.first(where: { $0.remoteIdentifier == buttonAction.buttonId })
+        managedObjectContext?.performGroupedBlock { [managedObjectContext] in
```
typfel: I would also verify here if the `performGroupedBlock` is necessary.

David-Henner: It seems to be called from the main thread

Source/Model/Message/ZMClientMessage+Composite.swift
David-Henner: Correct

Tests/Source/Model/Messages/ZMClientMessageTests+Composite.swift
```diff
@@ -114,4 +113,81 @@ class ZMClientMessageTests_Composite: BaseZMClientMessageTests {
...
+        }
+        
+        // THEN
+        XCTAssertEqual(buttonStates[0].state, WireDataModel.ButtonState.State.confirmed)
+        for buttonState in buttonStates[1...3] {
+            XCTAssertEqual(buttonState.state, WireDataModel.ButtonState.State.unselected)
+        }
+    }
+    
+    func testThatItSetsIsExpiredAndUnselectsButtonState() {
```
typfel: I would include the `when` case in the name, something like:

testThatItSetsIsExpiredAndUnselectsButtonState_WhenButtonActionMessageExpires()`

Tests/Source/Model/Messages/ZMClientMessageTests+Composite.swift
```diff
@@ -114,4 +113,81 @@ class ZMClientMessageTests_Composite: BaseZMClientMessageTests {
         XCTAssertNotNil(buttonState)
         XCTAssertEqual(WireDataModel.ButtonState.State.selected, buttonState?.state)
     }
+    
+    func testThatItUpdatesButtonStatesUponButtonActionConfirmation() {
```
typfel:
```suggestion
    func testThatItUpdatesButtonStates_WhenReceivingButtonActionConfirmation() {
```

David-Henner: done in 061c2f8

David-Henner:
Tests passed locally - will merge and we will run the tests on the feature branch before merging it

PR #916, opened 2020-03-17
Feature/alarming response

billypchan:
## What's new in this PR?

Draft for Feature/alarming response


Source/Model/ButtonState/ButtonState.swift
```diff
@@ -0,0 +1,72 @@
...
+        buttonState.message = message
+        return buttonState
+    }
+    
+    override static func entityName() -> String {
+        return String(describing: ButtonState.self)
+    }
+    
+    override static func isTrackingLocalModifications() -> Bool {
+        return false
```
David-Henner: should this be `true` in order to observe changes to the `ButtonState` ? 
(We need to be notified of changes in order to process the button action confirmation in the UI)

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -206,6 +215,9 @@ extension ZMGenericMessage {
             }
             recipientUsers = recipients
         }
+        else if self.hasButtonAction() {
```
billypchan:
```suggestion
        } else if hasButtonAction() {
```

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -206,6 +215,9 @@ extension ZMGenericMessage {
...
+            recipientUsers = recipientForButtonActionMessage()
+        }
         else if let deletedEphemeral = recipientsForDeletedEphemeral() {
```
billypchan:
```suggestion
        } else if let deletedEphemeral = recipientsForDeletedEphemeral() {
```

Source/Model/ButtonState/ButtonState.swift
typfel: `isTrackingLocalModifications` are not related to the observer system. It controls if changes to properties are tracked in the `modifiedKeys` property, which is used when syncing the object with the backend.

We are not syncing the ButtonState entity to the backend AFAIK so this should be `false`.

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -151,6 +151,15 @@ extension ZMGenericMessage {
     func recipientUsersForMessage(in conversation: ZMConversation, selfUser: ZMUser) -> (users: Set<ZMUser>, strategy: MissingClientsStrategy) {
         let (services, otherUsers) = conversation.localParticipants.categorizeServicesAndUser()
 
+        func recipientForButtonActionMessage() -> Set<ZMUser> {
+            guard self.hasButtonAction(),
```
David-Henner:
```suggestion
            guard hasButtonAction(),
```

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -0,0 +1,244 @@
...
+extension ZMClientMessage: ConversationCompositeMessage {
+    public var compositeMessageData: CompositeMessageData? {
+        guard case .some(.composite) = underlyingMessage?.content else {
+            return nil
+        }
+        return self
+    }
+}
+
+extension ZMClientMessage {
```
typfel: Let's move the button related to code to its own extension?

billypchan:
@zenkins release major

PR #918, opened 2020-03-20
[Fix]Unable to get correct device class on iPad

summic:
## What's new in this PR?

Just a bug fix

### Issues

When the user is logged in, Wire will report client information to the server
But under iPad, the device class will be reported as "phone"

It's should be "tablet"

### Causes

userClient.deviceClass = model.hasSurffix("iPad") ? .tablet : .phone
Surffix should be Prefix.

### Solutions

userClient.deviceClass = model.hasPrefix("iPad") ? .tablet : .phone


zenkins:

typfel:
Thank you @summic for the bug fix!

summic:
> Thank you @summic for the bug fix!

Pleasure is all mine

PR #919, opened 2020-03-23
[Fix] Sending alarming response button action 

David-Henner:
## What's new in this PR?

- Fix button action not sending 
- Create button state if necessary on confirmation

### DIFF
--- a/Source/Model/Message/ZMClientMessage+Composite.swift
+++ b/Source/Model/Message/ZMClientMessage+Composite.swift
@@ -103,10 +103,15 @@ extension ZMClientMessage {
     }
     
     private func updateButtonStates(withConfirmation confirmation: ZMButtonActionConfirmation) {
-        guard let states = buttonStates, states.contains(where: { $0.remoteIdentifier == confirmation.buttonId }) else {
-            return
+        if !containsButtonState(withId: confirmation.buttonId) {
+            guard let moc = managedObjectContext else { return }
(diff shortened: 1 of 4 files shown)


Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -103,10 +103,15 @@ extension ZMClientMessage {
     }
     
     private func updateButtonStates(withConfirmation confirmation: ZMButtonActionConfirmation) {
-        guard let states = buttonStates, states.contains(where: { $0.remoteIdentifier == confirmation.buttonId }) else {
-            return
+        if !containsButtonState(withId: confirmation.buttonId) {
+            guard let moc = managedObjectContext else { return }
```
billypchan: I would prefer put this `guard` at the beginning of the func

Source/Model/Message/ZMClientMessage+Composite.swift
```diff
@@ -103,10 +103,16 @@ extension ZMClientMessage {
     }
     
     private func updateButtonStates(withConfirmation confirmation: ZMButtonActionConfirmation) {
-        guard let states = buttonStates, states.contains(where: { $0.remoteIdentifier == confirmation.buttonId }) else {
-            return
+        guard let moc = managedObjectContext else { return }
+
+        if !containsButtonState(withId: confirmation.buttonId) {
+            ButtonState.insert(with: confirmation.buttonId, message: self, inContext: moc)
```
typfel: A test for this scenario?

PR #920, opened 2020-03-23
[Refactor] Add cache generic message properties to ZMClientMessage

KaterinaWire:
## What's new in this PR?

This PR is part of the ongoing Swift Protobuf refactoring effort.

1. Convert  ZMClientMessage class to Swift.
2. Add cache generic message properties to ZMClientMessage.


Source/Model/Message/ZMClientMessage+GenericMessage.swift
```diff
@@ -0,0 +1,135 @@
...
+
+import Foundation
+
+extension ZMClientMessage {
+    
+    public override var genericMessage: ZMGenericMessage? {
+        return genericClientMessage
+    }
+    
+    public var genericClientMessage: ZMGenericMessage? {
```
typfel: Why do we have this additional property in addition to `var genericMessage`?

Source/Model/Message/ZMClientMessage+GenericMessage.swift
```diff
@@ -0,0 +1,135 @@
...
+            add(message.data())
+            updateNormalizedText()
+        } else {
+            applyLinkPreviewUpdate(message, from: updateEvent)
+        }
+    }
+}
+
+// MARK: Message data
+extension ZMClientMessage: ZMKnockMessageData {
```
typfel: This extension isn't strictly about generic message? Maybe it could be moved to it's own file?

typfel:
The old Objective-C implementation is not deleted.

Source/Model/Message/ZMClientMessage+GenericMessage.swift
KaterinaWire: We don't need it. I removed it.

Source/Model/Message/ZMClientMessage+GenericMessage.swift
KaterinaWire: Yes, I agree. Moved it.

KaterinaWire:
> The old Objective-C implementation is not deleted.

ZMClientMessage.h and ZMClientMessage.m have been removed.

Source/Model/Message/ZMClientMessage.swift
```diff
@@ -0,0 +1,173 @@
...
+    public override func markAsSent() {
+        super.markAsSent()
+        if linkPreviewState == ZMLinkPreviewState.uploaded {
+            linkPreviewState = ZMLinkPreviewState.done
+        }
+        setObfuscationTimerIfNeeded()
+    }
+
+    private func setObfuscationTimerIfNeeded() {
+        guard self.isEphemeral else {
```
typfel:
```suggestion
        guard isEphemeral else {
```

Source/Model/Message/ZMClientMessage.swift
```diff
@@ -0,0 +1,173 @@
...
+            return
+        }
+        if let genericMessage = self.genericMessage,
+            let _ = genericMessage.textData,
```
typfel:
```suggestion
            genericMessage.textData != nil,
```

In my opinion this way it's more clear, that we are checking that the property is not nil.

Source/Model/Message/ZMClientMessage.swift
```diff
@@ -0,0 +1,173 @@
...
+            linkPreviewState != ZMLinkPreviewState.done {
+            // If we have link previews and they are not sent yet, we wait until they are sent
+            return
+        }
+        startDestructionIfNeeded()
+    }
+
+    func hasDownloadedImage() -> Bool {
+        if let textMessageData = self.textMessageData,
+            let _ = textMessageData.linkPreview,
```
typfel: Same as above

Source/Model/Message/ZMClientMessage+Location.swift
```diff
@@ -28,6 +28,24 @@ public protocol LocationMessageData: NSObjectProtocol {
...
+            switch data.content {
+            case .location(_)?:
+                return self
+            default:
+                return nil
+            }
+            
+        default:
+            return nil
+        }
```
David-Henner, line 46:
```suggestion
        guard let content = underlyingMessage?.content else { return nil }
        switch content {
        case .location:
            return self
        case .ephemeral(let data):
            guard case .some(.location) = data.content else { return nil }
            return self
        default:
            return nil
        }
```

Source/Model/Message/ZMClientMessage+TextMessageData.swift
```diff
@@ -29,6 +29,11 @@ extension ZMClientMessage: ZMTextMessageData {
     
     @NSManaged public var quote: ZMMessage?
     
+    public override var textMessageData: ZMTextMessageData? {
+        let isTextMessage = self.genericMessage?.textData != nil
```
David-Henner, line 33: suggestion to improve readability same as above

Source/Model/Message/ZMClientMessage.swift
```diff
@@ -0,0 +1,173 @@
...
+        }
+        super.resend()
+    }
+    
+    public override func update(withPostPayload payload: [AnyHashable : Any], updatedKeys: Set<AnyHashable>?) {
+        // we don't want to update the conversation if the message is a confirmation message
+        guard let genericMessage = self.genericMessage else { return }
+        if genericMessage.hasConfirmation() || genericMessage.hasReaction() {
+            return
+        }
```
David-Henner, line 98: we could group those under the same `guard` statement

Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
```diff
@@ -309,7 +309,7 @@ class DatabaseMigrationTests: DatabaseBaseTest {
     }
     
     func testThatItPerformsMigrationFrom_Between_2_24_1_and_PreLast_ToCurrentModelVersion() {
-        let allVersions = ["2-24-1", "2-25-0", "2-26-0", "2-27-0", "2-28-0", "2-29-0", "2-30-0", "2-31-0", "2-39-0", "2-40-0", "2-41-0", "2-42-0", "2-43-0", "2-44-0", "2-45-0", "2-46-0", "2-47-0", "2-48-0", "2-49-0", "2-50-0", "2-51-0", "2-52-0", "2-53-0", "2-54-0", "2-55-0", "2-56-0", "2-57-0", "2-59-0", "2-60-0", "2-61-0", "2-62-0", "2-63-0", "2-64-0", "2-65-0", "2-66-0", "2-67-0", "2-68-0", "2-69-0", "2-70-0", "2-71-0", "2-72-0", "2-73-0", "2-74-0", "2-75-0", "2-76-0", "2-77-0", "2-78-0", "2-79-0"]
+        let allVersions = ["2-24-1", "2-25-0", "2-26-0", "2-27-0", "2-28-0", "2-29-0", "2-30-0", "2-31-0", "2-39-0", "2-40-0", "2-41-0", "2-42-0", "2-43-0", "2-44-0", "2-45-0", "2-46-0", "2-47-0", "2-48-0", "2-49-0", "2-50-0", "2-51-0", "2-52-0", "2-53-0", "2-54-0", "2-55-0", "2-56-0", "2-57-0", "2-59-0", "2-60-0", "2-61-0", "2-62-0", "2-63-0", "2-64-0", "2-65-0", "2-66-0", "2-67-0", "2-68-0", "2-69-0", "2-70-0", "2-71-0", "2-72-0", "2-73-0", "2-74-0", "2-75-0", "2-76-0", "2-77-0", "2-78-0", "2-79-0", "2-80-0"]
```
David-Henner, line 312: will also need to be updated when getting up to date with `develop`

Source/Model/Message/ZMClientMessage.swift
KaterinaWire: Done

Source/Model/Message/ZMClientMessage+Ephemeral.swift
```diff
@@ -0,0 +1,41 @@
...
+        return self.destructionDate != nil || self.ephemeral != nil || self.isObfuscated
+    }
+    
+    var ephemeral: ZMEphemeral? {
+        let first = self.dataSet.array
+            .compactMap { ($0 as? ZMGenericMessageData)?.genericMessage }
+            .filter { $0.hasEphemeral() }
+            .first
+        return first?.ephemeral
+    }
```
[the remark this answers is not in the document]
KaterinaWire: What do you mean?

Source/Model/Message/ZMClientMessage+GenericMessage.swift
```diff
@@ -0,0 +1,92 @@
...
+        return message
+    }
+    
+    @objc(addData:)
+    public func add(_ data: Data?) {
+        guard let data = data else {
+            return
+        }
+        let messageData = mergeWithExistingData(data)
+
```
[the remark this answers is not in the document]
KaterinaWire: Yes, because now we use cache properties to save values and we don't need an old objc implementation:
```
- (void)setGenericMessage:(ZMGenericMessage *)genericMessage
 {
     if ([genericMessage knownMessage] && genericMessage.imageAssetData == nil) {
         _genericMessage = genericMessage;
     }
 }
```

Source/Model/Message/ZMClientMessage.swift
KaterinaWire: I'm not sure it can improve readability if I put it under the same `guard`, smth like this: 
guard let genericMessage = self.genericMessage,
        (!genericMessage.hasConfirmation() || !genericMessage.hasReaction() )
            else { return }

Source/Model/Message/ZMClientMessage+TextMessageData.swift
KaterinaWire: Done

Source/Model/Message/ZMClientMessage+Location.swift
KaterinaWire: Done

Source/Model/Message/ZMClientMessage+Ephemeral.swift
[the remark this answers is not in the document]
David-Henner: I think that using `self.genericMessage.ephemeral` will have the same effect as the code above

Resources/zmessaging.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>zmessaging2.79.0.xcdatamodel</string>
+	<string>zmessaging2.80.0.xcdatamodel</string>
```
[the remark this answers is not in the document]
KaterinaWire: Done

Tests/Source/ManagedObjectContext/DatabaseMigrationTests.swift
KaterinaWire: Done

Source/Model/Message/ZMClientMessage.swift
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    func hasDownloadedImage() -> Bool {
+        if let textMessageData = self.textMessageData,
+            textMessageData.linkPreview != nil,
+            let managedObjectContext = self.managedObjectContext {
+            return managedObjectContext.zm_fileAssetCache.hasDataOnDisk(self, format: ZMImageFormat.medium, encrypted: false)
+                // processed or downloaded
+                || managedObjectContext.zm_fileAssetCache.hasDataOnDisk(self, format: ZMImageFormat.original, encrypted: false)
+            // original
```
johnxnguyen, line 173: You could clean this up with two variables, `hasMedium` and `hasOriginal`.

Source/Model/Message/ZMClientMessage+Deletion.swift
```diff
@@ -0,0 +1,39 @@
...
+    
+    func deleteContent() {
+        self.cachedGenericMessage = nil
+        self.cachedUnderlyingMessage = nil
+        self.dataSet.map { $0 as! ZMGenericMessageData }.forEach {
+            $0.managedObjectContext?.delete($0)
+        }
+        self.dataSet = NSOrderedSet()
+        self.normalizedText = nil
+        self.quote = nil
```
johnxnguyen, line 31: `self` is not necessary here.

Source/Model/Message/ZMClientMessage+Deletion.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+
+import Foundation
+
+extension ZMClientMessage {
+    
+    func deleteContent() {
+        self.cachedGenericMessage = nil
+        self.cachedUnderlyingMessage = nil
+        self.dataSet.map { $0 as! ZMGenericMessageData }.forEach {
```
johnxnguyen, line 26: Is it certain that `$0` is `ZMGenericMessageData`?

Source/Model/Message/ZMClientMessage+Deletion.swift
```diff
@@ -0,0 +1,39 @@
...
+        }
+        self.dataSet = NSOrderedSet()
+        self.normalizedText = nil
+        self.quote = nil
+    }
+    
+    public override func removeClearingSender(_ clearingSender: Bool) {
+        super.removeClearingSender(clearingSender)
+        
+        deleteContent()
```
johnxnguyen, line 37: In `ZMClientMessage.m`, `deleteContent()` was called before `super.removeClearingSender(clearingSender)`. Is the inversion intentional?

Source/Model/Message/ZMClientMessage+Ephemeral.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+
+import Foundation
+
+extension ZMClientMessage {
+    @objc
+    override public var isEphemeral: Bool {
+        return self.destructionDate != nil
+            || self.ephemeral != nil
+            || self.isObfuscated
```
johnxnguyen, line 26: `self` is not needed here.

Source/Model/Message/ZMClientMessage+GenericMessage.swift
```diff
@@ -0,0 +1,91 @@
...
+        if self.cachedGenericMessage == nil {
+            self.cachedGenericMessage = genericMessageFromDataSet()
+        }
+        return self.cachedGenericMessage
+    }
+    
+    private func genericMessageFromDataSet() -> ZMGenericMessage? {
+        let filteredMessages = self.dataSet
+            .compactMap { ($0 as? ZMGenericMessageData)?.genericMessage }
+            .filter{ $0.knownMessage() && $0.imageAssetData == nil }
```
johnxnguyen, line 35: It's worth making the a lazy chain so as to avoid unnecessary creation of intermediate arrays:

```swift
dataSet.lazy
  .compactMap { ... }
  .filter { ... }
```

Source/Model/Message/ZMClientMessage+GenericMessage.swift
```diff
@@ -0,0 +1,91 @@
...
+            self.cachedUnderlyingMessage = self.underlyingMessageMergedFromDataSet()
+        }
+        return self.cachedUnderlyingMessage
+    }
+    
+    private func underlyingMessageMergedFromDataSet() -> GenericMessage? {
+        let filteredData = self.dataSet
+            .compactMap { ($0 as? ZMGenericMessageData)?.underlyingMessage }
+            .filter { $0.knownMessage && $0.imageAssetData == nil }
+            .compactMap { try? $0.serializedData() }
```
johnxnguyen, line 59: Same here, make this a lazy chain.

Source/Model/Message/ZMClientMessage.swift
KaterinaWire: Done

Source/Model/Message/ZMClientMessage+Deletion.swift
KaterinaWire: Done

Source/Model/Message/ZMClientMessage+Deletion.swift
KaterinaWire: I fixet it

Source/Model/Message/ZMClientMessage+Deletion.swift
KaterinaWire: No

Source/Model/Message/ZMClientMessage+Ephemeral.swift
KaterinaWire: Done

KaterinaWire:
> Two things in general:
> 
> There's quite a few unnecessary uses of `self`, it'd be nice to clean those up.
> 
> There are a few different styles for `guard`. Could you make them consistent? I see these three styles:
> 
> ```swift
> guard let foo = bar else { return }
> 
> guard let foo = bar else {
>   return
> }
> 
> guard let foo = bar
>   else { return }
> ```

Fixed

Source/Model/Message/ZMClientMessage+GenericMessage.swift
KaterinaWire: I can't do that because I need an intermediate array to check if it's empty.

Source/Model/Message/ZMClientMessage+GenericMessage.swift
KaterinaWire: Same here

Source/Model/Message/ZMClientMessage+GenericMessage.swift
johnxnguyen: Lazy sequences are still sequences so you can still check if they're empty. What `lazy` does though is prevent intermediate arrays being created between each transformation.

For example, each time you `map` a sequence, it will create a new array and return it. If you chain `map` three times, you get two intermediate arrays and one final array. If you instead use `lazy`, you only get the sequence at the end of all of the transformations.

KaterinaWire: Done

Source/Model/Message/ZMClientMessage+GenericMessage.swift
KaterinaWire: Done

KaterinaWire:
@zenkins release major

PR #932, opened 2020-03-30
[Feature] Add helper methods to request data refreshes

johnxnguyen:
### Issues

The UI may need to trigger remote data fetches for certain objects. This is typically done by setting flags such as `needsToBeUpdatedFromBackend`, which will automatically schedule downstream requests to the backend.

These properties however are defined on concrete managed objects and are therefore not exposed through the various abstraction protocols, such as `UserType` and `TeamType`.

### Solutions

Extend `UserType` and `TeamType` to include methods to request data refreshes from the backend.

### DIFF
--- a/Source/Model/Conversation/Team.swift
+++ b/Source/Model/Conversation/Team.swift
@@ -30,6 +30,7 @@ public protocol TeamType: class {
     var imageData: Data? { get set }
 
     func requestImage()
+    func refreshMetadata()
 }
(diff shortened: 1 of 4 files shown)


typfel:
This will get merged into `feature/tier1`?

johnxnguyen:
@typfel yes... but I see it's pointing to develop. I'll fix that

PR #934, opened 2020-03-31
[Fix] make alarming messages not silenced

David-Henner:
### Issues

Alarming messages are silenced, but they should override all notifications preferences

### Solutions

Do not silence if message is an alarming message (composite message)

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Mute.swift
+++ b/Source/Model/Conversation/ZMConversation+Mute.swift
@@ -141,8 +141,7 @@ extension ZMConversationMessage {
         }
         
         // We assume that all composite messages are alarming messages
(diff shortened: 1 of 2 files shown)


Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -140,6 +140,12 @@ extension ZMConversationMessage {
             return false
         }
         
+        // We assume that all composite messages are alarming messages
+        let compositeMessage = self as? ConversationCompositeMessage
+        guard compositeMessage?.compositeMessageData == nil else {
```
billypchan: `compositeMessage` is not used below so we can:

```suggestion
        guard (self as? ConversationCompositeMessage)?.compositeMessageData == nil else {
```

PR #935, opened 2020-03-31
[Feature] Propagate the digital signature infos to the UI

marcomaddy1982:
## Dependencies

JIRA:  https://wearezeta.atlassian.net/browse/ZIOS-12764
sync-engine:  https://github.com/wireapp/wire-ios-sync-engine/pull/1199
ui: https://github.com/wireapp/wire-ios/pull/4236

### DIFF
--- a/Source/Model/Message/File/SignatureStatus.swift
+++ b/Source/Model/Message/File/SignatureStatus.swift
@@ -18,13 +18,22 @@
 
 import Foundation
 
+// MARK: - SignatureObserver
+@objc(ZMSignatureObserver)
+public protocol SignatureObserver: NSObjectProtocol {
+    func willReceiveSignatureURL()
+    func didReceiveSignatureURL(_ url: URL)
+    func signatureAvailable(_ signature: Data) //FIX ME: type of the file
+    func signatureInvalid(_ error: Error)
+}
+
 public extension NSNotification.Name {
-    static let didReceiveURLForSigningDocument = Notification.Name("DidReceiveURLForSigningDocument")
-    static let didReceiveDigitalSignature = Notification.Name("DidReceiveDigitalSignature")
-    static let didReceiveInvalidDigitalSignature = Notification.Name("DidReceiveInvalidDigitalSignature")
+    static let didReceiveDigitalSignature = Notification.Name("didReceiveDigitalSignature")
+    static let didReceiveInvalidDigitalSignature = Notification.Name("didReceiveInvalidDigitalSignature")
     static let willSignDocument = Notification.Name("willSignDocument")
 }
 
+// MARK: - SignatureStatus
 public enum PDFSigningState: Int {
     case initial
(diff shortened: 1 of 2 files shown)


Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -18,13 +18,22 @@
 
 import Foundation
 
+// MARK: - SignatureObserver
+@objc(ZMSignatureObserver)
+public protocol SignatureObserver: NSObjectProtocol {
+    func willReceiveSignatureURL()
+    func didReceiveSignatureURL(_ url: URL)
+    func signatureAvailable(_ signature: Data) //FIX ME: type of the file
```
johnxnguyen: Is the fix me still relevant?

Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -35,16 +44,19 @@ public enum PDFSigningState: Int {
 
 public final class SignatureStatus : NSObject {
     
+    // MARK: - Private Property
     private(set) var asset: ZMAsset?
-    private(set) var managedObjectContext: NSManagedObjectContext?
+    private(set) var managedObjectContext: NSManagedObjectContext
 
+    // MARK: - Private Property
```
johnxnguyen: I think you mean public properties.

Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -55,29 +67,72 @@ public final class SignatureStatus : NSObject {
...
+                    case let .consentURLReceived(consentURL):
+                        observer?.didReceiveSignatureURL(consentURL)
+                }
+            }
+        }
+    }
+}
+
+// MARK: - DigitallySignatureNotification
+public enum DigitallySignatureNotificationState {
```
johnxnguyen: Did you mean `DigitalSignatureNotificationState`?

johnxnguyen: Also, perhaps this should be declared within the scope of of the notification class?

Source/Model/Message/File/SignatureStatus.swift
marcomaddy1982: No no. I can remove that fix me.

Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -55,29 +67,77 @@ public final class SignatureStatus : NSObject {
...
+            }
+        }
+    }
+}
+
+// MARK: - DigitalSignatureNotification
+public class DigitalSignatureNotification: NSObject  {
+    
+    // MARK: - DigitalSignatureNotificationState
+    public enum DigitalSignatureNotificationState {
```
typfel: Since it's defined inside `DigitalSignatureNotification` it could be renamed to just `State`

PR #937, opened 2020-03-31
[Fix] expose text data of CompositeMessageItem for notifications

David-Henner:
## What's new in this PR?

- Expose optional `ZMTextMessageData` of `CompositeMessageItem` enum

Note: Sync Engine needs this for notifications


David-Henner:
@zenkins release minor

PR #939, opened 2020-04-01
[Feature] Limit status broadcast to known team members and contacts

johnxnguyen:
### Issues

Broadcasting status messages to the entire team is not feasible for unbounded teams. Therefore we must limit the number of recipients who will receive status updates from a member.

### Solutions

Limit the recipients of a broadcast message to known team members and contacts. Known team members are members of the same team of the self user who participate in at least one conversation with the self user. 

Additionally, we limit the maximum number of recipients. We prioritize team member recipients over contacts. This means we take as many known team members as possible, and include any contacts until we meet the specified maximum.

Finally, the return recipients is deterministically calculated. This is means that for any given state of the database, repeated calculations for the recipients will always return the same set. This is achieved by sorting candidate recipients by their remote identifier before filling up the recipient list.

### DIFF
--- a/Tests/Source/Model/User/ZMUserTests.swift
+++ b/Tests/Source/Model/User/ZMUserTests.swift
@@ -304,8 +304,9 @@ extension ZMUserTests {
 }
 
 extension ZMUser {
-    static func insert(in moc: NSManagedObjectContext, name: String, handle: String? = nil, connectionStatus: ZMConnectionStatus = .accepted) -> ZMUser {
+    static func insert(in moc: NSManagedObjectContext, id: UUID = .create(), name: String, handle: String? = nil, connectionStatus: ZMConnectionStatus = .accepted) -> ZMUser {
         let user = ZMUser.insertNewObject(in: moc)
+        user.remoteIdentifier = id
         user.name = name
         user.setHandle(handle)
         let connection = ZMConnection.insertNewSentConnection(to: user)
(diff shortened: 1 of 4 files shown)


Tests/Source/Model/User/ZMUserTests.swift
```diff
@@ -526,6 +495,211 @@ extension ZMUserTests {
...
+        let selfTeamConversation2 = createConversation(in: uiMOC)
+        selfTeamConversation2.addParticipantsAndUpdateConversationState(users: Set([selfUser, selfTeamUser2]), role: nil)
+
+        // self team members
+        let selfTeamConversation3 = createConversation(in: uiMOC)
+        selfTeamConversation3.addParticipantsAndUpdateConversationState(users: Set([selfTeamUser2, selfTeamUser3]), role: nil)
+
+        // self user and other team member
+        let otherTeamConversation = createConversation(in: uiMOC)
+        otherTeamConversation.addParticipantsAndUpdateConversationState(users: Set([selfUser, otherTeamUser]), role: nil)
```
David-Henner, line 553: Suggestion:

```
let selfTeamConversation = createConversation(withUsers: Set([selfUser, selfTeamUser2]), in: uiMOC)

func createConversation(withUsers users: Set<ZMUser>, in moc: NSManagedObjectContext) -> ZMConversation {
   let conversation = createConversation(in: moc)
   conversation.addParticipantsAndUpdateConversationState(users: users, role: nil)
}
```

Tests/Source/Model/User/ZMUserTests.swift
```diff
@@ -526,6 +495,211 @@ extension ZMUserTests {
...
+    }
+
+    func testThatReturnsExpectedRecipientsForBroadcast() {
+        // given
+        _ = ZMUser.insert(in: uiMOC, name: "1", handle: "1", connectionStatus: .pending)
+        _ = ZMUser.insert(in: uiMOC, name: "2", handle: "2", connectionStatus: .blocked)
+        _ = ZMUser.insert(in: uiMOC, name: "3", handle: "3", connectionStatus: .cancelled)
+        _ = ZMUser.insert(in: uiMOC, name: "4", handle: "4", connectionStatus: .ignored)
+        _ = ZMUser.insert(in: uiMOC, name: "5", handle: "5", connectionStatus: .sent)
+        _ = ZMUser.insert(in: uiMOC, name: "6", handle: "6", connectionStatus: .invalid)
```
David-Henner, line 569: Suggestion:

```
// Replace `ConnectionStatus` by correct type
@discardableResult func insertUsers(with dictionary: [String: ConnectionStatus], moc: NSManagedObjectContext) -> [ZMUser] {
   var users = [ZMUser]()
   for (key, value) in dictionary {
      users.append(ZMUser.insert(in: moc, name: key, handle: key, connectionStatus: value))
   }
   return users
}
```

johnxnguyen: I think it is not so necessary, since we would perhaps only use this helper method twice and it is readable as it is. I did make the return type discardable though.

PR #940, opened 2020-04-02
[Feature]: Digital Signature: Retrieve Signature

marcomaddy1982:
## Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-12765
UI: https://github.com/wireapp/wire-ios/pull/4246
sync-engine: https://github.com/wireapp/wire-ios-sync-engine/pull/1202

### DIFF
--- a/Source/Model/Message/File/SignatureStatus.swift
+++ b/Source/Model/Message/File/SignatureStatus.swift
@@ -29,7 +29,7 @@ public protocol SignatureObserver: NSObjectProtocol {
 
 public extension NSNotification.Name {
     static let willSignDocument = Notification.Name("willSignDocument")
-    static let willRetriveSignature = Notification.Name("willRetriveSignature")
+    static let willRetrieveSignature = Notification.Name("willRetrieveSignature")
 }
 
 // MARK: - SignatureStatus
@@ -78,9 +78,9 @@ public final class SignatureStatus : NSObject {
             .post(in: managedObjectContext.notificationContext)
     }
     
-    public func retriveSignature() {
+    public func retrieveSignature() {
         state = .waitingForSignature
-        NotificationCenter.default.post(name: .willRetriveSignature, object: nil)
+        NotificationCenter.default.post(name: .willRetrieveSignature, object: nil)
     }
     
     public func didReceiveConsentURL(_ url: URL) {
@@ -105,6 +105,10 @@ public final class SignatureStatus : NSObject {
         state = .signatureInvalid
         DigitalSignatureNotification(state: .signatureInvalid)
             .post(in: managedObjectContext.notificationContext)
+    }
+    
+    public func store() {
+        managedObjectContext.signatureStatus = self
     }
     
     // MARK: - Observable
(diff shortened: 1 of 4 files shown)


Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -24,19 +24,19 @@ public protocol SignatureObserver: NSObjectProtocol {
...
     func signatureAvailable(_ signature: Data)
-    func signatureInvalid(_ error: Error)
+    func didFailSignature()
 }
 
 public extension NSNotification.Name {
-    static let didReceiveDigitalSignature = Notification.Name("didReceiveDigitalSignature")
-    static let didReceiveInvalidDigitalSignature = Notification.Name("didReceiveInvalidDigitalSignature")
     static let willSignDocument = Notification.Name("willSignDocument")
+    static let willRetriveSignature = Notification.Name("willRetriveSignature")
```
johnxnguyen, line 32: Typo: "retrieve"

Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -160,6 +160,8 @@ import Foundation
             self.didChangeValue(forKey: key)
         }
     }
+    
+    public var signatureStatus: SignatureStatus?
```
typfel, line 164: Storing values on a `NSManagedObject` is problematic since they can get faulted and thus disappear from memory. For persistent object like a status we usually store them in the userInfo of the managed object context. Example: https://github.com/wireapp/wire-ios-sync-engine/blob/develop/Source/Data%20Model/NSManagedObjectContext+CTCallCenter.swift

Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -141,3 +162,16 @@ public class DigitalSignatureNotification: NSObject  {
                               userInfo: [DigitalSignatureNotification.userInfoKey: self]).post()
     }
 }
+
+// MARK: - NSManagedObjectContext
+private let SignatureStatusKey = "SignatureStatus"
```
johnxnguyen, line 167: I suggest to scope this inside the extension.

Source/Model/Message/ZMAssetClientMessage.swift
marcomaddy1982: Done

Source/Model/Message/File/SignatureStatus.swift
marcomaddy1982: Done

Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -141,3 +162,17 @@ public class DigitalSignatureNotification: NSObject  {
...
     }
 }
+
+// MARK: - NSManagedObjectContext
+extension NSManagedObjectContext {
+    private static let signatureStatusKey = "SignatureStatus"
+    
+    @objc public var signatureStatus: SignatureStatus? {
+        get {
+            return self.userInfo[NSManagedObjectContext.signatureStatusKey] as? SignatureStatus
```
typfel, line 172: We should make sure that we only access the status from one context to avoid threading issues:

`precondition(zm_isSyncContext, "signatureStatus can only be accessed on the sync context")`

I think it make sense to limit access to the sync context since that's where the strategy is running. This means that in `retrievePDFSignature()` and `signPDFDocument(observer:)` we need to switch to sync context like this:

```
syncContext.performGroupedBlock {
     syncContext.signatureStatus.retrieveSignature()
}
```

PR #943, opened 2020-04-03
[Refactor] Add support of External for GenericMessage

David-Henner:
## What's new in this PR?

Add support to init `GenericMessage` with `ZMUpdateEvent` and `External`

### DIFF
--- a/Source/Model/Message/GenericMessage+External.swift
+++ b/Source/Model/Message/GenericMessage+External.swift
@@ -1,0 +1,52 @@
+//
+// Wire
+// Copyright (C) 2020 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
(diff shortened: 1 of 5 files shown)


typfel:
This adds External support for `GenericMessage` but doesn't use it's not used anywhere yet. It would great if we could remove the old `ZMGenericMessage` implementation in the same PR.

David-Henner:
@typfel I would rather do that in a separate PR to keep the PRs light.
The aim of this PR is to add `External` support for `GenericMessage`.
There will be another PR adding `ZMUpdateEvent` support.
They're both needed to help replace `ZMGenericMessage` by `GenericMessage` in `ZMOTRMessage` and its subclasses.

typfel:
> @typfel I would rather do that in a separate PR to keep the PRs light.
> The aim of this PR is to add `External` support for `GenericMessage`.
> There will be another PR adding `ZMUpdateEvent` support.
> They're both needed to help replace `ZMGenericMessage` by `GenericMessage` in `ZMOTRMessage` and its subclasses.

Ok, let's just keep in mind that in the PR which starts using this new code, we should also remove the old implementation and update tests, otherwise we have little guarantees that it still works.

PR #946, opened 2020-04-03
[Feature] Support for team users in ZMSearchUser

typfel:
## What's new in this PR?

- Directory results will now return team users, `ZMSearchUser` has therefore been updated to represent a team user.
- Introduce the `materialize(in:)` method which transform a sequence of ZMSearchUser (UserType) into concrete ZMUser instances. 

We can now search for team members which doesn't exist in our local database, these users will represented by a `ZMSearchUser` until the user decides to perform an action with them (add to conversation, create conversation). At that point we will transform them into concrete `ZMUser` and insert then into the local database using the `materialize(in:)` method.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Creation.swift
+++ b/Source/Model/Conversation/ZMConversation+Creation.swift
@@ -32,7 +32,7 @@ extension ZMConversation {
 
     @objc
     static public func insertGroupConversation(session: ZMManagedObjectContextProvider,
-                                               participants: [ZMUser],
+                                               participants: [UserType],
                                                name: String? = nil,
                                                team: Team? = nil,
                                                allowGuests: Bool = true,
@@ -40,7 +40,7 @@ extension ZMConversation {
                                                participantsRole: Role? = nil) -> ZMConversation?
     {
         return self.insertGroupConversation(moc: session.managedObjectContext!,
-                                            participants: participants,
+                                            participants: participants.materialize(in: session.managedObjectContext),
                                             name: name,
                                             team: team,
                                             allowGuests: allowGuests,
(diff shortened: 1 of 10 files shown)


Source/Model/User/UserType+Materialize.swift
```diff
@@ -0,0 +1,62 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+fileprivate extension UserType {
+    
+    func materialize(in context: NSManagedObjectContext) -> ZMUser? {
```
johnxnguyen, line 23: This is an excellent name!

typfel: "(of a ghost, spirit, or similar entity) appear in bodily form."

Source/Model/Conversation/ZMConversation+Creation.swift
```diff
@@ -23,7 +23,7 @@ extension ZMConversation {
     
     @objc(insertGroupConversationIntoManagedObjectContext:withParticipants:)
     static public func insertGroupConversation(moc: NSManagedObjectContext,
-                                               participants: [UserType]) -> ZMConversation?
+                                               participants: [ZMUser]) -> ZMConversation?
```
johnxnguyen, line 26: Why can't we pass in `[UserType]`?

typfel: My reasoning is that we will use `UserType` in the methods exposed to the UI i.e. methods that take a `session` argument. Methods which takes a `ManagedObjectContext` argument I consider internal even though they are currently marked as public.

I'd like to do this distinction in order to only materialize users at the edge between UI / data model, which allows us to work with ZMUser objects internally in SE and data model.

As a follow up re-factoring task I'd like to make these `insertGroupConversation` methods private or make the UI use the https://github.com/wireapp/wire-ios-data-model/blob/develop/Source/Model/Conversation/ConversationCreationOptions.swift API which is unused to day.

johnxnguyen: Yep I agree. I was just confused because the methods are public.

PR #950, opened 2020-04-08
[Refactor] support `GenericMessage` init from `ZMUpdateEvent`

David-Henner:
## What's new in this PR?

- Add support to init `GenericMessage` from `ZMUpdateEvent` 

## Notes

- Will remove objc protobuf implementation in separate PR

### DIFF
--- a/Source/Model/Message/GenericMessage+UpdateEvent.swift
+++ b/Source/Model/Message/GenericMessage+UpdateEvent.swift
@@ -35,11 +35,11 @@ extension GenericMessage {
         
         var message = GenericMessage(withBase64String: base64Content)
         
-        if case .external? = message?.content {
-            message = GenericMessage(from: updateEvent, withExternal: message!.external)
+        if case .some(.external(let external)) = message?.content {
+            message = GenericMessage(from: updateEvent, withExternal: external)
         }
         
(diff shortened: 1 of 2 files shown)


Source/Model/Message/GenericMessage+UpdateEvent.swift
```diff
@@ -0,0 +1,45 @@
...
+            base64Content = updateEvent.payload.string(forKey: "data")
+        case .conversationOtrMessageAdd:
+            base64Content = updateEvent.payload.dictionary(forKey: "data")?.string(forKey: "text")
+        case .conversationOtrAssetAdd:
+            base64Content = updateEvent.payload.dictionary(forKey: "data")?.string(forKey: "info")
+        default:
+            return nil
+        }
+        
+        var message = GenericMessage(withBase64String: base64Content)
```
typfel, line 36: Can't we `guard var` the message here, so that we can get rid of the `message!.external` force unwrapping?

David-Henner: A few lines below we assign `message` to an optional `GenericMessage`, so it wouldn't be enough. I removed the force unwrapping here 3057bd0

PR #953, opened 2020-04-09
[Refactor] use `GenericMessage` in `update(with updateEvent:...)`

David-Henner:
## What's new in this PR?

- Use `GenericMessage` in `update(with updateEvent:...)` for `ZMClientMessage` and `ZMAssetClientMessage`
- Update tests


Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -392,6 +392,20 @@ extension Text: EphemeralMessageCapable {
     public func setContent(on message: inout GenericMessage) {
         message.text = self
     }
+    
+    public func updateLinkPreview(from text: Text) -> Text {
+        guard text.linkPreview.count > 0 else {
```
billypchan, line 397:
```suggestion
        guard !text.linkPreview.isEmpty else {
```

Source/Model/Message/ZMAssetClientMessage+UpdateEvent.swift
```diff
@@ -20,9 +20,26 @@ import Foundation
...
+        guard let message = GenericMessage(from: updateEvent) else {
             return
         }
-        update(with: message, updateEvent: updateEvent, initialUpdate: initialUpdate)
+
+        add(message)
+        version = 3 // We assume received assets are V3 since backend no longer supports sending V2 assets.
+        
+        guard case .some(.asset(let assetData)) = message.content, let status = assetData.status else { return }
+
```
KaterinaWire, line 31: In the original code, we check  message.assetData  != nil and, I think, it's more readable.
```
guard let assetData = message.assetData else {
      return
 }
```

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -392,6 +392,20 @@ extension Text: EphemeralMessageCapable {
     public func setContent(on message: inout GenericMessage) {
         message.text = self
     }
+    
+    public func updateLinkPreview(from text: Text) -> Text {
+        guard !text.linkPreview.isEmpty else {
+            return self
+        }
```
KaterinaWire, line 399: In this PR there are different styles for the `guard let` statement. Let's make it consistent with our style guide.

PR #958, opened 2020-04-14
[Hotfix] Read receipts for composite

David-Henner:
## What's new in this PR?

- Force read confirmation for composite messages


David-Henner:
@zenkins release patch

PR #959, opened 2020-04-14
[Fix] AssetId for the digital signature

KaterinaWire:
## What's new in this PR?

Send as a parameter to the BE the assetID instead of the assetID of the preview.


typfel:
Let's add a unit test for this as well?

PR #961, opened 2020-04-16
[Fix] Do not increase the security level for the connection

KaterinaWire:
### Issues
In the UI project the verified shield is shown on the connection request page.

### Causes
We increase the security level for the connection when the user verifies all of its own devices.

### Solutions
Do not increase the security level for the connection.


KaterinaWire:
@zenkins release patch

PR #964, opened 2020-04-20
use new api for report missing

billypchan:
## What's new in this PR?

set `NewOtrMessage.reportMissing` when `otrMessage` is a Conversation message.


Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -238,12 +254,29 @@ extension ZMGenericMessage {
...
         let nativePush = !hasConfirmation() // We do not want to send pushes for delivery receipts
-        let message = NewOtrMessage(withSender: selfClient, nativePush: nativePush, recipients: userEntries, blob: externalData)
+        var message = NewOtrMessage(withSender: selfClient, nativePush: nativePush, recipients: userEntries, blob: externalData)
+        
+        
+        // fill reportMissing if needed
+        if let conversation = conversation,
+           let context = conversation.managedObjectContext {
+        
+        let recipientsAndStrategy = recipientUsersForMessage(in: conversation, selfUser: ZMUser.selfUser(in: context))
```
typfel, line 270: I would not call `recipientUsersForMessage()` from inside this method. The strategy returned is not relevant when broadcasting. The code path we are interested in starts from `encryptedMessagePayloadDataForBroadcast()`.

I would suggest to add an optional argument to this method `strategy: MissingClientsStrategy`, which defaults to `.doNotIgnoreAnyMissingClient`. We can then use it to indicate that we should populate `reportMissing` property.

billypchan: updated PR.

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -238,12 +254,22 @@ extension ZMGenericMessage {
...
         
         let userEntries = self.recipientsWithEncryptedData(selfClient, recipients: recipients, sessionDirectory: sessionDirectory)
         let nativePush = !hasConfirmation() // We do not want to send pushes for delivery receipts
-        let message = NewOtrMessage(withSender: selfClient, nativePush: nativePush, recipients: userEntries, blob: externalData)
+        var message = NewOtrMessage(withSender: selfClient, nativePush: nativePush, recipients: userEntries, blob: externalData)
+        
+        
+        switch missingClientsStrategy {
+        case .ignoreAllMissingClientsNotFromUsers(let users):
+            message.reportMissing = Array(users.map{ $0.userId })
```
typfel, line 268: Since we handle this case here now, we stop handling it in the ClientMessageTranscoder:
 https://github.com/wireapp/wire-ios-request-strategy/blob/3434541ccf2fc54c46a1b32b0ecf722077ee7bc3/Sources/Request%20Strategies/Client%20Message/ClientMessageRequestFactory.swift#L101

billypchan: It will be done in a request strategies PR.

typfel: Cool, please link the to PRs together.

billypchan:
@zenkins release patch

PR #966, opened 2020-04-20
[Refactor] Convert to swift ZMGenericMessageData

KaterinaWire:
## What's new in this PR?

1. Convert  to swift ZMGenericMessageData class.
2. Remove genericMessage property form ZMGenericMessageData.


Source/Model/Message/ZMGenericMessageData.swift
```diff
@@ -18,8 +18,30 @@
 
 import Foundation
 
-extension ZMGenericMessageData {
-    var underlyingMessage: GenericMessage? {
+@objc(ZMGenericMessageData)
+@objcMembers public class ZMGenericMessageData: ZMManagedObject {
+    
+    @objc public static let dataKey = "data"
```
David-Henner, line 24: is the `@objc` necessary since the class is declared with `@objcMembers` attribute ?

Source/Model/Message/ZMOTRMessage+UpdateEvent.swift
```diff
@@ -142,3 +142,12 @@ extension ZMOTRMessage {
...
 }
+
+extension ZMOTRMessage {
+    public var dataSetDebugInformation: String {
+        let debugDescription = dataSet.lazy
+            .compactMap { ($0 as? ZMGenericMessageData)?.underlyingMessage.debugDescription }
+            .joined(separator: "\n")
+        return String.init(format: "<%@>: %@", NSStringFromClass(ZMGenericMessageData.self), debugDescription)
+    }
+}
```
David-Henner, line 153: I'm thinking we could move this code to a dedicated extension

Source/Model/Message/ZMGenericMessageData.swift
KaterinaWire: No, it's not necessary.
Fixed.

Source/Model/Message/ZMOTRMessage+UpdateEvent.swift
KaterinaWire: Done

PR #968, opened 2020-04-21
[Refactor] Implement hashOfContent() for the GenericMessage

KaterinaWire:
## What's new in this PR?

1. Implemented hashOfContent() for the GenericMessage.
2. Removed ZMGenericMessage implementation of hashOfContent().
3. Fixed related tests.


Source/Utilis/Protos/GenericMessage+Hashing.swift
```diff
@@ -18,6 +18,55 @@
...
+extension GenericMessage {
+    
+    func hashOfContent(with timestamp: Date) -> Data? {
+        
+        guard let content = content else {
+            return nil
+        }
+        switch content {
+        case .location(let data):
+            return data.hashWithTimestamp(timestamp: timestamp.timeIntervalSince1970)
```
billypchan, line 34: we can create a helper method takes `data` and `timestamp` as parameters to save some code?

KaterinaWire: I grouped cases, so now I don't need to repeat 
`            return data.hashWithTimestamp(timestamp: timestamp.timeIntervalSince1970)`

PR #972, opened 2020-04-27
[Refactor] Convert encrypted() methods

KaterinaWire:
## What's new in this PR?

Convert encrypted() methods


Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -165,39 +163,45 @@ extension ZMGenericMessage {
...
-
+    
     func recipientUsersForMessage(in conversation: ZMConversation, selfUser: ZMUser) -> (users: Set<ZMUser>, strategy: MissingClientsStrategy) {
         let (services, otherUsers) = conversation.localParticipants.categorizeServicesAndUser()
 
         func recipientForButtonActionMessage() -> Set<ZMUser> {
-            guard hasButtonAction(),
-                let message = ZMMessage.fetch(withNonce: UUID(uuidString: self.buttonAction.referenceMessageId), for: conversation, in: conversation.managedObjectContext!),
+            guard case .buttonAction? = content,
+                let message = ZMMessage.fetch(withNonce: UUID(uuidString: self.buttonAction.referenceMessageID), for: conversation, in: conversation.managedObjectContext!),
```
David-Henner, line 172: I think convention for multi line `guard` is 
```
guard 
    condition(),
    foo == bar
else {
    return
}
```

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -165,39 +163,45 @@ extension ZMGenericMessage {
...
             return Set(arrayLiteral: sender)
         }
         
         func recipientForConfirmationMessage() -> Set<ZMUser>? {
-            guard self.hasConfirmation(), self.confirmation.firstMessageId != nil else { return nil }
-            guard let message = ZMMessage.fetch(withNonce:UUID(uuidString:self.confirmation.firstMessageId), for:conversation, in:conversation.managedObjectContext!) else { return nil }
+            guard case .confirmation? = content,
+                confirmation.firstMessageID != "" else {
+                    return nil
+            }
```
David-Henner, line 183: Same here regarding `guard`

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -165,39 +163,45 @@ extension ZMGenericMessage {
...
+            guard let message = ZMMessage.fetch(withNonce:UUID(uuidString:self.confirmation.firstMessageID), for:conversation, in:conversation.managedObjectContext!) else { return nil }
```
David-Henner, line 184: maybe we should avoid explicit unwrapping

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -165,39 +163,45 @@ extension ZMGenericMessage {
...
         }
-
+        
         func recipientsForDeletedEphemeral() -> Set<ZMUser>? {
-            guard (self.hasDeleted() && conversation.conversationType == .group ) else { return nil }
-            let nonce = UUID(uuidString: self.deleted.messageId)
+            guard case .deleted? = content,
+                conversation.conversationType == .group else {
+                return nil
+            }
```
David-Henner, line 199: Same here for `guard`

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -260,8 +265,12 @@ extension ZMGenericMessage {
                                 externalData: Data?,
                                 sessionDirectory: EncryptionSessionsDirectory) -> NewOtrMessage {
         
-        let userEntries = self.recipientsWithEncryptedData(selfClient, recipients: recipients, sessionDirectory: sessionDirectory)
-        let nativePush = !hasConfirmation() // We do not want to send pushes for delivery receipts
+        let userEntries = recipientsWithEncryptedData(selfClient, recipients: recipients, sessionDirectory: sessionDirectory)
+        var nativePush = true // We do not want to send pushes for delivery receipts
+        if case .confirmation? = content {
```
David-Henner, line 270: You could write a helper method to avoid repeating those `if case` statements.
```
extension GenericMessage {
    func has(content: OneOf_Content?) -> Bool {
        if case content = self.content {
            return true
        }
        return false
    }
}
```

KaterinaWire: I can't, because `OneOf_Content` is enum. But I created a helper method `hasConfirmation() `.

Source/Model/Message/ZMClientMessage+Encryption.swift
KaterinaWire: Fixed

Source/Model/Message/ZMClientMessage+Encryption.swift
KaterinaWire: Fixed

Source/Model/Message/ZMClientMessage+Encryption.swift
KaterinaWire: Fixed

Source/Model/Message/ZMClientMessage+Encryption.swift
KaterinaWire: Fixed

Source/Model/Message/ZMClientMessage+Encryption.swift
David-Henner: Right, it wouldn't work because of the associated value

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -165,39 +163,54 @@ extension ZMGenericMessage {
...
+                hasConfirmation(),
+                confirmation.firstMessageID != "" else {
+                    return nil
+            }
+            guard
+                let managedObjectContext = conversation.managedObjectContext,
+                let message = ZMMessage.fetch(withNonce:UUID(uuidString:self.confirmation.firstMessageID), for:conversation, in:managedObjectContext) else {
+                return nil
+            }
+            guard let sender = message.sender else {
```
typfel, line 191:
```suggestion
            guard
                hasConfirmation(),
                confirmation.firstMessageID != "",
                let managedObjectContext = conversation.managedObjectContext,
                let message = ZMMessage.fetch(withNonce:UUID(uuidString:self.confirmation.firstMessageID), for:conversation, in:managedObjectContext),
               let sender = message.sender else {
```

typfel: Can't we merge all the guard statements?

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -209,47 +222,48 @@ extension ZMGenericMessage {
...
         let hasRestrictions: Bool = {
             if conversation.connectedUser != nil { return recipientUsers.count != 2 }
             return recipientUsers.count != conversation.localParticipants.count
         }()
-
+        
         let strategy : MissingClientsStrategy = hasRestrictions ? .ignoreAllMissingClientsNotFromUsers(users: recipientUsers)
-                                                                : .doNotIgnoreAnyMissingClient
-
+            : .doNotIgnoreAnyMissingClient
```
typfel, line 265: Isn't this indentation less readable?

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -165,39 +163,54 @@ extension ZMGenericMessage {
...
             }
             return Set(arrayLiteral: sender)
         }
         
         func recipientForConfirmationMessage() -> Set<ZMUser>? {
-            guard self.hasConfirmation(), self.confirmation.firstMessageId != nil else { return nil }
-            guard let message = ZMMessage.fetch(withNonce:UUID(uuidString:self.confirmation.firstMessageId), for:conversation, in:conversation.managedObjectContext!) else { return nil }
-            guard let sender = message.sender else { return nil }
+            guard
+                hasConfirmation(),
```
typfel, line 182: Isn't this statement redundant? We already check if the message has a confirmation in the method that calls `recipientForConfirmationMessage()` and the next line `confirmation.firstMessageID != ""` wouldn't pass if the didn't contain a confirmation.

KaterinaWire: Yes, this statement is redundant.

Source/Model/Message/ZMClientMessage+Encryption.swift
KaterinaWire: Fixed

PR #973, opened 2020-04-28
[Refactor] use `GenericMessage` in `createOrUpdate(...)`

David-Henner:
## What's new in this PR?

- Replaced `ZMGenericMessage` by `GenericMessage` in `ZMOTRMessage.createOrUpdate(...)` method
- Converted necessary protobuf objects to Swift protobufs
- Converted some objc methods to Swift to use Swift protobufs
- Updated tests


Source/Model/Message/ZMOTRMessage+UpdateEvent.swift
```diff
@@ -42,37 +48,46 @@ extension ZMOTRMessage {
...
-        } else if message.hasButtonActionConfirmation() {
+        case .buttonActionConfirmation:
             ZMClientMessage.updateButtonStates(withConfirmation: message.buttonActionConfirmation, forConversation: conversation, inContext: moc)
-        } else if message.hasEdited() {
+        case .edited:
             return ZMClientMessage.editMessage(withEdit: message.edited, forConversation: conversation, updateEvent: updateEvent, inContext: moc, prefetchResult: prefetchResult)
-        } else if conversation.shouldAdd(event: updateEvent) && !(message.hasClientAction() || message.hasCalling() || message.hasAvailability()) {
-            guard let nonce = UUID(uuidString: message.messageId) else { return nil }
-            let messageClass: AnyClass = ZMGenericMessage.entityClass(for: message)
+        case .clientAction, .calling, .availability:
```
KaterinaWire, line 80: The original code contains` !(message.hasClientAction()`

Source/Utilis/Protos/GenericMessage+MessageCapable.swift
```diff
@@ -0,0 +1,320 @@
...
+        set {}
+    }
+    
+    var defaultLegalHoldStatus: LegalHoldStatus {
+        return .unknown
+    }
+}
+
+extension GenericMessage.OneOf_Content {
+    var messageCapable: MessageCapable? {
```
KaterinaWire, line 49: Why do we need this extension?

Source/Model/Message/ZMOTRMessage+UpdateEvent.swift
David-Henner: I don't see the issue ?

Source/Utilis/Protos/GenericMessage+MessageCapable.swift
KaterinaWire, line 49: Why do we need this extension?

David-Henner: it's used in `ZMConversation+SecurityLevel` in the func `updateLegalHoldIfNeededWithHint`

David-Henner: we need it to access the `legalHoldStatus`

Source/Model/Message/ZMOTRMessage+UpdateEvent.swift
KaterinaWire: According to the original code, we should not return nil if message.hasCalling() || message.hasAvailability().

Source/Utilis/Protos/GenericMessage+MessageCapable.swift
KaterinaWire: Maybe we can try like this:

```
extension GenericMessage {
    var legalHoldStatus: LegalHoldStatus? {
        guard let content = content else {
            return nil
        }
        
        switch content {
        case .ephemeral:
            return ephemeral.legalHoldStatus
        case .reaction:
            return reaction.legalHoldStatus
        case .knock:
            return knock.legalHoldStatus
        case .text:
            return text.legalHoldStatus
        case .location:
            return location.legalHoldStatus
        case .asset:
            return asset.legalHoldStatus
        default:
            return nil
        }
    }
}
```

What do you think?

Source/Model/Message/ZMOTRMessage+UpdateEvent.swift
David-Henner: let's take a look a the original code

```
...
else if conversation.shouldAdd(event: updateEvent) && !(message.hasClientAction() || message.hasCalling() || message.hasAvailability()) {
   ...
   return clientMessage
}
return nil
```

For the sake of the conversation, let's assume `conversation.shouldAdd(event: updateEvent)` is always true
We end up with the following statement
```
else if !(message.hasClientAction() || message.hasCalling() || message.hasAvailability())
```

Let's say `message.hasClientAction()` returns `true` and the rest is `false`

```
else if !(true || false || false)

// or 

else if !true
```

In this case, the condition evaluates to `false` and we `return nil`
We can then say that if any of the conditions (`message.has...`) are true, we should return nil
That can translate to `case .clientAction, .calling, .availability: return nil`

Source/Utilis/Protos/GenericMessage+MessageCapable.swift
David-Henner: Yep, much better.
I updated it in ee2cb4a.
It seems that we don't need `legalHoldStatus` in `MessageCapable` protocol, so I removed it for simplicity

PR #975, opened 2020-04-29
[Fix] Don't update name for deleted users

johnxnguyen:
### Issues

When a user is deleted, their name is set to "default".

### Causes

When synchronizing a deleted user with the backend, the backend sets the name field to "default". This is likely to erase metadata for the deleted user. However, when parsing the payload we will set the local user objects name to "default", overwriting their existing username.

### Solutions

Don't update the user name if the user is marked as deleted.


johnxnguyen:
@zenkins release patch

PR #978, opened 2020-05-05
[Refactor] Replace ZMGenericMessage in ZMUser

KaterinaWire:
## What's new in this PR?

Replace `ZMGenericMessage` in `ZMUser` class.


Tests/Source/Model/User/ZMUserTests+Swift.swift
```diff
@@ -454,7 +454,10 @@ extension ZMUserTests_Swift {
         XCTAssertEqual(user.availability, .none)
                 
         // when
-        user.updateAvailability(from: ZMGenericMessage.message(content: ZMAvailability.availability(.away)))
+        let availability = WireProtos.Availability.with {
+            $0.type = .away
+        }
```
typfel, line 459: Move above the `// when` into the `// given` part which doesn't exist.

KaterinaWire: Done

PR #981, opened 2020-05-06
[Chore] Bump components

typfel:
## What's new in this PR?

Bump components so that we get pre-build binaries for latest Swift release.


typfel:
@zenkins release patch

PR #984, opened 2020-05-18
[Fix] Guard against connections referring to the self user

typfel:
### Issues

We have seen clients reporting to have a connection with itself.

### Causes

It's unclear exactly when such connection can be created.

### Solutions

Guard against such connection being created by validating that the `to` field is not referring to the self user.


typfel:
@zenkins release patch

PR #985, opened 2020-05-18
[Fix] Add patch for deleting invalid connection to self user

typfel:
### Issues

We have seen clients reporting to have a connection with itself.

### Causes

Unknown

### Solutions

https://github.com/wireapp/wire-ios-data-model/pull/984 guards against this happening again. This PR adds a data model patch for deleting already existing invalid connections.


typfel:
@zenkins release patch

billypchan:
@zenkins release patch

PR #988, opened 2020-05-22
[Fix] Add more detailed logging when aborting processing of a message

typfel:
## What's new in this PR?

Add more detailed logging when we abort processing of a message because we can't decode the protobuf.


typfel:
@zenkins patch

typfel:
@zenkins release patch

PR #989, opened 2020-05-25
[Refactor] hasXXX methods for GenericMessage

KaterinaWire:
## What's new in this PR?

Reduce duplicate code for hasXXX methods.

### DIFF
--- a/Source/Utilis/Protos/GenericMessage+Content.swift
+++ b/Source/Utilis/Protos/GenericMessage+Content.swift
@@ -22,11 +22,7 @@ import Foundation
 
 public extension GenericMessage {
     var hasText: Bool {
-        guard let _ = messageData as? Text else {
-            let ephemeral = messageData as? Ephemeral
-            return ephemeral?.hasText ?? false
-        }
-        return true
+        return messageData is Text
     }
     
     var hasConfirmation: Bool {
@@ -38,15 +34,7 @@ public extension GenericMessage {
     }
     
     var hasAsset: Bool {
-        guard let _ = messageData as? WireProtos.Asset else {
-            let ephemeral = messageData as? Ephemeral
-            return ephemeral?.hasAsset ?? false
-        }
-        return true
-    }
-    
-    var hasEphemeral: Bool {
-        return messageData is Ephemeral
+        return messageData is WireProtos.Asset
     }
     
     var hasClientAction: Bool {
@@ -62,11 +50,7 @@ public extension GenericMessage {
     }
     
     var hasKnock: Bool {
-        guard let _ = messageData as? Knock else {
-            let ephemeral = messageData as? Ephemeral
-            return ephemeral?.hasKnock ?? false
-        }
-        return true
+        return messageData is Knock
     }
     
     var hasExternal: Bool {
@@ -94,11 +78,7 @@ public extension GenericMessage {
     }
     
     var hasLocation: Bool {
-        guard let _ = messageData as? Location else {
-            let ephemeral = messageData as? Ephemeral
-            return ephemeral?.hasLocation ?? false
-        }
-        return true
+        return messageData is Location
     }
 }
 
--- a/Source/Utilis/Protos/GenericMessage+Content.swift
+++ b/Source/Utilis/Protos/GenericMessage+Content.swift
@@ -30,7 +30,7 @@ public extension GenericMessage {
     }
     
     var hasConfirmation: Bool {
-        return (messageData as? Confirmation) != nil ? true : false
+        return messageData is Confirmation
     }
     
     var hasReaction: Bool {
--- a/Source/Utilis/Protos/GenericMessage+Content.swift
+++ b/Source/Utilis/Protos/GenericMessage+Content.swift
@@ -34,7 +34,7 @@ public extension GenericMessage {
     }
     
     var hasReaction: Bool {
-        return (messageData as? Reaction) != nil ? true : false
+        return messageData is Reaction
     }
     
     var hasAsset: Bool {
@@ -46,19 +46,19 @@ public extension GenericMessage {
     }
     
     var hasEphemeral: Bool {
-        return (messageData as? Ephemeral) != nil ? true : false
+        return messageData is Ephemeral
     }
     
     var hasClientAction: Bool {
-        return (messageData as? ClientAction) != nil ? true : false
+        return messageData is ClientAction
     }
     
     var hasCleared: Bool {
-        return (messageData as? Cleared) != nil ? true : false
+        return messageData is Cleared
     }
     
     var hasLastRead: Bool {
-        return (messageData as? LastRead) != nil ? true : false
+        return messageData is LastRead
     }
     
     var hasKnock: Bool {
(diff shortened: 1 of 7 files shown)


Source/Utilis/Protos/GenericMessage+Content.swift
```diff
@@ -22,192 +22,102 @@ import Foundation
...
     
     var hasConfirmation: Bool {
-        guard let content = content else { return false }
-        switch content {
-        case .confirmation:
-            return true
-        default:
-            return false
-        }
+        return (messageData as? Confirmation) != nil ? true : false
```
typfel, line 33:
```suggestion
        return messageData is Confirmation
```

I guess we can do the same for all other cases as well. Maybe we can even remove the hasXXX methods completely but I guess that will affect a lot of tests so feel free to keep them.

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -61,6 +61,52 @@ public extension GenericMessage {
...
+        case .text(let data):
+            return data
+        case .confirmation(let data):
+            return data
+        case .reaction(let data):
+            return data
+        case .asset(let data):
+            return data
+        case .ephemeral(let data):
+            return data
```
typfel, line 82: shouldn't we do something like this?

```suggestion
            return data.messageData
```

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -61,6 +61,52 @@ public extension GenericMessage {
 }
 
 extension GenericMessage {
+    public var messageContent: OneOf_Content? {
```
typfel, line 64: What purpose does this property serve?

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -261,26 +323,25 @@ extension ButtonAction {
...
     public init(content: String, mentions: [Mention] = [], linkPreviews: [LinkMetadata] = [], replyingTo: ZMOTRMessage? = nil) {
         self = Text.with {
             $0.content = content
             $0.mentions = mentions.compactMap { WireProtos.Mention($0) }
             $0.linkPreview = linkPreviews.map { WireProtos.LinkPreview($0) }
             
             if let quotedMessage = replyingTo,
-               let quotedMessageNonce = quotedMessage.nonce,
-               let quotedMessageHash = quotedMessage.hashOfContent {
+                let quotedMessageNonce = quotedMessage.nonce,
```
typfel, line 343: Misaligned now

KaterinaWire: Fixed

Source/Utilis/Protos/GenericMessage+Helper.swift
KaterinaWire: We have this check that `content != nil` in many places. So, we don't need to repeat such code all the time 
`guard let content = content else { return nil }`
But these changes will cover many files.

Source/Utilis/Protos/GenericMessage+Content.swift
```diff
@@ -22,192 +22,102 @@ import Foundation
...
-        switch content {
-        case .knock:
-            return true
-        default:
-            return false
+        guard let _ = messageData as? Knock else {
+            let ephemeral = messageData as? Ephemeral
+            return ephemeral?.hasKnock ?? false
         }
+        return true
```
typfel, line 69: Since `messageData` now also unwraps ephemeral content should it be enough to do this:

```suggestion
        return messageData is Knock
```

Source/Utilis/Protos/GenericMessage+Content.swift
```diff
@@ -22,192 +22,102 @@ import Foundation
...
-        switch content {
-        case .hidden:
-            return true
-        default:
-            return false
+        return messageData is MessageHide
+    }
+    
+    var hasLocation: Bool {
+        guard let _ = messageData as? Location else {
```
typfel, line 97: Same suggestion as with the Knock

Source/Utilis/Protos/GenericMessage+Content.swift
```diff
@@ -22,192 +22,102 @@ import Foundation
...
     }
     
     var hasAsset: Bool {
-        guard let content = content else { return false }
-        switch content {
-        case .asset:
-            return true
-        default:
-            return false
+        guard let _ = messageData as? WireProtos.Asset else {
```
typfel, line 41: Same suggestion as with the Knock

Source/Utilis/Protos/GenericMessage+Content.swift
```diff
@@ -22,192 +22,102 @@ import Foundation
 
 public extension GenericMessage {
     var hasText: Bool {
-        guard let content = content else { return false }
-        switch content {
-        case .text:
-            return true
-        default:
-            return false
+        guard let _ = messageData as? Text else {
```
typfel, line 25: Same suggestion as with the Knock

Source/Utilis/Protos/GenericMessage+Content.swift
KaterinaWire: I was thinking that it's not correct to unwrap the ephemeral content. We have a hasEphemeral() method and that's why type of messageData should be as Ephemeral.

typfel: Well then I don't understand why we unwrap the ephemeral in `hasAsset` and `hasText`. The hasXXX methods behave in consistently (sometimes they include ephemeral content and sometimes they don't), which makes then hard to reason about.

I'd say we should stick to one logic for the hasXXX methods.

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -61,6 +61,46 @@ public extension GenericMessage {
...
+            return data
+        case .deleted(let data):
+            return data
+        case .calling(let data):
+            return data
+        case .hidden(let data):
+            return data
+        case .location(let data):
+            return data
+        default:
```
typfel, line 99: Perhaps we should leave out the `default` case so that we get an error when there's a case added.

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -207,6 +247,22 @@ extension Ephemeral {
...
+        switch content {
+        case .text(let data):
+            return data
+        case .asset(let data):
+            return data
+        case .knock(let data):
+            return data
+        case .location(let data):
+            return data
+        default:
```
typfel, line 262: Same thing here about removing the default case.

Source/Utilis/Protos/GenericMessage+Content.swift
KaterinaWire: Done

Source/Utilis/Protos/GenericMessage+Content.swift
KaterinaWire: Done

Source/Utilis/Protos/GenericMessage+Content.swift
KaterinaWire: Done

Source/Utilis/Protos/GenericMessage+Content.swift
KaterinaWire: Done

Source/Utilis/Protos/GenericMessage+Helper.swift
KaterinaWire: Done

Source/Utilis/Protos/GenericMessage+Helper.swift
KaterinaWire: Done

PR #992, opened 2020-06-05
[Fix] Dependency on wire-ios-protos in Cartfile

typfel:
### Issues

Dependency on `wire-ios-protos` as branch named "24.0" which doesn't exist

### Solutions

Fix syntax for declaring the dependency on `wire-ios-protos` of major version 24.0.


typfel:
@KaterinaWire I also had to remove references to `protobuf-objc` in the project.

typfel:
@zenkins release major

PR #993, opened 2020-06-09
[Feature] Targeted messages

johnxnguyen:
## What's new in this PR?

The upcoming conference calling feature requires clients to be able to send targeted encrypted messages to specific client devices. This PR includes changes to make this possible.

### Context

Currently, there are two ways in which an encrypted message can be sent through the backend to various recipients.

1. In a conversation (such as posting a new message), where all participant users should receive the message.
2. In a broadcast (such as changing of availability status), where all known team users / contacts should receive the message.

In both cases, a single request is made to the backend. This request contains an (individually) encrypted payload for each known client of all recipients. When then request is made, either it succeeds and the backend fans out the encrypted payloads to the corresponding devices, or it fails because the request is missing some recipient devices in the payload.

The method of broadcasting a message to a set of users is the closest we have to targeting specific clients, however it requires that **all** clients for each recipient user receives the message. What we want is to specify **exactly** which clients should receive the message. 

## Changes

The methods responsible for generating the encrypted data have been refactored to accept a map of recipients `[ZMUser: Set<UserClient>]` instead of a set of users. This allows us to define a new public method to generated an encrypted payload for specific clients.

There are also general refactoring changes such as reorganization of methods, renaming, and method break downs.


Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -25,59 +25,56 @@ public let ZMFailedToCreateEncryptedMessagePayloadString = "💣"
...
-}
+    typealias Payload = (data: Data, strategy: MissingClientsStrategy)
 
+    /// Returns the payload encrypted for each recipients and the strategy to use to handle missing clients.
 
-public protocol EncryptedPayloadGenerator {
-    /// Returns the payload encrypted for each recipients, and the strategy
-    /// to use to handle missing clients
-    func encryptedMessagePayloadData() -> (data: Data, strategy: MissingClientsStrategy)?
+    func encryptedPayload() -> Payload?
```
billypchan, line 34: we could use  `var encryptedPayload: Payload?` here since it is a computed property.

Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -25,59 +25,56 @@ public let ZMFailedToCreateEncryptedMessagePayloadString = "💣"
...
-        return true
-    case (.ignoreAllMissingClients, .ignoreAllMissingClients):
-        return true
-    case (.ignoreAllMissingClientsNotFromUsers(let leftUsers), .ignoreAllMissingClientsNotFromUsers(let rightUsers)):
-        return leftUsers == rightUsers
-    default:
-        return false
-    }
-}
+    typealias Payload = (data: Data, strategy: MissingClientsStrategy)
```
billypchan, line 30: How about using a struct?

johnxnguyen:
@zenkins release major

PR #994, opened 2020-06-09
[Fix] Internal debug message for the GenericMessage

KaterinaWire:
### Issues
We do not remove the text content of protobuf message before logging them. 

### Causes
It was removed during the protobuf refactoring.

### Solutions
Create helper methods that were in objc version.

## Dependencies

**Jira ticket:**
https://wearezeta.atlassian.net/browse/ZIOS-13496

**Old objc code:**
https://github.com/wireapp/wire-ios-data-model/blob/220.0.5/Source/Model/Message/ZMGenericMessage%2BDebug.swift#L73

### DIFF
--- a/Source/Utilis/Protos/GenericMessage+Debug.swift
+++ b/Source/Utilis/Protos/GenericMessage+Debug.swift
@@ -25,7 +25,8 @@ fileprivate let redactedValue = "<redacted>"
 
 fileprivate extension Text {
     func sanitize() -> Text {
-        var text = Text(content: redactedValue)
+        var text = self
+        text.content = redactedValue
         text.linkPreview = text.linkPreview.map { $0.sanitize() }
         return text
     }
@@ -37,11 +38,11 @@ fileprivate extension LinkPreview {
     func sanitize() -> LinkPreview {
         return LinkPreview(withOriginalURL: redactedValue,
                            permanentURL: redactedValue,
-                           offset: self.urlOffset,
+                           offset: urlOffset,
                            title: redactedValue,
                            summary: redactedValue,
-                           imageAsset: self.image,
-                           article: nil,
+                           imageAsset: image,
+                           article: article.sanitize(),
                            tweet: nil)
     }
 }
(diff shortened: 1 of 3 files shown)


Source/Utilis/Protos/GenericMessage+Debug.swift
```diff
@@ -0,0 +1,80 @@
...
+            $0.permanentURL = redactedValue
+            $0.summary = redactedValue
+        }
+    }
+}
+
+// MARK: - GenericMessage
+
+public extension GenericMessage {
+    var debugDescriptionForGenericMessage: String {
```
johnxnguyen, line 65: Can we override `debugDescription`? That way we don't risk to call the wrong property (non overridden `debugDescription` )

Source/Utilis/Protos/GenericMessage+Debug.swift
```diff
@@ -0,0 +1,80 @@
...
+        var message = self
+        guard let content = content else {
+            return ""
+        }
+        switch content {
```
David-Henner, line 70: should the ephemeral case be handled as well ?

KaterinaWire: I created this method, as it was in the old version. You can verify the old code in the description.

Source/Utilis/Protos/GenericMessage+Debug.swift
KaterinaWire: I didn't override it, but name is the same now.

johnxnguyen: So why was it being overridden in the Objc version?

KaterinaWire: My suspicion is because  `debugDescription` is a `NSObject` property

johnxnguyen: Discussed 1:1 that `GenericMessage` should conform to `CustomStringConvertible` to ensure that `print(someGenericMessage)` does not print the contents of the message, since `GenericMessage` is a struct and default conversion of structs to `String` would print the all properties.

KaterinaWire:
@zenkins release minor

PR #996, opened 2020-06-19
[Fix] Attempt to fix wrong badge count

typfel:
### Issues

A badge count is sometimes generated for a conversations where all messages are read.

### Causes

The badge count is calculated by counting the number of conversations which have a `estimatedUnreadCount > 0`. This is a conversation property in the database which is calculated by counting the messages which are newer than `lastReadServerTimestamp`. It's supposed to be re-calculated when a message is inserted, deleted, updated or when the `lastReadServerTimestamp` is changed.

We've observed that sometimes the `lastReadServerTimestamp` had been updated without re-calculating the `estimatedUnreadCount` value. It's unclear exactly why this started to happen now but my guess is that something changed related to when objects are faulted / re-fetched, because we explicitly re-calculate the value when we insert, delete or update a message but when we modify the `lastReadServerTimestamp` we implicitly rely on that the conversation is fetched again on the sync context by doing the unread calculation in `awakeFromFetch()`. This usually happens because we send out a last-read update to sync the read state with other devices.

### Solutions

Don't rely on the implicit re-calculation on the `awakeFromFetch()` but instead do the unread calculation directly after modifying the `lastReadServerTimestamp`. This is a bit complicated since we only do the unread calculation on the sync context but we mark messages as read on the UI context.

The way around this is to dispatch the request to update the `lastReadServerTimestamp` to the sync context.

### Notes

We could in theory remove the unread calculations in `awakeFromFetch()` and `awakeFromInsert()` now as an optimisation.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Timestamps.swift
+++ b/Source/Model/Conversation/ZMConversation+Timestamps.swift
@@ -232,6 +232,16 @@ extension ZMConversation {
     
     // MARK: - Mark as read
     
+    /// Mark all messages in the conversation as read
+    @objc
+    public func markAsRead() {
+        guard let timestamp = lastServerTimeStamp else { return }
+        
+        enqueueMarkAsReadUpdate(timestamp)
+        savePendingLastRead()
+    }
+    
+    /// Mark messages up until the given message as read
     @objc(markMessagesAsReadUntil:)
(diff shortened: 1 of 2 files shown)


johnxnguyen:
@zenkins release patch

PR #999, opened 2020-06-25
[Feature]: digital signature feature flag

marcomaddy1982:
## What's new in this PR?

- feature flag to enable the digitally signature.

## Dependencies

JIRA: https://wearezeta.atlassian.net/browse/ZIOS-13522


Source/Model/Conversation/Team.swift
```diff
@@ -106,6 +107,12 @@ extension Team {
     }
 }
 
+extension Team {
+    public func fetchFeatureFlag(with identifier: String) -> FeatureFlag? {
+        return featureFlags.filter() { $0.identifier == identifier }.first
+    }
```
johnxnguyen, line 113: I suggest to pass the `FeatureFlagType` as an argument rather than the `String`. Also, this is a bit cleaner:

```swift
return featureFlags.first { ... }
```

Source/Model/FeatureFlag/FeatureFlag.swift
```diff
@@ -0,0 +1,67 @@
...
+
+        if let existing = team.fetchFeatureFlag(with: type.rawValue) {
+            return existing
+        }
+
+        let featureFlag = FeatureFlag.insertNewObject(in: context)
+        featureFlag.identifier = type.rawValue
+        featureFlag.isEnabled = value
+        featureFlag.updatedTimestamp = Date()
+        featureFlag.team = team
```
johnxnguyen, line 51: You could replace this with a call to `insert(with:value:team:context)`, given that you return the feature flag from that method.

Source/Model/FeatureFlag/FeatureFlag.swift
```diff
@@ -0,0 +1,67 @@
+//
+//  FeatureFlag.swift
+//  WireDataModel
+//
+//  Created by Marco Maddalena on 24/06/2020.
+//  Copyright © 2020 Wire Swiss GmbH. All rights reserved.
+//
```
johnxnguyen, line 7: Don't forget to fix this header

Source/Model/FeatureFlag/FeatureFlag.swift
```diff
@@ -0,0 +1,83 @@
...
+    public override static func entityName() -> String {
+        return "FeatureFlag"
+    }
+    
+    public var updatedAt : Date? {
+        return updatedTimestamp
+    }
+    
+    @discardableResult
+    public static func fetchOrCreate(with type: FeatureFlagType,
```
johnxnguyen, line 48: One last nit pick, I think we can improve the name with `updateOrCreate`, as "fetch" indicates that it simply retrieves an existing value without mutating it.

marcomaddy1982: Yes exactly I was thinking the same

Source/Model/FeatureFlag/FeatureFlag.swift
```diff
@@ -80,4 +80,19 @@ public class FeatureFlag: ZMManagedObject {
...
+                             context: NSManagedObjectContext) -> FeatureFlag? {
+        precondition(context.zm_isSyncContext)
+
+        let fetchRequest = NSFetchRequest<FeatureFlag>(entityName: FeatureFlag.entityName())
+        fetchRequest.predicate = NSPredicate(format: "%K == %@ && identifier == %@",
+                                             FeatureFlag.teamKey,
+                                             team,
+                                             FeatureFlagType.digitalSignature.rawValue)
+        fetchRequest.fetchLimit = 1
+        return context.fetchOrAssert(request: fetchRequest).first ?? nil
```
johnxnguyen, line 96: `?? nil` is redundant, since `first` is optional.

marcomaddy1982: Thanks nice catch!!!

marcomaddy1982: Pushed!!!

PR #1001, opened 2020-07-08
[Feature]: Avoid Digital Signature Multiple Requests

marcomaddy1982:
## What's new in this PR?

The PR try to solve the following issue:
If the user tap multiple time in a super short interval on the "Complete Signing" button before we close the OTP screen the request for the digital signature is made multiple times, this cause we don't have control to enable end disable the button (Swisscom web page).

### DIFF
--- a/Source/Model/Message/File/SignatureStatus.swift
+++ b/Source/Model/Message/File/SignatureStatus.swift
@@ -86,8 +86,9 @@ public final class SignatureStatus: NSObject {
     }
     
     public func retrieveSignature() {
-        state = .waitingForSignature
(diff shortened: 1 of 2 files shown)


Source/Model/Message/File/SignatureStatus.swift
```diff
@@ -85,8 +85,9 @@ public final class SignatureStatus: NSObject {
     }
     
     public func retrieveSignature() {
-        state = .waitingForSignature
-        RequestAvailableNotification.notifyNewRequestsAvailable(nil)
+         guard case .waitingForCodeVerification = state else { return }
```
typfel, line 88: We could add a test for this.

PR #1003, opened 2020-07-17
[Refactor] Support creating messages notifications

KaterinaWire:
## What's new in this PR?

Created isMessageSilenced function for the `ZMConversation`. These changes are necessary to manage update events.


Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -158,3 +158,35 @@ extension ZMConversationMessage {
     }
     
 }
+
+
+public extension ZMConversation {
+    func isMessageSilenced(_ message: GenericMessage, from senderID: UUID, in moc: NSManagedObjectContext) -> Bool {
```
typfel, line 164: We can retrieve the `managedObjectContext` from the `ZMConversation` and remove `moc` argument. This also has the advantage that we can supply the wrong `NSManagedObjectContext` by mistake.

typfel, line 164: We should modify the existing `isSilenced` method on `ZMConversationMessage` to use this method.

KaterinaWire: Fixed

KaterinaWire: It was done for the EAR and merged to this branch

PR #1005, opened 2020-07-22
[Feature] Add isMessageSilenced() to the ZMConversation

KaterinaWire:
## What's new in this PR?

This PR a part of the EAR feature.
Add` isMessageSilenced()` function to the `ZMConversation`

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Mute.swift
+++ b/Source/Model/Conversation/ZMConversation+Mute.swift
@@ -158,3 +158,38 @@ extension ZMConversationMessage {
     }
     
 }
+
+public extension ZMConversation {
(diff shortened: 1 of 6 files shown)


Source/Model/Conversation/ZMConversation+Mute.swift
```diff
@@ -158,3 +158,38 @@ extension ZMConversationMessage {
     }
     
 }
+
+public extension ZMConversation {
+    func isMessageSilenced(_ message: GenericMessage, senderID: UUID) -> Bool {
```
typfel, line 163: This logic is now duplicated in `ZMConversationMessage.isSilenced`. We should remove that logic delegate to this method.

KaterinaWire: Removed old `isMessageSilenced`

PR #1007, opened 2020-07-30
[Feature] Add isMessagesEncrypted property to the ZMConversation

KaterinaWire:
## What's new in this PR?

Add `isMessagesEncrypted` property to the `ZMConversation` that will show is EAR is enabled. We will use this property in the UI project to show or hide the search button in a conversation.

## Dependencies

https://wearezeta.atlassian.net/browse/ZIOS-13593


Source/Model/Conversation/ZMConversation+MessageEncryption.swift
```diff
@@ -0,0 +1,27 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+
+extension ZMConversation {
+    /// Returns true if EAR (encryption at rest) is enabled
+    public var isMessagesEncrypted: Bool {
```
johnxnguyen, line 24: I suggest `areMessagesEncryptedAtRest` or `areMessagesEncryptedInDatabase` to be more specific, because the word `encrypted` by itself is very general in the context of an encrypted messenger.

KaterinaWire: Changed

typfel:
We already expose `encryptMessagesAtRest` on the `ZMUserSession` couldn't we use that instead?

KaterinaWire:
> We already expose `encryptMessagesAtRest` on the `ZMUserSession` couldn't we use that instead?

@typfel Yes, we can use it. This PR is not needed now.

johnxnguyen:
@KaterinaWire feel free to close this.

PR #1008, opened 2020-08-03
Feature: applock store passcode

billypchan:
## What's new in this PR?

Extract `EncryptionKeys` struct form https://github.com/wireapp/wire-ios-data-model/pull/1004 and add passcode saving and fetching public methods.


billypchan:
We have a similar Key chain help class in Wire Tansport, I would move this taks to Wire Utility and we can use the code.

PR #1009, opened 2020-08-07
[Feature] Store the database key in each contexts userInfo

johnxnguyen:
## What's new in this PR?

In order to encrypt and decrypt message content in the database, we need access to the database key. The database key should never persisted and only live in RAM. We should aim to keep the references to the key to an absolute minimum so as to avoid the key surviving for extended periods of time. Furthermore, keeping the references to the key to a strictly small set of known objects allows us to remove those references when we want, such as when moving to the background.

For these reasons, I believe it's best to avoid passing the database key through all the codepaths from high level business logic to low level database access, where the encryption / decryption takes place. Instead, we can provide access to the database key for all managed objects through their corresponding managed object context.

The key is stored in the `userInfo` of each of the three managed object contexts defined in the `ManagedObjectContextDirectory`. Two helper methods on this directory object allow us to set the key and remove it from all contexts synchronously. The intended use is to store the key when the database is unlocked (after authentication) and remove the key when the database is locked.

### DIFF
--- a/Source/ManagedObjectContext/ManagedObjectContextDirectory+EncryptionAtRest.swift
+++ b/Source/ManagedObjectContext/ManagedObjectContextDirectory+EncryptionAtRest.swift
@@ -18,25 +18,22 @@
 
 import Foundation
 
-extension NSManagedObjectContext {
-    
-    public var encryptMessagesAtRest: Bool {
-        set {
-            setPersistentStoreMetadata(NSNumber(booleanLiteral: newValue),
-                                       key: PersistentMetadataKey.encryptMessagesAtRest.rawValue)
-        }
-        get {
-            (persistentStoreMetadata(forKey: PersistentMetadataKey.encryptMessagesAtRest.rawValue) as? NSNumber)?.boolValue ?? false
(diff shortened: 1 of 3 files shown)


johnxnguyen:
> I'm not sure I understand the full context (how this will be used exactly). But if you tell me that `clearDatabaseKeyInAllContexts` is called every time the app is leaving the focus/foreground and this deletes the key from memory, I'm fine with this 🙂

In the [accompanying PR](https://github.com/wireapp/wire-ios-sync-engine/pull/1279) you can see that we store the key when the database is unlocked, which occurs after successful authentication via the App Lock feature. The key is then removed from the contexts when we lock the database, which occurs every time the app goes enters the background.

> Is the a way to zeroize memory? If so, it would be good to do when clearing.

I'll need to do some research to see if we can do it. A quick google search suggests it is possible if you allocate the memory yourself.

johnxnguyen:
@franziskuskiefer so as not to block our work I will merge this now. I've create a [ticket](https://wearezeta.atlassian.net/browse/ZIOS-13703) regarding the memory zeroing, we'll take a look at it when we have some more capacity.

PR #1010, opened 2020-08-11
[Feature] Encrypt generic message data in the database

johnxnguyen:
## What's new in this PR?

The `GenericMessageData` database entity is a container for the serialized protobuf data, which is stored in the `data` property. When encryption at rest is enabled, this data must be encrypted and decrypted with the database key when stored and retrieved from Core Data. 

In order to make the encryption process transparent, it is implemented through custom accessor methods of the `data` property.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-cryptobox/pull/42


Source/Model/Message/ZMGenericMessageData.swift
```diff
@@ -17,36 +17,139 @@
...
+        guard let key = managedObjectContext?.databaseKey else { throw EncryptionError.missingDatabaseKey }
+        return try AES256GCMEncryption.decrypt(ciphertext: data, nonce: nonce, context: Data(), key: key)
+    }
+
+    private func encryptDataIfNeeded(data: Data, in moc: NSManagedObjectContext) throws -> (data: Data, nonce: Data?) {
+        guard moc.encryptMessagesAtRest else { return (data, nonce: nil) }
+        guard let key = moc.databaseKey else { throw EncryptionError.missingDatabaseKey }
+        let (ciphertext, nonce) = try AES256GCMEncryption.encrypt(message: data, context: Data(), key: key)
+        return (ciphertext, nonce)
+    }
```
johnxnguyen, line 134: The value for `context` is still to be determined.

Source/Model/Message/ZMGenericMessageData.swift
```diff
@@ -17,36 +17,139 @@
...
+    private static let log = ZMSLog(tag: "EAR")
+
+    // MARK: - Static
+
+    override open class func entityName() -> String {
+        return "GenericMessageData"
+    }
+
     public static let dataKey = "data"
+    public static let nonceKey = "nonce"
```
KaterinaWire, line 34: Do we need this key?

johnxnguyen: I'm not sure, I just added it there for completeness. Also the other two keys don't appear to have any usage.

PR #1014, opened 2020-08-17
[Feature] Encryption at rest

typfel:
## What's new in this PR?

Add support for encrypting the message data when its persisted to the database which can only be decrypted after the user authenticates inside the application.

---

- Add `EncryptionKeys` which manages the creation/deletion/access of the encryption keys involved in EAR.
- Add support for encrypting/decrypting the `underlyingMessage` of client message
- Delete code for appending delivery receipts (now sent using separate strategy)


johnxnguyen:
@zenkins release major

PR #1015, opened 2020-08-17
[Feature] Provide additional data when encrypting

johnxnguyen:
## What's new in this PR?

When encrypting generic message data we now provide additional data which will be bound to the ciphertext. A ciphertext bound to contextual data can only be decrypted with the same contextual data. For this reason, it should not change over the life of the ciphertext. 

The chosen contextual data is a combination of the user id and client id. In the case of restoring a backup on a new device, the client id will change, however we will not be backing up generic message data in its encrypted form.

### DIFF
--- a/Source/Model/Message/ZMGenericMessageData.swift
+++ b/Source/Model/Message/ZMGenericMessageData.swift
@@ -45,12 +45,14 @@ import WireCryptobox
 
     private var data: Data? {
         get {
+            guard let moc = managedObjectContext else { return nil }
+
             willAccessValue(forKey: Self.dataKey)
             let d = primitiveData
             didAccessValue(forKey: Self.dataKey)
 
             do {
-                return try decryptDataIfNeeded(data: d)
+                return try decryptDataIfNeeded(data: d, in: moc)
             } catch {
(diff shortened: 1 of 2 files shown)


Source/Model/Message/ZMGenericMessageData.swift
```diff
@@ -120,19 +122,36 @@ import WireCryptobox
...
+    private func contextData(for moc: NSManagedObjectContext) -> Data {
+        let selfUser = ZMUser.selfUser(in: moc)
+
+        guard
+            let selfClient = selfUser.selfClient(),
+            let selfUserId = selfUser.remoteIdentifier?.transportString(),
+            let selfClientId = selfClient.remoteIdentifier,
+            let context = (selfUserId + selfClientId).data(using: .utf8)
+        else {
+            fatalError("Could not obtain self user id and self client id")
```
johnxnguyen, line 149: @typfel I'm not sure what is the appropriate thing to do here. Can we reasonably expect to always have both ids?

typfel: I think so, we don't consider a user logged in until he has a client. Similarly if the client is deleted we logout.

PR #1017, opened 2020-08-19
[Fix]: NSValueTransformer Warning

marcomaddy1982:
### Issues

At runtime we have the following warning: 

CoreData: warning: Property 'payload' on Entity 'StoredUpdateEvent' is using nil or an insecure NSValueTransformer. Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead.

This is because using NSValueTransformer for core data model attribute is insecure and deprecated

More info:  https://www.kairadiagne.com/2020/01/13/nssecurecoding-and-transformable-properties-in-core-data.html

### Solutions

Created a new model version and all the transformable attributes (properties) are now subclass of NSSecureUnarchiveFromData


johnxnguyen (dismissed):
Don't forget the migration test

johnxnguyen (approved):

PR #1018, opened 2020-08-20
[Fix]: Path Traversal on Attachments

marcomaddy1982:
### Issues

Someone can send to the user a file in Wire which will be sent as an `Asset`.
The asset contains among other things the name of the file. In the code that I linked to we are creating a path to a file on the file system which has the same name.
An attacker can choose to send a file path in the name like ".../.../secret.txt" this will fool our code which creates the file URL to navigate to place on disk which was not intended.

### Solutions

We should do some validation on name property so that an attacker can't traverse the file system. The easiest way is always taking the lastPathComponent from the file name.

### Attachments

[lastPathComponent.playground.zip](https://github.com/wireapp/wire-ios-data-model/files/5101320/lastPathComponent.playground.zip)

### DIFF
--- a/Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
+++ b/Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
@@ -142,18 +142,14 @@ extension ZMAssetClientMessage: ZMFileMessageData {
         guard
             let assetURL = asset?.fileURL,
             let temporaryDirectoryURL = temporaryDirectoryURL,
-            var filename = filename
+            let filename = filename,
+            !(filename as NSString).lastPathComponent.isEmpty
         else {
             return nil
         }
         
-        filename = (filename as NSString).lastPathComponent
-        
-        guard !filename.isEmpty else {
-            return nil
-        }
-        
-        var temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(filename)
+        let secureFilename = (filename as NSString).lastPathComponent
+        var temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(secureFilename)
         
         if let fileExtension = fileExtension,
             richAssetType == .audio,
--- a/Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
+++ b/Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
@@ -148,6 +148,11 @@ extension ZMAssetClientMessage: ZMFileMessageData {
         }
         
         filename = (filename as NSString).lastPathComponent
+        
+        guard !filename.isEmpty else {
+            return nil
+        }
+        
         var temporaryFileURL = temporaryDirectoryURL.appendingPathComponent(filename)
         
         if let fileExtension = fileExtension,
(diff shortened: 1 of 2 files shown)


typfel:
Let's add a unit test for this

Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
```diff
@@ -139,9 +139,19 @@ extension ZMAssetClientMessage: ZMFileMessageData {
...
+            var filename = filename
+        else {
+            return nil
+        }
+        
+        filename = (filename as NSString).lastPathComponent
+        
+        guard !filename.isEmpty else {
+            return nil
+        }
```
johnxnguyen, line 154: Perhaps this could be written in a way that conveys the intent a bit more, something like:

```swift
guard
  let assetURL = asset?.fileURL,
  let temporaryDirectoryURL = temporaryDirectoryURL,
  let filename = filename,
  let secureFilename = (filename as NSString).lastPathComponent,
  !secureFilename.isEmpty
else {
  return nil
}
```

marcomaddy1982: `let secureFilename = (filename as NSString).lastPathComponent`

this is not an optional

johnxnguyen: Ah right, then `let secureFilename = (filename as? NSString)?.lastPathComponent`

Tests/Source/Model/Messages/ZMAssetClientMessageTests.swift
```diff
@@ -108,6 +108,18 @@ class ZMAssetClientMessageTests : BaseZMAssetClientMessageTests {
...
+    func testThatItCreatesFileAssetMessageWithNotRelativePath()
+    {
+        // given
+        let metadata = createFileMetadata(filename: "../../fileName")
+        let sut = appendFileMessage(to: conversation, fileMetaData: metadata)
+        
+        // then
+        XCTAssertNotNil(sut)
+        XCTAssertNotNil(sut?.fileURL)
+        XCTAssertTrue(((sut?.fileURL?.absoluteString.contains("../../")) != nil))
```
johnxnguyen, line 120: I think this assert would pass even if the absolute string contains "../../".

marcomaddy1982: Why?

johnxnguyen: Suppose that the `absoluteString` does contain `../../`.

Then `set?.fileURL?.absoluteString.contains("../../")` will be equal to `true`.

And `true != nil` is equal to `true`.

Therefore `XCTAssertTrue(true)` succeeds.

marcomaddy1982: Yep Yep you are right.

johnxnguyen (approved):

marcomaddy1982:
@zenkins release patch

PR #1020, opened 2020-08-26
[Improvement] Handle encryption at rest errors

johnxnguyen:
### Issues

While encryption at rest is enabled, it is possible to create broken messages in the database.

### Causes

Setting the protobuf data in an instance of `ZMGenericMessageData` may fail for a variety of reasons, but most likely it is because the database key is missing and therefore the protobuf data can't be encrypted. However, this failure is silent, meaning that the data is not encrypted and therefore not stored. This causes the broken message because any call to `underlyingMessage` will always return `nil`.

### Solutions

Propagate errors up the call stack so that the errors can be handled appropriately.


johnxnguyen:
@typfel I think we need to look at each case where an error is caught and discuss what would be the best way to handle that error. It may also be good to define errors at different layers and map the errors as they pass through the layers, as well as documenting the errors thrown.

Source/Model/Conversation/ZMConversation+GenericMessage.swift
```diff
@@ -25,20 +25,27 @@ extension ZMConversation {
     /// @param expires wether the message should expire or tried to be send infinitively
     /// @param hidden wether the message should be hidden in the conversation or not
     public func appendClientMessage(with genericMessage: GenericMessage, expires: Bool = true, hidden: Bool = false) -> ZMClientMessage? {
-        guard let nonce = UUID(uuidString: genericMessage.messageID) else { return nil }
-        guard let moc = self.managedObjectContext else { return nil }
+        guard
```
johnxnguyen, line 28: Here we can map caught errors to a new domain and rethrow them so that callers can properly handle the failure. We can do this for all "append" methods.

Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
```diff
@@ -214,36 +214,44 @@ extension ZMAssetClientMessage: ZMFileMessageData {
...
-            message.update(asset: assetData)
-            replaceGenericMessageForThumbnail(with: message)
+
+            do {
+                var message = GenericMessage()
+                try message.merge(serializedData: thumbnailMessage.serializedData())
+                message.update(asset: assetData)
+                try replaceGenericMessageForThumbnail(with: message)
+            } catch {
+                // No op
```
typfel, line 236: We should who calls this property, probably some asset strategy. Might need move the setter into separate method which can throw.

Source/Model/Message/ZMAssetClientMessage+UpdateEvent.swift
```diff
@@ -20,11 +20,14 @@ import Foundation
 
 extension ZMAssetClientMessage {
     override open func update(with updateEvent: ZMUpdateEvent, initialUpdate: Bool) {
-        guard let message = GenericMessage(from: updateEvent) else {
+        guard let message = GenericMessage(from: updateEvent) else { return }
+
+        do {
+            try add(message)
+        } catch {
             return
```
typfel, line 28: This happens inside the event processing so this should not fail due the databaseKey.

Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -467,8 +469,9 @@ struct CacheAsset: Asset {
                 genericMessage.updateAssetPreview(withUploadedOTRKey: keys.otrKey, sha256: keys.sha256!)
             }
         }
-        
-        owner.add(genericMessage)
+
+        // TODO: Handle failure?
+        try? owner.add(genericMessage)
```
typfel, line 474: Yes, this is part of the asset preprocessing which we can re-try at later point when the database key is present.

Source/Model/Message/ZMClientMessage+Editing.swift
```diff
@@ -42,18 +42,23 @@ extension ZMClientMessage {
...
+            case .text? = messageEdit.content,
+            senderUUID == sender?.remoteIdentifier
+        else {
+            return false
+        }
         
         do {
-            add(try GenericMessage(content: originalText.applyEdit(from: messageEdit.text), nonce: nonce).serializedData())
+            let genericMessage = GenericMessage(content: originalText.applyEdit(from: messageEdit.text), nonce: nonce)
+            try add(genericMessage.serializedData())
```
typfel, line 57: Also happens inside the event processing...

Source/Model/Message/ZMClientMessage+LinkPreview.swift
```diff
@@ -89,18 +89,22 @@ extension ZMClientMessage {
...
+            return
+        }
         
         let expiresAfter = deletionTimeout > 0 ? deletionTimeout : nil
         let message = GenericMessage(content: originalText.updateLinkPreview(from: updatedText), nonce: nonce, expiresAfter: expiresAfter)
-        guard let data = try? message.serializedData() else { return }
-        add(data)
+
+        // TODO: Handle this?
+        try? add(message.serializedData())
```
typfel, line 107: I think this for an incoming link preview update, if so then this should be fine since it's inside the event processing.

Source/Model/Message/ZMClientMessage+LinkPreview.swift
```diff
@@ -198,7 +202,8 @@ extension ZMClientMessage: ZMImageOwner {
             }
             
             do {
-                add(try GenericMessage(content: messageUpdate, nonce: nonce).serializedData())
+                let genericMessage = GenericMessage(content: messageUpdate, nonce: nonce)
+                try add(genericMessage.serializedData())
```
typfel, line 206: This is part of the outgoing link preview processing, which we can retry at later point if it fails.

Source/Model/Message/ZMClientMessage+Obfuscate.swift
```diff
@@ -21,41 +21,52 @@ import Foundation
...
+        else {
+            return
+        }
+
+        deleteContent()
+
+        do {
+            let data = try obfuscatedMessage.serializedData()
+            try mergeWithExistingData(data)
+        } catch {
```
typfel, line 38: Don't remember exactly when we call obfuscate, when some timer fires? Would need check if that can happen in the background.

Source/Model/Message/ZMClientMessage+UpdateEvent.swift
```diff
@@ -20,14 +20,18 @@ import Foundation
...
             applyLinkPreviewUpdate(message, from: updateEvent)
-        } else if let data = try? message.serializedData() {
-            add(data)
-            updateNormalizedText()
+        } else {
+            do {
+                try add(message.serializedData())
+                updateNormalizedText()
+            } catch {
+                // TODO: Handle?
```
typfel, line 32: Should happen inside the event processing

Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
johnxnguyen: As far as I could find, the setter is never used in production code, only in tests, and that test is about setting the thumbnail id.

Source/Model/Message/ZMAssetClientMessage+UpdateEvent.swift
johnxnguyen: Added an assertion failure.

Source/Model/Message/ZMAssetClientMessage.swift
johnxnguyen: It's not so clear how we can re-try the processing later, perhaps we can discuss this together?

Source/Model/Message/ZMClientMessage+LinkPreview.swift
johnxnguyen: Added an assertion failure.

Source/Model/Message/ZMClientMessage+LinkPreview.swift
johnxnguyen: Also here it's not clear how we can retry this.

Source/Model/Message/ZMClientMessage+Obfuscate.swift
johnxnguyen: It can happen when a `ZMMessageDestructionTimer` fires, or when we delete old ephemerals (happens when coming into the foreground, activating a session, creating a new client message transcoder).

Source/Model/Message/ZMClientMessage+UpdateEvent.swift
johnxnguyen: Added assertion failure.

Source/Model/Message/ZMAssetClientMessage.swift
johnxnguyen: One thing I noticed is that most of the methods here guard that  `underlyingMessage` is not `nil`. Previously we would only encounter `nil` if the generic message couldn't be created from the serialized data. But now it can happen when the database key is missing. We probably also have to check that usages of `underlyingMessage` also handle the `nil` case appropriately.

Source/Model/Conversation/ZMConversation+Confirmations.swift
```diff
@@ -32,11 +32,20 @@ extension ZMConversation {
...
+                let confirmation = Confirmation(messageIds: messages.compactMap(\.nonce), type: .read)
+            else {
+                continue
+            }
+
+            do {
+                let confirmationMessage = try appendClientMessage(with: GenericMessage(content: confirmation), expires: false, hidden: true)
                 confirmationMessages.append(confirmationMessage)
+            } catch {
+                Logging.messageProcessing.warn("Failed to append confirmation. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 46: Logged warning.

Source/Model/Message/Composite/CompositeMessageItemContent.swift
```diff
@@ -116,7 +116,13 @@ extension CompositeMessageItemContent: ButtonMessageData {
                 return
             }
             buttonState.state = .selected
-            self.parentMessage.conversation?.append(buttonActionWithId: buttonId, referenceMessageId: messageId)
+
+            do {
+                try self.parentMessage.conversation?.appendButtonAction(havingId: buttonId, referenceMessageId: messageId)
+            } catch {
+                Logging.messageProcessing.warn("Failed to append button action. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 123: Logged warning.

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -22,14 +22,20 @@ import WireCryptobox
...
+            return
         }
         
-        let genericMessage = GenericMessage(content: MessageHide(conversationId: conversationId, messageId: messageId))
-        ZMConversation.appendSelfConversation(genericMessage: genericMessage, managedObjectContext: message.managedObjectContext!)
+        do {
+            let genericMessage = GenericMessage(content: MessageHide(conversationId: conversationId, messageId: messageId))
+            _ = try ZMConversation.appendMessageToSelfConversation(genericMessage, in: message.managedObjectContext!)
+        } catch {
+            Logging.messageProcessing.warn("Failed to append hide message. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 37: Logged warning.

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -61,12 +67,17 @@ extension ZMMessage {
...
-        return message
+
+        do {
+            let genericMessage = GenericMessage(content: MessageDelete(messageId: messageNonce))
+            let message = try conversation.appendClientMessage(with: genericMessage, expires: false, hidden: true)
+            removeClearingSender(false)
+            updateCategoryCache()
+            return message
+        } catch {
+            Logging.messageProcessing.warn("Failed delete message for everyone. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 78: Logged warning.

Source/Model/Message/ZMAssetClientMessage+Ephemeral.swift
```diff
@@ -56,7 +56,11 @@ extension ZMAssetClientMessage {
         deleteContent()
         
         if let obfuscatedMessage = obfuscatedMessage {
-            _ = try? createNewGenericMessage(with: obfuscatedMessage.serializedData())
+            do {
+                _ = try createNewGenericMessageData(with: obfuscatedMessage)
+            } catch {
+                Logging.messageProcessing.warn("Failed to process obfuscated message. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 62: Logged warning.

Source/Model/Message/ZMAssetClientMessage+FileMessageData.swift
```diff
@@ -214,35 +214,46 @@ extension ZMAssetClientMessage: ZMFileMessageData {
...
-            message.update(asset: assetData)
-            replaceGenericMessageForThumbnail(with: message)
+
+            do {
+                var message = GenericMessage()
+                try message.merge(serializedData: thumbnailMessage.serializedData())
+                message.update(asset: assetData)
+                try replaceGenericMessageForThumbnail(with: message)
+            } catch {
+                Logging.messageProcessing.warn("Failed to set thumbnail asset id. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 236: Logged warning.

Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -414,8 +410,9 @@ struct CacheAsset: Asset {
         }
         
         do {
-            _ = owner.mergeWithExistingData(data: try genericMessage.serializedData())
+            try owner.mergeWithExistingData(message: genericMessage)
         } catch {
+            Logging.messageProcessing.warn("Failed to update asset id. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 415: Logged warning.

Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -442,7 +439,12 @@ struct CacheAsset: Asset {
         case .thumbnail:
             genericMessage.updateAssetPreview(withImageProperties: imageProperties)
         }
-        owner.add(genericMessage)
+
+        do {
+            try owner.setUnderlyingMessage(genericMessage)
+        } catch {
+            Logging.messageProcessing.warn("Failed to update preprocessed image data. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 446: Logged warning.

Source/Model/Message/ZMAssetClientMessage.swift
```diff
@@ -467,8 +469,12 @@ struct CacheAsset: Asset {
                 genericMessage.updateAssetPreview(withUploadedOTRKey: keys.otrKey, sha256: keys.sha256!)
             }
         }
-        
-        owner.add(genericMessage)
+
+        do {
+            try owner.setUnderlyingMessage(genericMessage)
+        } catch {
+            Logging.messageProcessing.warn("Failed to encrypt asset message data. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 476: Logged warning.

Source/Model/Message/ZMClientMessage+LinkPreview.swift
```diff
@@ -89,18 +89,25 @@ extension ZMClientMessage {
...
         
         let expiresAfter = deletionTimeout > 0 ? deletionTimeout : nil
         let message = GenericMessage(content: originalText.updateLinkPreview(from: updatedText), nonce: nonce, expiresAfter: expiresAfter)
-        guard let data = try? message.serializedData() else { return }
-        add(data)
+
+        do {
+            try setUnderlyingMessage(message)
+        } catch {
+            assertionFailure("Failed to set generic message: \(error.localizedDescription)")
```
johnxnguyen, line 109: Logged warning.

Source/Model/Message/ZMClientMessage+Obfuscate.swift
```diff
@@ -19,43 +19,25 @@
...
-    private func createNewGenericMessage(with data: Data) -> ZMGenericMessageData? {
-        guard let moc = self.managedObjectContext else {
-            fatalError()
+
+        deleteContent()
+
+        do {
+            try mergeWithExistingData(obfuscatedMessage)
+        } catch {
+            Logging.messageProcessing.warn("Failed to merge obfuscated message. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 39: Logged warning.

Source/Model/Message/ZMClientMessage+Editing.swift
```diff
@@ -42,18 +42,24 @@ extension ZMClientMessage {
...
+        else {
+            return false
+        }
         
         do {
-            add(try GenericMessage(content: originalText.applyEdit(from: messageEdit.text), nonce: nonce).serializedData())
+            let genericMessage = GenericMessage(content: originalText.applyEdit(from: messageEdit.text), nonce: nonce)
+            try setUnderlyingMessage(genericMessage)
         } catch {
+            Logging.messageProcessing.warn("Failed to process message edit. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 59: Logged warning.

Source/Model/Message/ZMClientMessage+LinkPreview.swift
```diff
@@ -198,8 +205,12 @@ extension ZMClientMessage: ZMImageOwner {
             }
             
             do {
-                add(try GenericMessage(content: messageUpdate, nonce: nonce).serializedData())
-            } catch { return }
+                let genericMessage = GenericMessage(content: messageUpdate, nonce: nonce)
+                try setUnderlyingMessage(genericMessage)
+            } catch {
+                Logging.messageProcessing.warn("Failed to link preview image data. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 211: Logged warning.

Source/Model/Message/ZMClientMessage+TextMessageData.swift
```diff
@@ -46,21 +46,21 @@ extension ZMClientMessage: ZMTextMessageData {
...
-                                                                 text: editedText),
-                                            nonce: editNonce)
+        let content = MessageEdit(replacingMessageID: nonce, text: editedText)
+        let updatedMessage = GenericMessage(content: content, nonce: editNonce)
+
         do {
-            self.add(try updatedMessage.serializedData())
+            try setUnderlyingMessage(updatedMessage)
         } catch {
+            Logging.messageProcessing.warn("Failed to edit text message. Reason: \(error.localizedDescription)")
```
johnxnguyen, line 60: Logged warning.

typfel (changes requested):

Source/Model/Message/Composite/CompositeMessageItemContent.swift
```diff
@@ -116,7 +116,13 @@ extension CompositeMessageItemContent: ButtonMessageData {
                 return
             }
             buttonState.state = .selected
-            self.parentMessage.conversation?.append(buttonActionWithId: buttonId, referenceMessageId: messageId)
+
+            do {
```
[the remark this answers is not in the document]
johnxnguyen: Yeah I think so. Probably also we want to make everything in the block contingent on the success of the message append.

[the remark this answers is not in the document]
johnxnguyen: Actually, maybe not.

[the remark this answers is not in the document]
johnxnguyen: Fixed

Source/Model/Message/ConversationMessage+Deletion.swift
```diff
@@ -29,7 +29,8 @@ extension ZMConversation {
         }
         
         let genericMessage = GenericMessage(content: MessageHide(conversationId: conversationId, messageId: messageId))
-        ZMConversation.appendSelfConversation(genericMessage: genericMessage, managedObjectContext: message.managedObjectContext!)
+        // TODO [John] Handle?
```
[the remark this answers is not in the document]
johnxnguyen: Moved the do catch to `hideForSelfUser` and only return early in the catch.

Source/Model/Message/Composite/CompositeMessageItemContent.swift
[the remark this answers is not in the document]
typfel: So we should not update `buttonState` after the block. Can you explain why?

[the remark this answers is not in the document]
typfel: Sorry, GitHub was showing me an outdated diff.

typfel (approved):
😅

johnxnguyen:
> 😅

Thanks @typfel , I owe you a beer!

johnxnguyen:
@zenkins release major

PR #1022, opened 2020-08-28
[Feature] Encrypt message draft when EAR is enabled

typfel:
## What's new in this PR?

Add support for storing encrypted message drafts if EAR is enabled. This is done by adding a new property on the conversation entity: `draftMessageNonce`, which if non-nil indicates that the `draftMessageData` is encrypted.

## Notes

- I moved some helper methods for encrypting/decrypting data using the database key an extension on the `NSManagedObjectContext` so that I can use them in DraftMessage.swift.
- I also refactored the tests related encrypting/decrypting using the database key to make it easier to write such tests.


johnxnguyen (approved):

typfel:
@zenkins release minor

PR #1023, opened 2020-09-07
[Fix] Dont set normalizedText when encrypt messages at rest is enabled

typfel:
### Issues

When inserting a message we update `normalizedText`, which contains a normalised version of the text message suitable for searching. This was also happening when `encryptMessagesAtRest` was enabled which defeats its purpose since we would then store a plain text version of the message in the database. 

### Solutions

Skip updating the `normalizedText` property when `encryptMessagesAtRest` is enabled in the same way we do for obfuscated messages.

### DIFF
--- a/Source/Model/Message/TextSearchQuery.swift
+++ b/Source/Model/Message/TextSearchQuery.swift
@@ -36,9 +36,12 @@ extension ZMClientMessage {
     /// Reccomputes the message's `normalizedText` property if the message
     /// has a message text, otherwise sets it to an empty String.
     override func updateNormalizedText() {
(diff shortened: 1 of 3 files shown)


marcomaddy1982 (approved):

johnxnguyen (approved):
I suppose when we support migrating to and away from the feature, we'll need to ensure we clear / update the normalized text.

typfel:
> I suppose when we support migrating to and away from the feature, we'll need to ensure we clear / update the normalized text.

True, I'll make a note of that in the migration ticket.

typfel:
@zenkins release patch

PR #1024, opened 2020-09-08
[Refactor] Extend PushToken class

KaterinaWire:
## What's new in this PR?

Move `PushTokenType` from the SE.


marcomaddy1982 (approved):

johnxnguyen (changes requested):
Do you mean to nest `TokenType` inside `PushToken`? It makes little sense to nest it inside a different struct.

KaterinaWire:
> Do you mean to nest `TokenType` inside `PushToken`? It makes little sense to nest it inside a different struct.

Yep, it make sense to do it inside `PushToken`. But in SE I need to have access to the PushTokenType.transportType, so I would like to keep it outside of `PushToken` struct.

johnxnguyen (approved):
> Yep, it makes sense to do it inside PushToken. But in SE I need to have an access to the PushTokenType.transportType, so I would like to keep it outside of PushToken struct.

You can still access it if you mark it as `public`.

KaterinaWire:
> > Yep, it makes sense to do it inside PushToken. But in SE I need to have an access to the PushTokenType.transportType, so I would like to keep it outside of PushToken struct.
> 
> You can still access it if you mark it as `public`.

Moved to the `PushToken`

PR #1025, opened 2020-09-08
add analytics id to user class

billypchan:
## What's new in this PR?

Add `analyticsIdentifier` to `ZMUser` class


marcoconti83:
The analytics ID is only necessary for one user, the logged in user. Other users in the DB won't need the analytics ID. Does it make sense to add it to the DB, or should we store it somewhere else (session variables, user defaults, keychain, ...)? @typfel

typfel:
> The analytics ID is only necessary for one user, the logged in user. Other users in the DB won't need the analytics ID. Does it make sense to add it to the DB, or should we store it somewhere else (session variables, user defaults, keychain, ...)? @typfel

That's true but we still sometimes store properties only concerning the self user in the `User` entity since we don't have a better place to put it. I thought it would be OK store it on the User entity in this case, given that we only need to access it when we are logged in.

Other places we could store it:

**CoreData key/value store:**
`+` Not adding a nil column on the user entity
`-` Need to build some scaffolding around setting and retrieving the ID

**UserDefaults:**
`-` We don't use the user defaults inside the SE so this wouldn't fit very well without our existing patterns

**Keychain**
`-` Our existing Keychain abstraction doesn't support storing anything than the cookie.

typfel (approved):

PR #1028, opened 2020-09-14
[Feature] Make database key volatile

johnxnguyen:
## What's new in this PR?

Use the `VolatileData` class to hold the database key so that it will be zeroed-out when it gets deinitialized.

## Dependencies

Needs releases with:

- [x] https://github.com/wireapp/wire-ios-utilities/pull/99


typfel (approved):

johnxnguyen:
@zenkins release major

PR #1030, opened 2020-09-17
[Feature] Migrate database to and from encryption at rest

johnxnguyen:
## What's new in this PR?

When enabling encryption at rest, we now migrate existing (relevant) data in the database so that it is in the same state as it would be if it was created while the feature was already enabled. Conversely, when we disable encryption at rest, we migrate the same data back to the normal unencrypted state.

The relevant entities are:

- `ZMGenericMessageData`: contains the message content which we encrypt / decrypt when migrating.
- `ZMMessage`: an abstract parent class that contains references to `ZMGenericMessageData`. We need to update / clear the `normalizedText` property, which contains message content in plaintext optimized for search queries.
- `ZMConversation`: it may have a draft message, which is a message that the user has typed but not yet sent. We need to encrypt / decrypt this message when migrating.

We also introduce the `migrateLocalStorage()` method on the `StorageStack`, which performs changes to a local store in an atomic manner by operating on a temporary copy of the local store. This method is used in `wire-ios-sync-engine` to perform the encryption at rest migration in a safe and performant way.


typfel:
Any idea what the perforis like? Have you tried this out on an account with a large history?

johnxnguyen:
> Any idea what the performance is like? Have you tried this out on an account with a large history?

@typfel not yet but I'll try it out today.

Source/ManagedObjectContext/NSManagedObjectContext+EncryptionAtRest.swift
```diff
@@ -19,29 +19,157 @@
 import Foundation
 import WireCryptobox
 
-extension NSManagedObjectContext {
+extension Sequence where Element: NSManagedObject {
     
-    enum EncryptionError: LocalizedError {
+    /// Perform changes on a sequence of NSManagedObjects and save at a regular internal and fault
```
johnxnguyen, line 24:
```suggestion
    /// Perform changes on a sequence of NSManagedObjects and save at a regular interval and fault
```

Tests/Source/ManagedObjectContext/NSManagedObjectContextTests+EncryptionAtRest.swift
```diff
@@ -0,0 +1,297 @@
...
+        let request = NSFetchRequest<T>(entityName: T.entityName())
+        request.returnsObjectsAsFaults = false
+        return try request.execute()
+    }
+
+    // MARK: - Positive Tests
+
+    // MARK: - Message Content
+
+    func test_existing_message_content_is_encrypted_when_ear_is_enabled() throws {
```
johnxnguyen, line 43: I suppose we should change the test names to camel case.

johnxnguyen (commented):
Looks good to me. @marcomaddy1982 you assigned me but I can't approve because I was the one that opened this PR.

Tests/Source/ManagedObjectContext/NSManagedObjectContextTests+EncryptionAtRest.swift
typfel: Done

marcomaddy1982 (approved):

typfel (approved):

PR #1031, opened 2020-09-21
[Fix] Badge count not updating on receiving a push notification

typfel:
### Issues

Badge count is not updating when receiving a notification in the background.

### Causes

After the EAR refactoring we started to [calculate an estimated unread count](https://github.com/wireapp/wire-ios-sync-engine/pull/1305) when scheduling notifications. This change was relying on the fact this estimated unread count would get updated again when the app put into the foreground and we perform the full event processing and read unread count can be calculated.

This didn't work because the estimated unread count was updated too early because we calculate the unread count when a `ZMConversation` is awoken by a fetch. This was happening before we could update the badge count and was reverting estimated unread count since messages haven't been inserted into the database at this point.

### Solutions

Stop calculating the unread count on fetches and inserts. The fact we were doing this probably is a left over from a time when we were calculating unread counts in a different way. Nowadays we re-calculate unread count on very message insert and  every time a message is read, which should cover all cases when it needs to re-calculated.

### DIFF
--- a/Source/Model/Conversation/ZMConversation.m
+++ b/Source/Model/Conversation/ZMConversation.m
@@ -233,27 +233,12 @@ const NSUInteger ZMConversationMaxTextMessageLength = ZMConversationMaxEncodedTe
 {
     [super awakeFromFetch];
     self.lastReadTimestampSaveDelay = ZMConversationDefaultLastReadTimestampSaveDelay;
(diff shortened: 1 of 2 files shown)


johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1033, opened 2020-09-29
[Performance] Optimize conversation type calculation

johnxnguyen:
### Issues

Accessing the `conversationType` property on `ZMConversation` is a relatively expensive operation.

### Causes

Although the `conversationType` is a Core Data attribute, we have a custom accessor that will override the value if the conversation is a group conversation in a team but should be considered as a 1:1 conversation. There are certain criteria that must be satisfied to consider the group as a 1:1, and most of these criteria are fast to check. The most expensive check involves querying the conversation participants, and in the worst case scenario the entire set of all participants is enumerated 4 times. This is due to non-lazy sequence transformations over the `participantRoles` property (mapping and filtering).

### Solutions

By checking the count of `localParticipantRoles` (an O(1) expense) before accessing `localParticipantsExcludingSelf`, we can ensure that we would only ever enumerate over two participants. This make check a O(1) expense.

### DIFF
--- a/Source/Model/Conversation/ZMConversation.m
+++ b/Source/Model/Conversation/ZMConversation.m
@@ -432,18 +432,17 @@ const NSUInteger ZMConversationMaxTextMessageLength = ZMConversationMaxEncodedTe
     // 3. Conversation has only one other participant.
     // 4. This participant is not a service user (bot).
 
+    // Performance note: localParticipantsExcludingSelf will enumerate over all
+    // local participant roles, so check its count first to avoid unncessary iterations.
+
     if (conversationType == ZMConversationTypeGroup &&
         self.teamRemoteIdentifier != nil &&
         self.userDefinedName.length == 0 &&
-        self.localParticipantRoles.count == 1)
+        self.localParticipantRoles.count == 2 &&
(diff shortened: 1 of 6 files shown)


Source/Model/Conversation/ZMConversation.m
```diff
@@ -431,12 +431,19 @@ - (ZMConversationType)conversationType
     // 2. Has no name given.
     // 3. Conversation has only one other participant.
     // 4. This participant is not a service user (bot).
+
```
marcomaddy1982, line 434: Could it be written this part using a guard statement? If yes, IMHO it would be more easier to read.

marcomaddy1982 (approved):

Source/Model/Conversation/ZMConversation.m
johnxnguyen: There are no `guard`s in objective c 😄 .

marcomaddy1982: I guess you are right!!! 😄 😄

johnxnguyen:
@zenkins release patch

PR #1035, opened 2020-09-30
[Performance] Calculating unread messages

KaterinaWire:
### Issues

We re-calculate the unread message count on every message insert or timestamp change. This is wasteful in a scenario where we receive x messages in the same conversation, we would then calculate the unread count x times when it's enough to do it once after inserting all the messages.
### Solutions

Mark a conversation as needed to update the unread message counter. 

## Dependencies

https://wearezeta.atlassian.net/browse/ZIOS-13812


johnxnguyen (changes requested):
You need to set the flag back to `false` after updating the unread count. Don't forget to assert this behavior with tests.

KaterinaWire:
> You need to set the flag back to `false` after updating the unread count. Don't forget to assert this behavior with tests.



> You need to set the flag back to `false` after updating the unread count. Don't forget to assert this behavior with tests.

Yes, I missed this part. I assume that the best place to set flag back to `false` in `calculateLastUnreadMessages()`, since we call this method after inserting messages from all events. I added a test to check the behavior of this flag, but now I'm thinking of edge cases for testing.

johnxnguyen (approved):

Source/Public/ZMConversation.h
```diff
@@ -96,6 +96,9 @@ extern NSString * _Null_unspecified const ZMIsDimmedKey; ///< Specifies that a r
 /// For group conversation this will be nil, for one to one or connection conversation this will be the other user
 @property (nonatomic, readonly, nullable) ZMUser *connectedUser;
 
+/// Returns whether unread messages should be re-calculated
+@property (nonatomic) BOOL needsToCalculateUnreadMessages;
```
typfel, line 100: I think we would need persist this property otherwise we would lose the state of this flag and not re-calculate the unread messages when we need to.

typfel (changes requested):

typfel:
I experienced that I had an unread count which didn't go away after https://github.com/wireapp/wire-ios-data-model/pull/1031 was merged. 

I would propose that we add back the calculating the unread count in `awakeFromFetch` and `awakeFromInsert` but only if `needsToCalculateUnreadMessages` is `true`.

Source/Public/ZMConversation.h
KaterinaWire: Added this property to CoreData

KaterinaWire:
> I experienced that I had an unread count which didn't go away after #1031 was merged.
> 
> I would propose that we add back the calculating the unread count in `awakeFromFetch` and `awakeFromInsert` but only if `needsToCalculateUnreadMessages` is `true`.

Done

typfel:
You've missed a few places where we also call `calculateUnreadMessages()`

`updateTimestampsAfterUpdatingMessage()`
`updateTimestampsAfterInsertingMessage()`
`updateTimestampsAfterDeletingMessage()`

Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -319,7 +319,7 @@ extension ZMConversation {
     /// Calculates the the last unread knock, missed call and total unread unread count. This should be re-calculated
     /// when the last read timetamp changes or a message is inserted / deleted.
     @objc
-    func calculateLastUnreadMessages() {
+    public func calculateLastUnreadMessages() {
```
typfel, line 322: No longer needs to be public

typfel:
Looking good 👍

 Let's also add a test for `calculateLastUnreadMessages(in managedObjectContext:)`

Source/Model/Conversation/ZMConversation+Timestamps.swift
KaterinaWire: Done

KaterinaWire:
> Looking good 👍
> 
> Let's also add a test for `calculateLastUnreadMessages(in managedObjectContext:)`

Added a test

Tests/Source/Model/Conversation/ZMConversationTests+UnreadMessages.swift
```diff
@@ -0,0 +1,55 @@
...
+            XCTAssertEqual(conversation.estimatedUnreadCount, 0)
+            
+            // when
+            ZMConversation.calculateLastUnreadMessages(in: self.syncMOC)
+            XCTAssertTrue(self.syncMOC.saveOrRollback())
+            
+            // then
+            XCTAssertEqual(conversation.estimatedUnreadCount, 1)
+            XCTAssertFalse(conversation.needsToCalculateUnreadMessages)
+        }
```
typfel, line 53:
```suggestion
        syncMOC.performGroupedBlockAndWait {
            // given
            let conversation = ZMConversation.insertNewObject(in:self.syncMOC)
            conversation.conversationType = .group
            conversation.remoteIdentifier = UUID.create()
            
           let knock = GenericMessage(content: Knock.with { $0.hotKnock = false })
            let message = ZMClientMessage(nonce: UUID(), managedObjectContext: self.syncMOC)
            do {
                try message.setUnderlyingMessage(knock)
            } catch {
                XCTFail()
            }
            message.serverTimestamp = Date()
            message.visibleInConversation = conversation
            conversation.updateTimestampsAfterUpdatingMessage(message)
            
            XCTAssertTrue(conversation.needsToCalculateUnreadMessages)
            XCTAssertEqual(conversation.estimatedUnreadCount, 0)
            
            // when
            ZMConversation.calculateLastUnreadMessages(in: self.syncMOC)
            XCTAssertTrue(self.syncMOC.saveOrRollback())
            
            // then
            XCTAssertEqual(conversation.estimatedUnreadCount, 1)
            XCTAssertFalse(conversation.needsToCalculateUnreadMessages)
        }
```

I'd propose to only a have single `given, when, then`, because then it's more clear what we are testing in this test.  That `needsToCalculateUnreadMessages` is set to `true` after `conversation.updateTimestampsAfterUpdatingMessage ()` is already tested in another test you added.

Tests/Source/Model/Conversation/ZMConversationTests+Timestamps.swift
```diff
@@ -159,6 +159,75 @@ class ZMConversationTests_Timestamps: ZMConversationTestsBase {
         }
     }
     
+    func testThatUnreadCountIsUpdatedWhenMessageIsInsertedFromUpdateEvent() {
+
```
typfel, line 163: Looking at these tests, I think they are testing too many things. We already have other tests for testing that the `estimatedUnreadCount` is increased when calling `calculateLastUnreadMessages()`. What's new and what we should test in isolation is that: 
- `needsToCalculateUnreadMessages` is set to true after inserting a message from an update event.
- `needsToCalculateUnreadMessages` is set to false after calling `calculateLastUnreadMessages()`

KaterinaWire: Done. Separated the logic of this test into 2 new tests.

typfel (approved):

KaterinaWire:
@zenkins release minor

PR #1036, opened 2020-10-02
[Feature]: Analytics ID Improvements

marcomaddy1982:
## What's new in this PR?

we are converting the UUID into a transportString

### DIFF
--- a/Source/Model/User/ZMUser.m
+++ b/Source/Model/User/ZMUser.m
@@ -707,7 +707,7 @@ static NSString *const AnalyticsIdentifierKey = @"analyticsIdentifier";
     
     if (selfUser == nil) {
         selfUser = [ZMUser insertNewObjectInManagedObjectContext:moc];
-        selfUser.analyticsIdentifier = [[NSUUID UUID] UUIDString];
+        selfUser.analyticsIdentifier = [[NSUUID UUID] transportString];
         RequireString([moc obtainPermanentIDsForObjects:@[selfUser] error:&error],
                       "Failed to get ID for self user: %lu", (long) error.code);
     }
--- a/Tests/Source/Model/User/ZMUserTests.m
+++ b/Tests/Source/Model/User/ZMUserTests.m
@@ -2545,5 +2545,21 @@ static NSString * const domainValidCharactersLowercased = @"abcdefghijklmnopqrst
     XCTAssertFalse(untrusted);
 }
 
+- (void)testThatItCanParseLowercaseUUID
+{
+    // given
+    NSString *string = [@"E3F77380-A03E-45BB-A598-D361E3220001" lowercaseString];
+    NSUUID *expected = [[NSUUID alloc] initWithUUIDString:@"E3F77380-A03E-45BB-A598-D361E3220001"];
+    
+    // when
+    NSString *analyticsIdentifier = [expected transportString];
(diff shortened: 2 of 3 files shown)


Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -2545,5 +2545,21 @@ - (void)testThatItReturns_UnTrusted_NO_WhenThereAreNoUnTrustedClients
     XCTAssertFalse(untrusted);
 }
 
+- (void)testThatItCanParseLowercaseUUID
```
typfel, line 2548: This test doesn't exercise the code path where the `analyticsIdentifier` is initialised. We don't need need test that `transportString` generate a lower case UUID representation, that's already tested in `wire-ios-transport`.

typfel (changes requested):

Source/Model/User/ZMUser.m
```diff
@@ -707,7 +707,7 @@ + (ZMUser *)createSessionIfNeededInContext:(NSManagedObjectContext *)moc withSel
     
     if (selfUser == nil) {
         selfUser = [ZMUser insertNewObjectInManagedObjectContext:moc];
-        selfUser.analyticsIdentifier = [[NSUUID UUID] UUIDString];
+        selfUser.analyticsIdentifier = [[NSUUID UUID] transportString];
```
billypchan, line 710: It would be clean to create a method like `selfUser.assignAnalyticsIdentifier` and create a test for it.

marcomaddy1982: I added a new test, check if it good as well.

Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -2561,5 +2561,10 @@ - (void)testThatItCanParseLowercaseUUID
     XCTAssertEqualObjects(string, sut.analyticsIdentifier.lowercaseString);
 }
 
+- (void)testThatAnalyticsIdentifierIsNotNill {
+    ZMUser *sut = [ZMUser selfUserInContext:self.uiMOC];
+    XCTAssertNotNil(sut.analyticsIdentifier);
```
billypchan, line 2566: It will be nice to test the UUID's format, e.g. regex to check the format of the UUID. And you may write this test in Swift.

typfel: I don't think we should start testing the `transportString` with regex such tests belong to `wire-ios-transport`.

typfel: It would enough to verify that the `analyticsIdentifier` can be converted back into a Date with calling `[NSDate dateFromTransportString:]`

Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -2545,5 +2545,19 @@ - (void)testThatItReturns_UnTrusted_NO_WhenThereAreNoUnTrustedClients
     XCTAssertFalse(untrusted);
 }
 
+- (void)testThatAnalyticsIdentifierCanBeConvertedBack
```
typfel, line 2548:
```suggestion
- (void)testThatAnalyticsIdentifierIsEncodedAsUUIDTransportString
```

typfel (approved):

marcomaddy1982:
@zenkins release patch

PR #1039, opened 2020-10-07
[Refactor] Decouple change detection

johnxnguyen:
## What's new in this PR?

This PR is part of ongoing work to refactor the `NotificationDispatcher` such that it can operate in an "eco" mode which favors speed over detail.

### Issues

It is difficult to change the change detection logic in the `NotificationDispatcher`. 

### Causes

Change detection is built right into the `NotificationDispatcher` in a highly coupled way.

### Solutions

Decouple the change detection into a separate class. The idea is that in the future we can easily swap in a different "eco" change detector.


Source/Notifications/Change detection/DetailedChangeDetector.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        // Check for changed keys and affected keys.
+        let changes: [ZMManagedObject: Changes] = changedObjects.mapToDictionary{ object in
+            // (1) Get all the changed keys since last Save.
+            let changedKeys = getChangedKeysSinceLastSave(object: object)
+            guard !changedKeys.isEmpty else { return nil }
+
+            // (2) Get affected changes.
+            detectChangesCausedByChangeInObjects(updatedObject: object, knownKeys: changedKeys)
```
typfel, line 87: Why is the argument called `knownKeys` and not `changedKeys`?

Source/Notifications/Change detection/DetailedChangeDetector.swift
```diff
@@ -0,0 +1,120 @@
...
+
+            guard !affectedKeys.isEmpty else { return nil }
+            return Changes(changedKeys: affectedKeys)
+        }
+
+        // (4) Merge the changes with the other ones.
+        allChanges = allChanges.merged(with: changes)
+    }
+
+    private func detectChangesCausedByChangeInObjects(updatedObject: ZMManagedObject, knownKeys: Set<String>) {
```
typfel, line 102: I think it would be easier to follow the code if this method returns the result instead of directly altering `allChanges`, because it's called from `detectChanges(for:)` and there we update `allChanges` at the end and it's surprising that we modify `allChanges` also while calculating the changes.

typfel (changes requested):

Source/Notifications/Change detection/DetailedChangeDetector.swift
johnxnguyen: Not sure, so far I've just moved code. I intend to refactor and clean up this logic too (and also elsewhere because it's super hard to follow) but didn't want to bloat this PR any further.

Source/Notifications/Change detection/DetailedChangeDetector.swift
johnxnguyen: Yep I agree 👍

typfel (approved):

PR #1042, opened 2020-10-09
[Performance] Create a detector for potential changes

johnxnguyen:
### Context

This PR is part of ongoing work to refactor the `NotificationDispatcher` such that it can operate in an "economical" mode which favors speed over detail.

### New

`PotentialChangeDetector` is a light weight detector that simply keeps track of whether an object was updated, inserted or deleted.

`PotentialObjectChangeInfo` is the counterpart to `ObjectChangeInfo`, which contains the potential changes for a given object.

`ChangeInfo` describes two types of changes, `explicit` or `potential`, it is an abstraction / container for concrete change info objects.

### Changed

All instance of `ObjectChangeInfo` have been wrapped inside `ChangeInfo.explicit(changes: ObjectChangeInfo)`.


typfel:
So we now have `PotentialChangeDetector` and `ExplicitChangeDetector`. How will they be used by NotificationDispatcher? Will the `PotentialChangeDetector` also trigger change notifications?

Source/Notifications/Change detection/ChangeInfo.swift
```diff
@@ -0,0 +1,46 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+
+public enum ChangeInfo {
```
typfel, line 22: Is this a new interface for the consumer of the observation system?

johnxnguyen:
> So we now have `PotentialChangeDetector` and `ExplicitChangeDetector`. How will they be used by `NotificationDispatcher`? Will the PotentialChangeDetector also trigger change notifications?

@typfel , the `NotificationDispatcher` has a property called `changeDetector` which can be either of the two detectors. We'll introduce an operation mode for the dispatcher, when it changes mode we'll swap in a different detector. The mode can also be used to control how frequently we fire notifications. My intention is that we only fire notifications for potential changes when we transition away from the "eco" mode.

Then in the object observers, when we unpack the change info from the notification, we will first check for normal (explicit) change info, and if they exist, invoke the "objectDidChange" callback on the observers. Otherwise, if the change is a potential one, we will invoke a new "objectMayHaveChanged" callback on the observers, passing in the potential changes.

Source/Notifications/Change detection/ChangeInfo.swift
johnxnguyen: Yes, as I mention in the other comment, all change observers will now have to callbacks which will be invoked depending on the type of `ChangeInfo`. 

For example: 

If the change detector returns a `ChangeInfo.explicit(changes: someUserChangeInfo)`, this would lead to the invocation of `observer.userDidChange(someUseChangeInfo)`. If on the other hand the detector returns `ChangeInfo.potential(changes: somePotentialChangeInfo)`, then we would invoke `observer.userMayHaveChanged(somePotentialChangeInfo)`.

typfel: Does the `somePotentialChangeInfo` in `observer.userMayHaveChanged(somePotentialChangeInfo)` ever contain any extra information? Isn't it binary, it changed or didn't?

Just thinking that has a consumer of the observation system it would nicer to not have two different callbacks.

typfel: But I guess we can this this in the next PR.

johnxnguyen: It contains some values to indicate the change type, such as whether the object was inserted, deleted, modified, or any combination of those. I thought it might be helpful for the user to have some indication of the type of change.

Initially I was thinking to avoid the second callback but I decided against it for the following reasons:

Often in the callbacks different code is conditionally executed based on the change info. We may for example have 3 different side effects depending on 3 change flags. In the case that we don't know the change flags (because of potential changes), we may actually want to execute all of the possible side effects, to be sure. But this would mean we would need to extend the all conditions to account for potential changes, which feels a bit clunky. 

This leads to another issue of discoverability: if the we include "hasPotentialChanges` in the existing change info, then the observers must explicitly remember to handle the case of potential changes. By conceptually separating the types of changes into different change info types and callbacks, the observers are required to handle the case.

Maybe I'm over-engineering it though, what do you think?

johnxnguyen: Hmm, we could pass the new `ChangeInfo` directly to the existing callback and make the observers handle both cases there, instead of branching out to separate callbacks. That would solve both these issues 😄

typfel: Consumers of the observation system in the UI will probably have these to use cases.

```
If (the object I'm interested in changed and it the change concerns what I'm displaying)
   "reload the view"

If (the object I'm interested in changed but we don't know what changed)
   "reload the view"
```

if you handle the potential changes in `changedKeysContain(keys:)` you wouldn't have to update all the observers UI.

johnxnguyen: So we would say that if an object has potentially changed, then `changedKeysContains(keys:)` always returns `true`?

typfel: Yes

typfel: If realise that we need more fine grained control over this we can always expand the API later.

typfel (approved):

Source/Notifications/Change detection/ChangeInfo.swift
johnxnguyen: Ok I think it can work.

PR #1045, opened 2020-10-12
[Fix] Don't fault users which are likely to be accessed again soon

typfel:
### Issues

Accessing users during event processing was often ending up doing a fetch request which is expensive.

### Causes

All users were faulted on a save even though we have a rule to keep users in memory if they belong to a recently modified conversation. This rule was broken since the introduction of participant roles, it was putting `ParticipantRole` objects in the set of `usersToKeep`, which is expected to only contains `ZMUser` objects.

### Solutions

Call `localParticipants` instead which returns a set of `ZMUser`:s.


KaterinaWire (approved):

johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1046, opened 2020-10-13
[Performance] Change tracking operation mode

johnxnguyen:
### Context

This PR is part of ongoing work to refactor the NotificationDispatcher such that it can operate in an "economical" mode which favors speed over detail.

### Goal

To be able to change the tracking behavior of `NotificationDispatcher` at runtime so that it uses the `PotentialChangeDetector` and fires notifications less frequently.

### New

Added `public operationMode: OperationMode` to the notification dispatcher to differentiate between normal operation and economical operation.

### DIFF
--- a/Source/Notifications/NotificationDispatcher.OperationMode.swift
+++ b/Source/Notifications/NotificationDispatcher.OperationMode.swift
@@ -30,7 +30,7 @@ public extension NotificationDispatcher {
 
     enum OperationMode {
 
-        /// Change detection is highly detailed and observers are notified infrequntly.
+        /// Change detection is highly detailed and observers are notified frequently.
 
         case normal
 
--- a/Source/Notifications/NotificationDispatcher.OperationMode.swift
+++ b/Source/Notifications/NotificationDispatcher.OperationMode.swift
@@ -1,0 +1,44 @@
+//
+// Wire
+// Copyright (C) 2020 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+
+public extension NotificationDispatcher {
+
+    /// Describes how the `NotificationDispatcher` should behave in relation to the
+    /// expected work load (the number of database changes it must process).
+    ///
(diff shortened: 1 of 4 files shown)


marcomaddy1982 (approved):

Source/Notifications/NotificationDispatcher.swift
```diff
@@ -240,22 +268,21 @@ import CoreData
...
+    }
+
+    private func forwardChangesToConversationListObserver(modifiedObjects: ModifiedObjects) {
+        let insertedLabels = modifiedObjects.inserted.compactMap { $0 as? Label }
+        let deletedLabels = modifiedObjects.deleted.compactMap { $0 as? Label }
+        conversationListObserverCenter.folderChanges(inserted: insertedLabels, deleted: deletedLabels)
+
+        let insertedConversations = modifiedObjects.inserted.compactMap { $0 as? ZMConversation }
+        let deletedConversations = modifiedObjects.deleted.compactMap { $0 as? ZMConversation }
+        conversationListObserverCenter.conversationsChanges(inserted: insertedConversations, deleted: deletedConversations)
```
johnxnguyen, line 285: I moved this here to clean it up but also to express that it is part of processing and not dependent on the operation mode.

billypchan:
Is there any test result about the amount of performance improvement?

johnxnguyen:
> Is there any test result about the amount of performance improvement?

@billypchan not yet but I'm trying to gather some.

johnxnguyen:
> > Is there any test result about the amount of performance improvement?
> 
> @billypchan not yet but I'm trying to gather some.

So it's a bit hard to set up a realistic test scenario that mimics processing a large number of Core Data changes. Ideally what we want is a lot of different changes (updates, insertions, deletions) with many relationships between objects and a lot of observers. This really hard to set up so don't think we can get performance results for this.

However, I was able to set up a super basic test that only deals with a large number of inserted objects. When running in `economical` mode, processing 100 insertions was about 10% faster than when running in `normal` mode.

This isn't a large improvement, but it's also not a very good test. Processing insertions is fairly simple even in `normal` mode. But it also suggests that the performance improvement is greater in more realistic scenarios because the bottleneck in normal mode is detecting changed keys and affected objects, which is not done at all in `economical` mode.

Source/Notifications/NotificationDispatcher.OperationMode.swift
```diff
@@ -0,0 +1,43 @@
...
+    /// Describes how the `NotificationDispatcher` should behave in relation to the
+    /// expected work load (the number of database changes it must process).
+    ///
+    /// The normal operation mode is a relatively expensive operation and may cause
+    /// a considerably large amount of observer code to be triggered. If several
+    /// changes are made to the database, you may wish to switch to the `econmical` mode.
+
+    enum OperationMode {
+
+        /// Change detection is highly detailed and observers are notified infrequntly.
```
typfel, line 33:
```suggestion
        /// Change detection is highly detailed and observers are notified frequently.
```

typfel, line 33:
```suggestion
        /// Change detection is highly detailed and observers are notified frequently.
```

typfel (approved):

PR #1048, opened 2020-10-16
[Performance] Object change tracking

johnxnguyen:
## What's new in this PR?

This PR refactors the observation system used to broadcast changes to specific Core Data entities. Specifically, two operation modes have been defined to control how detailed and frequent the object change notifications are broadcast. 

In the `normal` mode, modified objects are scrutinized in order to determine not only which of their keys have changed, but also which keys of other objects that have dependencies on changed keys. The result is a highly detailed and complete understanding of how the database has changed. These notifications are also sent as frequently as possible, typically as soon as the change data has been committed to the database.

In the `economical` mode by contrast, the detail of the changes are kept to an absolute minimum: we only keep track of which objects have potentially changed. Furthermore, the object change notifications not broadcasted until we switch back to `normal` mode. When an object has potentially changed, the observers see this as the object has potentially changed *all* of its keys. 

These differences between modes can be best summarized by their intended use cases:

Use `normal` mode when you expect database changes in small, (relatively) infrequent batches, such as is expected when the application is synchronized and the websocket is connected.

Use `economical` mode when you expect a large amount of database changes in a short amount of time, such as when the application is synchronizing the local state with the remote state.

## Notes

All commits in this PR have be approved in separate PRs.


typfel (approved):

johnxnguyen:
@zenkins release major

johnxnguyen:
@zenkins release major

billypchan:
@zenkins release major

billypchan:
@zenkins release major

billypchan:
@zenkins release major

billypchan:
@zenkins release major

PR #1051, opened 2020-10-27
[Fix] Restoring a large backup freezes the app

typfel:
### Issues

Restoring a large backup freezes the app

### Causes

After restoring a large backup all messages goes through the observer system and specifically the `NewUnreadMessageChangeObserver`. This change observer relies the `unreadMessages` property in the `NotificationDispatcher` which collects unread messages. These unread messages never gets reset though so we keep reporting more and more message as unread, which causes the UI thread to freeze.

### Solutions

Reset the `unreadMessages` property after firing the notifications.

## Notes

We also don't want to run the all existing messages through observer system when restoring a backup but that will be fixed in a separate PR.


billypchan (approved):

johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1053, opened 2020-10-30
[Improvement] Normalize asset filenames

johnxnguyen:
## Issue

Asset filenames currently have excessing diacritics stripped from them, but it would be better to normalize the filename to contain only characters from the POSIX fully portable character set.

## Solution

Fortunately we already have a `String` extension that handles this, we just have to use it.


marcomaddy1982 (approved):

johnxnguyen:
@zenkins release patch

PR #1058, opened 2020-11-09
[Fix] Crash when accessing modifiedKeys property on iOS 12

typfel:
### Issues

App crashes when accessing a `modifiedKeys` property on an entity when running iOS 12

### Causes

```
NSInvalidArgumentException:
Object of class __NSSingleObjectSetI is not among allowed top level class list (
    NSArray,
    NSDictionary,
    NSString,
    NSNumber,
    NSDate,
    NSData,
    NSURL,
    NSUUID,
    NSNull
)
```

Is thrown by `[NSSecureUnarchiveFromDataTransformer reverseTransformedValue:]`.

This started to happen because we switched to use secure encoding for our transformable properties in Core Data since it's recommend by Apple and will soon be the default. This works great on iOS 13 and iOS 14 but fails on iOS 12 since `NSSet` is not included in the list of allowed top level classes and `modifiedKeys` stores a `NSSet`

### Solutions

Add `NSSet` to the list of allowed top level classes in our custom secure transformer and use it for every transformable property.

## Notes

These changes to data model doesn't affect the migration since they are only evaluated at runtime. See https://github.com/wireapp/wire-ios-data-model/pull/1019


johnxnguyen (approved):

marcomaddy1982 (approved):

typfel:
@zenkins release patch

PR #1061, opened 2020-11-10
[Feature] Broadcast analytics identifier when generated

johnxnguyen:
### Goal

When the analytics identifier is generated for the self user, we must send it to the other devices for that user so that they can be notified of the new identifier to use.

### Solutions

To send the identifier within an EE2E payload, we post a `DataTransfer` message containing the identifier to the self conversation.

## Notes

In a future PR, this data transfer message will be received on other clients and processed.

### DIFF
--- a/Tests/Source/Model/User/ZMUserTests+AnalyticsIdentifier.swift
+++ b/Tests/Source/Model/User/ZMUserTests+AnalyticsIdentifier.swift
@@ -22,7 +22,7 @@ import XCTest
 
 class ZMUserTests_AnalyticsIdentifier: ModelObjectsTests {
 
-    func test_the_analytics_identifier_is_automatically_generated() {
+    func testTheAnalyticsIdentifierIsAutomaticallyGenerated() {
         // Given
         let sut = createUser(selfUser: true, inTeam: true)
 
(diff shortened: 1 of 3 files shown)


Tests/Source/Model/User/ZMUserTests+AnalyticsIdentifier.swift
```diff
@@ -56,13 +56,58 @@ class ZMUserTests_AnalyticsIdentifier: ModelObjectsTests {
...
+
+        // Then
+        let selfConversation = ZMConversation.selfConversation(in: uiMOC)
+        XCTAssertEqual(selfConversation.numberOfDataTransferMessagesContaining(analyticsIdentifier: identifier!), 1)
+    }
+
+}
+
+// MARK: - Helpers
+
```
KaterinaWire, line 89: It might make sense to add a test and check that there are no `DataTransfer` messages if we aren't using the `analyticsIdentifier`.

KaterinaWire (approved):

Tests/Source/Model/User/ZMUserTests+AnalyticsIdentifier.swift
johnxnguyen: Added some extra asserts to check for this.

billypchan (approved):

PR #1064, opened 2020-11-10
[Feature] Process data transfer message

johnxnguyen:
## What's new in this PR?

In this PR, we add support to process an incoming `DataTransfer` message containing a tracking identifier.

### Goal

When a new analytics identifier is broadcast from another device, all other devices must receive it through a `DataTransfer` message and store it locally.

### Solutions

Process the message like from the update event by extracting the identifier and overwriting the `analyticsIdentifier` of the self user.

## Notes

This PR also makes the `analyiticsIdentifier` observable, which will be used observe changes from the UI.

### DIFF
--- a/Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
+++ b/Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
@@ -52,7 +52,8 @@ extension ZMUser : ObjectInSnapshot {
             #keyPath(ZMUser.serviceIdentifier),
             #keyPath(ZMUser.providerIdentifier),
             ZMUserKeys.legalHoldRequest,
-            #keyPath(ZMUser.isUnderLegalHold)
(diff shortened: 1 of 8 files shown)


Source/Notifications/ObjectObserverTokens/UserChangeInfo.swift
```diff
@@ -160,6 +161,10 @@ extension ZMUser : ObjectInSnapshot {
         return changedKeys.contains(#keyPath(ZMUser.participantRoles))
     }
 
+    public var analyticsIdentifierChanged: Bool {
```
KaterinaWire, line 164: Should we add a test to check that it notifies the observer?

johnxnguyen: Good point, added one.

KaterinaWire (approved):

billypchan (approved):

PR #1065, opened 2020-11-11
[Feature] Sync analytics identifier

johnxnguyen:
## What's new in this PR?

This is the feature branch for syncing the analytics identifier across the self user's devices. This allows correlate events to a single user, not for each device.


KaterinaWire (approved):

marcomaddy1982 (approved):

Source/Model/User/ZMUser+AnalyticsIdentifier.swift
```diff
@@ -0,0 +1,66 @@
...
+            let value = primitiveAnalyticsIdentifier
+            didAccessValue(forKey: #keyPath(analyticsIdentifier))
+
+            if let value = value {
+                return value
+            } else {
+                let identifier = UUID()
+                let identifierString = identifier.transportString()
+                self.analyticsIdentifier = identifierString
+                broadcast(identifier: identifier)
```
marcoconti83, line 48: If you receive an ID from another device, is because that device has sent it to all your other devices. 
If you have an old local ID, and you receive a new one, this will trigger you sending the new ID to all other devices again, even if not necessary.

I think we can live with this redundancy for now, but it's something to optimize later.

billypchan (approved):

Source/Model/User/ZMUser+AnalyticsIdentifier.swift
marcoconti83: Ah sorry, I misunderstood completely!
Please disregard my comment :)

johnxnguyen: I'm not sure I understand. The `broadcast` method is only invoked when we try to access the identifier but none exists. 

If we already have an old local id, and we receive a new one, then it would go through the setter, but in that case it doesn't broadcast it.

Or am I missing something?

johnxnguyen:
@zenkins release major

PR #1066, opened 2020-11-16
[Feature] Define a Feature entity

KaterinaWire:
### Issues
This PR is a part of the` Enforce App-Lock` feature. To be able to configure features, we will fetch it from BE and store them in Core Data.

### Solutions
Define a Feature entity that will contain feature configurations.


Source/Model/FeatureConfig/Feature.swift
```diff
@@ -0,0 +1,31 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+@objcMembers
+public class Feature: ZMManagedObject {
+   
+    @NSManaged public var name: String
+    @NSManaged public var status: String
```
johnxnguyen, line 25: In another PR I discussed having a `Status` enum. If we defined it in data model, we could also use it here to give this a stronger type.

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -0,0 +1,45 @@
...
+public enum FeatureStatus: String, Codable {
+  case enabled
+  case disabled
+}
+
+@objcMembers
+public class Feature: ZMManagedObject {
+   
+    @NSManaged public var name: String
+    @NSManaged public var statusValue: String
```
johnxnguyen, line 30: Better to make `statusValue` private since you have a custom accessor.

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -0,0 +1,45 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public enum FeatureStatus: String, Codable {
+  case enabled
+  case disabled
+}
```
johnxnguyen, line 24: Better to declare this within `Feature`, e.g so you would access it like `Feature.Status.enabled`

johnxnguyen (changes requested):

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: Done

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: Fixed

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: Fixed

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -18,25 +18,25 @@
...
     
     public override static func entityName() -> String {
         return "Feature"
     }
     
-    public var status: FeatureStatus {
+    public var status: Status {
         get {
-            return FeatureStatus(rawValue: statusValue) ?? .disabled
+            return Status(rawValue: statusValue) ?? .disabled
```
johnxnguyen, line 35: The only way that we would get a `nil` value here is if the case names change while there exists values in the database. In which case, that would be a programmer error. Do you think in case it is better to intentionally crash (with a message) rather than disable the feature?

KaterinaWire: Do you mean to have a third case as .unknown and manage it somewhere?

johnxnguyen: I mean something like:

```swift
guard let status = Status(rawValue: statusValue) else { fatalError() }
return status
```

The question is whether returning `disabled` as a fallback could ever lead to issues. I can't really think of any though.

johnxnguyen (approved):

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -0,0 +1,46 @@
...
+   }
+    
+    @NSManaged public var name: String
+    @NSManaged public var config: Data?
+    @NSManaged private var statusValue: String
+    
+    public var status: Status {
+        get {
+            guard let status = Status(rawValue: statusValue) else { fatalError() }
+            return status
```
johnxnguyen, line 36: This was more intended as a topic of discussion, I'm not sure whether it is better to intentionally crash or provide some fallback.

johnxnguyen: If we do crash, then we should provide a message including the `statusValue` as it would be needed to know how to solve the issue should it occur.

KaterinaWire: Added logs

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -0,0 +1,50 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+@objcMembers
+public class Feature: ZMManagedObject {
+    
+   public enum Status: String, Codable {
```
marcomaddy1982, line 24: I like the idea to have an enum for the Status, even if it has (for now) just two value. 👍

marcomaddy1982 (approved):

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -0,0 +1,50 @@
...
+    @NSManaged public var config: Data?
+    @NSManaged private var statusValue: String
+    
+    public var status: Status {
+        get {
+            guard let status = Status(rawValue: statusValue) else {
+                zmLog.error("Failed to retrieve status: \(statusValue)")
+                fatalError()
+            }
+            return status
```
johnxnguyen, line 40: Instead of logging an error, pass a message to the fatalError:

```swift
fatalError("Failed to decode Feature statusValue: \(statusValue)")
```

This will allow us to see the message in the crash log (I think) so we wouldn't need to get the logs to check the value.

johnxnguyen (changes requested):

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: Done

johnxnguyen (approved):

johnxnguyen:
@KaterinaWire don't forget to delete this branch.

PR #1067, opened 2020-11-16
[Feature] Save Feature configs to CoreData

KaterinaWire:
### Issues
Save the feature configs to the database after retrieving them from the BE.

## Needs releases with:
https://github.com/wireapp/wire-ios-data-model/pull/1066


Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    func testThatItCreatesFeature() {
+        // given
+        let configData = try? JSONEncoder().encode(jsonToData(json: jsonConfig))
+        let feature = Feature.createOrUpdate("applock",
```
johnxnguyen, line 39: .This should be in "when"

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
```diff
@@ -0,0 +1,100 @@
...
+                                             status: "enabled",
+                                             config: configData,
+                                             context: uiMOC)
+        
+        // when
+        let fetchedFeature = Feature.createOrUpdate("applock",
```
johnxnguyen, line 45: This should be in "then", and you should ideally explicitly fetch it instead of relying on the same method that is being tested.

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
```diff
@@ -0,0 +1,100 @@
...
+        XCTAssertEqual(feature, fetchedFeature)
+    }
+    
+    func testThatItUpdatesFeature() {
+        // given
+        let configData = try? JSONEncoder().encode(jsonToData(json: jsonConfig))
+        let feature = Feature.createOrUpdate("applock",
+                                             status: "enabled",
+                                             config: configData,
+                                             context: uiMOC)
```
johnxnguyen, line 60: I would explicitly insert the object rather than using the method that is being tested.

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
KaterinaWire: Fixed

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
KaterinaWire: Fixed

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
KaterinaWire: Done

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -42,4 +42,44 @@ public class Feature: ZMManagedObject {
...
+        fetchRequest.predicate = NSPredicate(format: "name == %@", featureName)
+        fetchRequest.fetchLimit = 1
+        return context.fetchOrAssert(request: fetchRequest).first
+    }
+    
+    @discardableResult
+    public static func createOrUpdate(_ featureName: String,
+                                      status: Status,
+                                      config: Data?,
+                                      context: NSManagedObjectContext) -> Feature? {
```
johnxnguyen, line 60: You never return `nil`, so you might as well make the return type non optional.

KaterinaWire: Fixed

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
```diff
@@ -0,0 +1,97 @@
...
+                       "inactivity_timeout_secs": 30
+        ] as [String : Any]
+    
+    override func setUp() {
+        super.setUp()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+    }
```
johnxnguyen, line 34: These methods aren't overriding anything so can be removed.

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
```diff
@@ -0,0 +1,97 @@
...
+//
+
+import XCTest
+@testable import WireDataModel
+
+final class FeatureTest: ZMBaseManagedObjectTest {
+    
+    let jsonConfig = [ "enforce_app_lock": true,
+                       "inactivity_timeout_secs": 30
+        ] as [String : Any]
```
johnxnguyen, line 26: Instead of converting this to `Data` in each test, you could store it as data here:

```swift
let config: Data = {
  let json = """
  {
    "enforce_app_lock": true, 
    "inactivity_timeout_secs": 30
  }
  """

  return json.data(using: .utf8)!
}()
```

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
KaterinaWire: Removed

Tests/Source/Model/FeatureConfiguration/FeatureTest.swift
KaterinaWire: Done

johnxnguyen (approved):

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -47,4 +47,44 @@ public class Feature: ZMManagedObject {
...
+                             context: NSManagedObjectContext) -> Feature? {
+        
+        let fetchRequest = NSFetchRequest<Feature>(entityName: Feature.entityName())
+        fetchRequest.predicate = NSPredicate(format: "name == %@", featureName)
+        fetchRequest.fetchLimit = 1
+        return context.fetchOrAssert(request: fetchRequest).first
+    }
+    
+    @discardableResult
+    public static func createOrUpdate(_ featureName: String,
```
marcomaddy1982, line 62: I know that this is as it was when I've made this part :-) and I remember that I didn't liked the name here, cause actually we are `creating or updating and inserting`. So we are missing the `insert` action  in the method name. Of course this is just a detail and you are free to keep as it is but if you and @johnxnguyen will find a more descriptive name it would be perfect. 👌

marcomaddy1982 (approved):

marcomaddy1982:
@KaterinaWire and @johnxnguyen you already made a great job here. I approved. I left just a tiny comment that you could ignore cause it is a small detail.

Source/Model/FeatureConfig/Feature.swift
johnxnguyen: sn't it more accurate to say, that we are "fetching and updating, or creating and inserting"? I would say that "create" implies inserting a new object, and "update" implies fetching an existing object. For me createOrUpdate seems like an appropriate name. Or am I missing something?

marcomaddy1982: Sounds Good for me as well!!!

johnxnguyen:
@KaterinaWire don't forget to delete this branch.

PR #1068, opened 2020-11-16
[Feature]: EAR and export backup

marcomaddy1982:
## What's new in this PR?

In the version of the app without the `EAR` when a user start a backup the following steps are executed:

1. User initiates backup from the settings
2. Enter password for the backup
3. We generate the backup
4. Copy local store
5. Put the backup into an encrypted container using the password entered in step 2

with the introduction of the `EAR` and when it is enabled there is an additional layer of encryption on the database. When the user will create a new backup, if the `EAR` is enabled we have first `decrypt` the `EAR` before put the backup into an encrypted container using the password he/she previously entered. This mean that with the `EAR` we will have the following steps when a backup is created

1. User initiates backup from the settings
2. Enter password for the backup
3. We generate the backup
4. Copy local store
5. **Decrypt local store if necessary (encryption keys should already be available)**
6. Put backup into an encrypted container using the password entered in step 2


Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -137,13 +141,26 @@ extension StorageStack {
...
         let store = try coordinator.addPersistentStore(ofType: NSSQLiteStoreType, configurationName: nil, at: location, options: options)
         let context = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
         context.persistentStoreCoordinator = coordinator
         var storeMetadata = store.metadata
         
+        if let encryptionKeys = encryptionKeys {
+            context.performAndWait {
+                do {
+                    try context.disableEncryptionAtRest(encryptionKeys: encryptionKeys)
+                } catch {
```
typfel, line 158: You should not catch error because we want it to bubble up so that whole backup operation fails.

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -137,13 +141,26 @@ extension StorageStack {
...
         // Add persistent store at the new location to allow creation of NSManagedObjectContext
         let store = try coordinator.addPersistentStore(ofType: NSSQLiteStoreType, configurationName: nil, at: location, options: options)
         let context = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
         context.persistentStoreCoordinator = coordinator
         var storeMetadata = store.metadata
         
+        if let encryptionKeys = encryptionKeys {
+            context.performAndWait {
+                do {
+                    try context.disableEncryptionAtRest(encryptionKeys: encryptionKeys)
```
typfel, line 157: How about doing this only if `context.encryptMessagesAtRest` is `true`?

Source/ManagedObjectContext/ManagedObjectContextDirectory+EncryptionAtRest.swift
```diff
@@ -39,6 +39,10 @@ public extension ManagedObjectContextDirectory {
         return encryptionKeys
     }
 
+    func getEncryptionKeys() throws -> EncryptionKeys {
```
typfel, line 42: I don't we need another accessor for the encryption keys since `uiContext` is a public property and exposing like this is inviting users to use it on the wrong context (It will crash unless used from the main context).

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -121,7 +122,10 @@ extension StorageStack {
                 )
 
                 // Mark the database as imported from a history backup
-                try markAsImported(coordinator: coordinator, location: backupLocation, options: options)
+                try markAsImported(coordinator: coordinator,
```
typfel, line 125: I think this name no longer describing what the method does. Maybe something like `prepareStoreForBackup()`?

typfel (changes requested):

Source/ManagedObjectContext/StorageStack+Backup.swift
johnxnguyen: Additionally, we should probably throw an error when EAR is enabled but encryption keys were not passed to the method. There's no point creating a backup if we can't decrypt it first.

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -83,6 +83,7 @@ extension StorageStack {
         clientIdentifier: String,
         applicationContainer: URL,
         dispatchGroup: ZMSDispatchGroup? = nil,
+        encryptionKeys: EncryptionKeys? = nil,
```
johnxnguyen, line 86: Don't forget to document the new argument.

johnxnguyen: Also... just wondering why we need to pass the encryption keys as an argument?

johnxnguyen (changes requested):

Source/ManagedObjectContext/StorageStack+Backup.swift
marcomaddy1982: Cause we don't have here direct access to the `activeUserSession`, so we retrieve them in the SE (`SessionManager+Backup`) and we inject here. Do you have a better way in mind?

Source/ManagedObjectContext/StorageStack+Backup.swift
marcomaddy1982: I will add an `throw` error for the case you mentioned but right before the start of the backup.

marcomaddy1982:
@typfel @johnxnguyen I guess I have addressed all your comments. Thanks 👍

Source/ManagedObjectContext/StorageStack+Backup.swift
typfel: > I will add an `throw` error for the case you mentioned but right before the start of the backup.

Hmm why not do check here? Then you also don't need to expose the `getEncryptionKeys()` since then you could access the encryption keys directly on the context `uiContext.encryptionsKeys`.

Source/ManagedObjectContext/NSManagedObjectContext+EncryptionAtRest.swift
```diff
@@ -245,7 +245,7 @@ extension NSManagedObjectContext {
         get { userInfo[Self.encryptionKeysUserInfoKey] as? EncryptionKeys }
     }
     
-    func getEncryptionKeys() throws -> EncryptionKeys {
+    public func getEncryptionKeys() throws -> EncryptionKeys {
```
typfel, line 248: Doesn't need to be public anymore

typfel (changes requested):

typfel:
We are still missing tests

Source/ManagedObjectContext/StorageStack+Backup.swift
johnxnguyen: I thought you could get the encryption keys from the context?

typfel: @johnxnguyen the only context available here is a temporary context we create for the copied local store, since we create the context here it won't have the encryption keys set.

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -137,13 +143,26 @@ extension StorageStack {
...
+                                              options: [String: Any],
+                                              encryptionKeys: EncryptionKeys? = nil) throws {
         // Add persistent store at the new location to allow creation of NSManagedObjectContext
         let store = try coordinator.addPersistentStore(ofType: NSSQLiteStoreType, configurationName: nil, at: location, options: options)
         let context = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
         context.persistentStoreCoordinator = coordinator
         var storeMetadata = store.metadata
         
+        if context.encryptMessagesAtRest {
+            if let encryptionKeys = encryptionKeys {
```
johnxnguyen, line 157: A `guard` would make this a bit more readable:

```swift
guard let encryptionKeys = encryptionKeys else { throw BackupError.missingEAREncryptionKey }
```

marcomaddy1982:
> We are still missing tests

Done

Source/ManagedObjectContext/StorageStack+Backup.swift
marcomaddy1982: Done

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
```diff
@@ -130,7 +133,90 @@ class StorageStackTests_Backup: DatabaseBaseTest {
...
+        let databaseDirectory = url.appendingPathComponent("data")
+        let metadataURL = url.appendingPathComponent("export.json")
+        
+        XCTAssertTrue(fm.fileExists(atPath: databaseDirectory.path))
+        XCTAssertTrue(fm.fileExists(atPath: metadataURL.path))
+        XCTAssertTrue(try fm.contentsOfDirectory(atPath: databaseDirectory.path).count > 1)
+        XCTAssertTrue(try fm.contentsOfDirectory(atPath: url.path).count > 1)
+    }
+    
+    func testThaItNotFindsTheStorage_WhenEARIsEnabledAndEncryptionKeysAreNil() throws {
```
typfel, line 193: Suggestion: `testThatItFailsWhenEARIsEnabledAndEncryptionKeysAreNil`

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
```diff
@@ -130,7 +133,90 @@ class StorageStackTests_Backup: DatabaseBaseTest {
...
+        let databaseDirectory = url.appendingPathComponent("data")
+        let metadataURL = url.appendingPathComponent("export.json")
+        
+        XCTAssertTrue(fm.fileExists(atPath: databaseDirectory.path))
+        XCTAssertTrue(fm.fileExists(atPath: metadataURL.path))
+        XCTAssertTrue(try fm.contentsOfDirectory(atPath: databaseDirectory.path).count > 1)
+        XCTAssertTrue(try fm.contentsOfDirectory(atPath: url.path).count > 1)
+    }
+    
+    func testThaItFindsTheStorage_WhenEARIsDisabledAndEncryptionKeysAreNil() throws {
```
typfel, line 165: Isn't this already coved by `testThatItFindsTheStorageWithCorrectAccountIdentifier`?

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
```diff
@@ -130,7 +133,90 @@ class StorageStackTests_Backup: DatabaseBaseTest {
         default: XCTFail("unexpected error type")
         }
     }
+    
+    func testThatItFindsTheStorage_WhenEARIsEnableAndEncryptionKeysAreValid() throws {
```
typfel, line 137: We are not testing that the backup has encryption at rest disabled. I think we need to open the backup and verify that `encryptMessagesAtRest` is `false`. 

You can look at `testThatItMarksStoreAsBackupWhenExporting` for how this is done.

typfel: Also if we are testing that EAR was disabled int the backup I would change the name to something like:
`testThatItDisablesEncryptionAtRest_WhenEARIsEnableAndEncryptionKeysAreValid`

johnxnguyen (approved):

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
```diff
@@ -146,6 +232,28 @@ class StorageStackTests_Backup: DatabaseBaseTest {
         let fetchConversations = ZMConversation.sortedFetchRequest()
         XCTAssertEqual(try directory.uiContext.count(for: fetchConversations), 1)
     }
+    
+    func testThatItPreservesOriginalDataAfterBackup_WhenEARIsEnableAndEncryptionKeysAreValid() {
```
typfel, line 236: Do we really need this test? I think `testThatItPreservesOriginalDataAfterBackup` is enough, since it's testing that we operate on a local store copy and that also applies for the EAR case.

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
```diff
@@ -164,6 +272,26 @@ class StorageStackTests_Backup: DatabaseBaseTest {
         let fetchConversations = ZMConversation.sortedFetchRequest()
         XCTAssertEqual(try anotherDirectory.uiContext.count(for: fetchConversations), 1)
     }
+    
+    func testThatItPreservesOriginaDataAfterBackupIfStackIsNotActive_WhenEARIsDisabledAndEncryptionKeysAreNil() throws {
```
typfel, line 276: Do we really need this test? See previous comment.

typfel (changes requested):

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
marcomaddy1982: Yes Right that actually it is the same scenario. I will remove this one.

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
marcomaddy1982: Why not, I mean we are testing that after exporting a db with EAR enabled no weird stuffs are happening in the original data. In the `testThatItPreservesOriginalDataAfterBackup` the EAR is disable so we are not exactly in the came condition.  Anyway if you prefer we can remove.

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
marcomaddy1982: See previous comment.

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
typfel: It's true that is not exactly the same condition but for this test to fail we would need to open and modify or delete  original local store which is also covered by first test.

I'm very cautious about having tests which cover the same scenario because it:
- Makes it painful when changing the behaviour of the code, multiple tests break when one behaviour is changed.
- Makes it unclear what is broken because multiple tests will always break together.

marcomaddy1982: Ok, I got your point. Makes sense. I'm going to remove them.

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
marcomaddy1982: Done.

Tests/Source/ManagedObjectContext/StorageStackTests_Backup.swift
marcomaddy1982: Done.

typfel (approved):

PR #1069, opened 2020-11-19
[Fix]: User is logged out when a backup is imported

marcomaddy1982:
### Issues

the following ticket is related to the scenario described [here](https://wearezeta.atlassian.net/browse/SQCORE-205) and [here](https://github.com/wireapp/wire-ios/issues/4717)

### Causes

In the current version when a backup is exported the db it is marked (see `func markAsImported`). Basically under the key  `PersistentMetadataKey.importedFromBackup`  a  `bool` is saved as `true` in its persistent metadata of the data base.
Often On iOS14 when the backup is imported  the `bool` stored in the metadata of the database under the `PersistentMetadataKey.importedFromBackup` key is corrupted. Somehow the value not present in the persistent metadata.

P.s :Till now  we didn't understand why the PersistentMetadataKey.importedFromBackup is missing in the metadata.

### Solutions

The following solution is not the cleanest: we added a `bool` `isImportedFromBackup` in the `StorageStack` singleton and we set to `true` when a backup is imported. In this manner we don't relay anymore to to data base metadata that could be corrupted but we relay just on the user flow.  In this manner also old corrupted backup could be loaded with the user is logged out.

### DIFF
--- a/Source/ManagedObjectContext/NSManagedObjectContext+BackupImport.swift
+++ b/Source/ManagedObjectContext/NSManagedObjectContext+BackupImport.swift
@@ -22,9 +22,9 @@ extension NSManagedObjectContext {
     
     /// Prepare a backed up database for being imported, deleting self client, push token etc.
     func prepareToImportBackup() {
+        StorageStack.shared.isImportedFromBackup = false
         require(self.zm_isSyncContext, "Needs to be run on Sync Context to avoid race conditions")
         setPersistentStoreMetadata(nil as Data?, key: ZMPersistedClientIdKey)
-        setPersistentStoreMetadata(nil as Data?, key: PersistentMetadataKey.importedFromBackup.rawValue)
         setPersistentStoreMetadata(nil as Data?, key: PersistentMetadataKey.pushToken.rawValue)
         setPersistentStoreMetadata(nil as Data?, key: PersistentMetadataKey.pushKitToken.rawValue)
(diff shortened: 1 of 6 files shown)


typfel:
We will need to modify the store when importing a backup when EAR is enabled by default, I think it will therefore be cleaner to perform the `prepareToImportBackup()` inside `importLocalStorage()`. That way we also don't set the flag when creating a backup and we don't rely on a static boolean.

marcomaddy1982:
> We will need to modify the store when importing a backup when EAR is enabled by default, I think it will therefore be cleaner to perform the `prepareToImportBackup()` inside `importLocalStorage()`. That way we also don't set the flag when creating a backup and we don't rely on a static boolean.

Your suggestion it is a cleaner solution and I thought to it yesterday too but in the method `importLocalStorage()` we still don't have the `ManagedObjectContextDirectory` and then the `context` to `prepareToImportBackup`.

typfel:
> Your suggestion it is a cleaner solution and I thought to it yesterday too but in the method `importLocalStorage()` we still don't have the `ManagedObjectContextDirectory` and then the `context` to `prepareToImportBackup`.

I'll open a PR with this solution.

marcomaddy1982:
> > Your suggestion it is a cleaner solution and I thought to it yesterday too but in the method `importLocalStorage()` we still don't have the `ManagedObjectContextDirectory` and then the `context` to `prepareToImportBackup`.
> 
> I'll open a PR with this solution.

Ok, thanks.

PR #1070, opened 2020-11-19
[Fix] Metadata key not always being persisted on backup

typfel:
### Issues

Metadata key not always being persisted on backup leading to issues when importing the backup.

### Causes

The store files still connected to a `NSPersistentStore` when being copied.

### Solutions

Remove `NSPersistentStore` from the `NSPersistentStoreCoordinator` before copying the store files.

### DIFF
--- a/Source/ManagedObjectContext/StorageStack+Backup.swift
+++ b/Source/ManagedObjectContext/StorageStack+Backup.swift
@@ -152,6 +152,9 @@ extension StorageStack {
         try context.performGroupedAndWait { context in
             try context.save()
         }
+        
+        // Close the store, not doing so could lead to data loss when copying the store files.
+        try coordinator.remove(store)
     }
     
     public enum BackupImportError: Error {


marcomaddy1982 (approved):

johnxnguyen:
Interesting, probably this was the cause of the issue I had a little while ago where my backup was not marked as imported. But I guess this issue has been around for a long time? Do you think this is iOS14 related since we're only seeing issues recently?

typfel:
> Interesting, probably this was the cause of the issue I had a little while ago where my backup was not marked as imported. But I guess this issue has been around for a long time? Do you think this is iOS14 related since we're only seeing issues recently?

Yes, I we suspect something have changed in iOS 14 which makes this bug happen more frequently.

johnxnguyen (approved):

typfel:
This PR was made obsolete by https://github.com/wireapp/wire-ios-data-model/pull/1071

PR #1071, opened 2020-11-20
[Fix] Logout when importing backup

typfel:
### Issues

Users would sometimes get logged after importing a backup.

### Causes

Due to a bug when exporting a backup the flag `importedFromBackup` was not sometimes not included into store metadata. When importing a backup we were deleting metadata related to the users's client if this flag was was present in the backup store.

### Solutions

Always delete the metadata when importing a backup and don't rely on the `importedFromBackup` flag.


marcomaddy1982 (approved):

Source/ManagedObjectContext/StorageStack+Backup.swift
```diff
@@ -210,8 +208,8 @@ extension StorageStack {
                     sourceOptions: options,
                     ofType: NSSQLiteStoreType
                 )
-
-                log.info("succesfully imported backup with metadata: \(metadata)")
+                
+                log.info("successfully imported backup with metadata: \(metadata)")
```
johnxnguyen, line 212: Just wondering, do we actually see info logs? I have a feeling we only see debug and error tags.

johnxnguyen (approved):

Source/ManagedObjectContext/StorageStack+Backup.swift
typfel: The order should be: debug, info, warn, error. So they should be visible unless we have a bug in the logging system :)

typfel:
@zenkins release patch

PR #1073, opened 2020-11-20
[Fix] We were not using newValue and therfore not updating the value

typfel:
### Issues

We were getting build warring that weren't accessing the `newValue` in the setters of `activationLocationLatitude` and `activationLocationLatitude `.

This would always update `activationLocationLatitude` and `activationLocationLatitude ` with the old value.

### Solutions

Use `newValue`.


KaterinaWire (approved):

billypchan (approved):

typfel:
@zenkins release patch

PR #1075, opened 2020-11-24
[Refactor] Feature configuration

johnxnguyen:
## What's new in this PR?

This PR contains some refactoring that aims to simplify and clarify the usage of feature configurations. 

---
### Issue 
Feature instances are identified with a `String` which is error prone.

### Solution 
Define `Feature.Name` enum to give `name` a strong type.

---
### Issue 
Fetching features from the database is cumbersome (requires a fetch request and context).

### Solution 
Features are associated with a team, so add a relationship to team. The teams features can now be accessed through the self user, like: `selfUser.team?.feature(for: Feature.AppLock.self)`.

---

### Issue
Instances of `Feature` are not suitable to pass around because the `config` property is just `Data` and needs encoding.

### Solution:
Create an abstraction over `Feature` that provides stronger types. `FeatureLike` replaces the `Named` and `Configurable` protocols from wire-request-strategy.


KaterinaWire (approved):

Tests/Source/Model/FeatureConfiguration/FeatureLikeTests.swift
```diff
@@ -0,0 +1,50 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import XCTest
+@testable import WireDataModel
+
+class FeatureLikeTests: ZMBaseManagedObjectTest {
+
+    func testThatItStoresAFeatureLikeObject() throws {
```
KaterinaWire, line 24: Should we add a test to check if the team is nil then it won't save the feature?

johnxnguyen: But we pass the team as an argument to store. The test that you describe would be on the feature controller or feature config request strategy, since that is where the team object is passed from.

KaterinaWire: Ok, I will check in request strategy do we have such test.

KaterinaWire (approved):

marcomaddy1982 (approved):

PR #1076, opened 2020-11-25
[Fix] Exporting backups under EAR

typfel:
### Issues

Backups exported when EAR is enabled also have EAR enabled when imported which is not the expected behaviour.

### Causes

When we create a backup we disable EAR for the backup but we were missing to persist the store metadata so the `encryptMessagesAtRest` flag was still set to `true` even though EAR had been disabled.

### Solutions

Persist store metadata.


marcomaddy1982 (approved):

johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1077, opened 2020-11-26
[Refactor] Add support for downstream object sync

johnxnguyen:
### Issues

We want to use a downstream object sync to fetch features from the backend and store them locally. The `Feature` entity does not yet support this capability.

### Causes

There are some missing implementations for `Feature`.

### Solutions

- Add `needsToBeUpdatedFromBackend` property
- Define sort key
- Fix relationship from team to feature (it was to one and not to many).
- Provide mechanism to create default instances when needed.


Source/Model/FeatureConfig/Feature.AppLock.swift
```diff
@@ -62,7 +66,7 @@ public extension Feature {
 
             public let inactivityTimeoutSecs: UInt
 
-            public init(enforceAppLock: Bool, inactivityTimeoutSecs: UInt) {
+            public init(enforceAppLock: Bool = false, inactivityTimeoutSecs: UInt = 60) {
```
KaterinaWire, line 69: Are default values needed here? We pass some  default values  in the line #47.

johnxnguyen: Yep good point. I'll simplified the init on line 47 to use the default values provided here.

KaterinaWire (approved):

marcomaddy1982 (approved):

PR #1082, opened 2020-12-03
[Refactor] Modify enqueueBackendRefresh() to refresh  features

KaterinaWire:
### Issues

We currently have an `enqueueBackendRefresh()` method that can trigger a BE to update the feature. Since we don't have any feature in the new entity of Core data, we should create a new default  feature and then trigger BE to update it.

### DIFF
--- a/Source/Model/Conversation/Team+Feature.swift
+++ b/Source/Model/Conversation/Team+Feature.swift
@@ -64,11 +64,19 @@ extension Team {
     /// To trigger the request immediately, post a `RequestAvailableNotification`.
     ///
     /// - Parameter name: The name of the feature to refresh.
-
+    
     public func enqueueBackendRefresh(for name: Feature.Name) {
         guard let context = managedObjectContext else { return }
-        let feature = Feature.fetch(name: name, context: context)
-        feature?.needsToBeUpdatedFromBackend = true
+        
+        if let feature = Feature.fetch(name: name, context: context) {
+            feature.needsToBeUpdatedFromBackend = true
+        } else {
+            switch name {
+            case .appLock:
+                let defaultInstance = try? Feature.AppLock().store(for: self, in: context)
+                defaultInstance?.needsToBeUpdatedFromBackend = true
+            }
+        }
     }
 
 }
--- a/Source/Model/Conversation/Team+Feature.swift
+++ b/Source/Model/Conversation/Team+Feature.swift
@@ -68,15 +68,8 @@ extension Team {
     public func enqueueBackendRefresh(for name: Feature.Name) {
         guard let context = managedObjectContext else { return }
         
-        if let feature = Feature.fetch(name: name, context: context) {
-            feature.needsToBeUpdatedFromBackend = true
-        } else {
-            switch name {
-            case .appLock:
-                let defaultInstance = try? Feature.AppLock().store(for: self, in: context)
-                defaultInstance?.needsToBeUpdatedFromBackend = true
-            }
-        }
+        let feature = Feature.fetchOrCreate(name: name, team: self, context: context)
+        feature.needsToBeUpdatedFromBackend = true
     }
 
 }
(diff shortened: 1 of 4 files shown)


Source/Model/Conversation/Team+Feature.swift
```diff
@@ -64,11 +64,19 @@ extension Team {
...
         guard let context = managedObjectContext else { return }
-        let feature = Feature.fetch(name: name, context: context)
-        feature?.needsToBeUpdatedFromBackend = true
+        
+        if let feature = Feature.fetch(name: name, context: context) {
+            feature.needsToBeUpdatedFromBackend = true
+        } else {
+            switch name {
+            case .appLock:
+                let defaultInstance = try? Feature.AppLock().store(for: self, in: context)
```
typfel, line 76: Wouldn't we need to somehow associate this is new `Feature` instance with `applock`? For example If I call `enqueueBackendRefresh(for .applock)` twice It would create two default instances which I assume is not good.

typfel (changes requested):

Source/Model/Conversation/Team+Feature.swift
KaterinaWire: The default instance will only be for the `applock` feature (`Feature.AppLock()`).
After our call, I changed the `store(for: self, in: context)` method. This method is now just adding a new feature to Core Data.

typfel: I thought we would remove the `if else` here since this logic is already in `Feature.createOrUpdate()`

KaterinaWire: I noticed that `Feature.createOrUpdate() `method updates the existing feature with default value or creates a new one.
If I remove ` if else` and keep like this:
```
switch name {
            case .appLock:
                let defaultInstance = try? Feature.AppLock().store(for: self, in: context)
                defaultInstance?.needsToBeUpdatedFromBackend = true
            }
```
It will override the existing feature (or create a new one) with a default value and mark it as `needsToBeUpdatedFromBackend`.
I don't like this idea to override the existing feature.

typfel: Ok makes sense. I'll approve this now, maybe in a future PR it makes sense to add a method like `Feature.fetchOrCreate(name:context)` which can do this logic in a generic way without switching over all features.

typfel (approved):

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -96,6 +96,32 @@ public class Feature: ZMManagedObject {
...
+    static func fetchOrCreate(name: Name,
+                              team: Team,
+                              context: NSManagedObjectContext) -> Feature {
+        
+        let fetchRequest = NSFetchRequest<Feature>(entityName: Feature.entityName())
+        fetchRequest.predicate = NSPredicate(format: "nameValue == %@", name.rawValue)
+        fetchRequest.fetchLimit = 2
+        
+        let results = context.fetchOrAssert(request: fetchRequest)
+        require(results.count <= 1, "More than instance for feature: \(name.rawValue)")
```
David-Henner, line 110:
```suggestion
        require(results.count <= 1, "More than one instance for feature: \(name.rawValue)")
```

Source/Model/Conversation/Team+Feature.swift
KaterinaWire: @typfel Yes, good point. I've made changes to this PR after your approval.

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -96,6 +96,32 @@ public class Feature: ZMManagedObject {
...
+        guard let feature = results.first else {
+            switch name {
+            case .appLock:
+                let defaultInstance = Feature.AppLock()
+                let feature = insert(name: name,
+                                     status: defaultInstance.status,
+                                     config: try? JSONEncoder().encode(defaultInstance.config),
```
David-Henner, line 117: should there be some error handling in case of encoding errors?

KaterinaWire: Yes, make sense.
Done

David-Henner (approved):

PR #1084, opened 2020-12-07
[Refactor] Provide a screen to create a custom passcode

KaterinaWire:
## What's new in this PR?

According to the design flow
`(https://zpl.io/bArgMpp or zpl://screen?sid=5fbbd5b02956135e8e9f3849&pid=5f0f1143f783828d19da7721)`

user should try to use biometrics to unlock Wire. If there are no biometrics or phone passcode, we should provide a screen to create a custom passcode.

**Important:** 
In this PR I renamed `.needAccountPassword` with `.needCustomPasscode` since we don't need anymore unlock Wire with account password. I got an approval from Eva, customers will be informed. 
According to the design flow, we can unlock Wire using biometrics (if available) or a custom passcode.

### DIFF
--- a/Source/Model/FeatureConfig/AppLock/AppLockController.swift
+++ b/Source/Model/FeatureConfig/AppLock/AppLockController.swift
@@ -27,7 +27,8 @@ public protocol AppLockType {
     var isCustomPasscodeNotSet: Bool { get }
     var config: AppLockController.Config { get }
     
-    func evaluateAuthentication(description: String,
+    func evaluateAuthentication(scenario: AppLockController.AuthenticationScenario,
+                                description: String,
                                 with callback: @escaping (AppLockController.AuthenticationResult, LAContext) -> Void)
     func persistBiometrics()
 }
@@ -93,7 +94,8 @@ public final class AppLockController: AppLockType {
     // MARK: - Methods
     
     // Creates a new LAContext and evaluates the authentication settings of the user.
-    public func evaluateAuthentication(description: String,
+    public func evaluateAuthentication(scenario: AppLockController.AuthenticationScenario,
+                                       description: String,
                                        with callback: @escaping (AuthenticationResult, LAContext) -> Void) {
         
         guard self.weakLAContext == nil else { return }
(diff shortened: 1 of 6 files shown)


Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -94,38 +93,29 @@ public final class AppLockController: AppLockType {
     // MARK: - Methods
     
     // Creates a new LAContext and evaluates the authentication settings of the user.
-    public func evaluateAuthentication(scenario: AuthenticationScenario,
```
typfel, line 97: `AuthenticationScenario` is not necessary anymore? How do you now deal with the fact that when encryption at rest is enabled we cannot fallback to asking the user for an account password since only biometrics will unlock the database.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -110,24 +111,16 @@ public final class AppLockController: AppLockType {
             callback(.needAccountPassword, context)
             return
         }
-        
-        if canEvaluatePolicy {
```
David-Henner, line 114: could you clarify why this isn't necessary anymore?

Source/Model/FeatureConfig/AppLock/AppLockController.swift
typfel: Since you revived the AuthenticationScenario I'm resolving this.

typfel:
@KaterinaWire is this ready to be reviewed again?

Source/Model/FeatureConfig/AppLock/AppLockController.swift
KaterinaWire: I updated this PR. 
You are right, we need it.

KaterinaWire:
> @KaterinaWire is this ready to be reviewed again?

Yes. I updated the PR description.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -163,47 +161,40 @@ public final class AppLockController: AppLockType {
         case denied
         /// There's no authenticated method available (no passcode is set)
         case unavailable
-        /// Biometrics failed and account password is needed instead of device PIN
-        case needAccountPassword
```
typfel, line 167: So we are removing support for account password fallback?

typfel (approved):

Source/Model/FeatureConfig/AppLock/AppLockController.swift
typfel: Katerina:  "Yes"

David-Henner (approved):

PR #1085, opened 2020-12-07
[Refactor] Needs to inform user that feature has been changed

KaterinaWire:
## What's new in this PR?

According to the design flow
(https://zpl.io/bArgMpp or zpl://screen?sid=5fbbd5b02956135e8e9f3849&pid=5f0f1143f783828d19da7721)

we should provide a warning screen that feature has been changed. 
I added a `needToInformUserOfFeatureChange` flag to the Feature. This value would be set to `true` when we know there has been a change, and it would be set to `false` when the user has acknowledged the change.


typfel:
I'm missing the part where you set the `needToInformUserOfFeatureChange` to `true`.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -77,6 +79,16 @@ public final class AppLockController: AppLockType {
         return Keychain.fetchPasscode() == nil
     }
     
+    public var needsToInformUserOfFeatureChange: Bool {
```
marcomaddy1982, line 82: Could we call this just something like `isFeatureChanged`?  i mean i don't think that is so important to who we are informing :-).

typfel: We have  many properties in data model named like`needsTo...`, like: 
- `needsToBeUpdatedFromBackend`
- `needsToNotifyUser`
- `needsToUploadSignalingKeys`

So I think the naming follows existing conventions in this case.

marcomaddy1982: Ok, thanks for the clarification.

KaterinaWire:
> I'm missing the part where you set the `needToInformUserOfFeatureChange` to `true`.

@typfel In this PR: https://github.com/wireapp/wire-ios-request-strategy/pull/262

Source/Model/FeatureConfig/AppLock/AppLockController.swift
KaterinaWire: I renamed to `needsToNotifyUser`.

KaterinaWire:
> > I'm missing the part where you set the `needToInformUserOfFeatureChange` to `true`.
> 
> @typfel In this PR: [wireapp/wire-ios-request-strategy#262](https://github.com/wireapp/wire-ios-request-strategy/pull/262)

@typfel I closed the PR in the request strategy and add a `didSet` block to the `configData` property.

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -43,10 +43,26 @@ public class Feature: ZMManagedObject {
...
+        didSet {
+            guard let oldValue = oldValue,
+                let newValue = configData else {
+                    return
+            }
+            let decoder = JSONDecoder()
+            let oldConfig = try? decoder.decode(Feature.AppLock.Config.self, from: oldValue)
+            let newConfig = try? decoder.decode(Feature.AppLock.Config.self, from: newValue)
+            if oldConfig?.enforceAppLock != newConfig?.enforceAppLock {
+                needsToNotifyUser = true
```
marcomaddy1982, line 61: this could just be 

`needsToNotifyUser = oldConfig?.enforceAppLock != newConfig?.enforceAppLock`

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -43,10 +43,26 @@ public class Feature: ZMManagedObject {
 
     @NSManaged private var nameValue: String
     @NSManaged private var statusValue: String
-    @NSManaged public var configData: Data?
+    @NSManaged var needsToNotifyUser: Bool
```
marcomaddy1982, line 46: so does it to be `internal` now?

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -43,10 +43,26 @@ public class Feature: ZMManagedObject {
...
+    
+    @objc
+    public var configData: Data? {
+        didSet {
+            guard let oldValue = oldValue,
+                let newValue = configData else {
+                    return
+            }
+            let decoder = JSONDecoder()
+            let oldConfig = try? decoder.decode(Feature.AppLock.Config.self, from: oldValue)
```
typfel, line 58: You can't assume that this is always going to be a `.applock` config. Perhaps one option would be to switch on the `name` and delegate to a static method on the `AppLock` struct.

`static func needsToNotifyUserOnConfigUpdate(old: Config, new: Config) -> Bool`

In my previous comment on the request-strategy PR  I missed that this is not the decoded config so it's a bit unfortunate that we have to encode & decode the config so many times.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -25,6 +25,7 @@ public protocol AppLockType {
     var isActive: Bool { get set }
     var lastUnlockedDate: Date { get set }
     var isCustomPasscodeNotSet: Bool { get }
+    var needsToNotifyUserOfFeatureChange: Bool { get set }
```
typfel, line 28: Should we also rename this to `needsToNotifyUser`?

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -144,6 +161,7 @@ public final class AppLockController: AppLockType {
         public var forceAppLock: Bool
         public var appLockTimeout: UInt
         public var isAvailable: Bool
+        public var needsToNotifyUserOfFeatureChange: Bool
```
typfel, line 164: Couldn't you remove this property? Seems like `AppLockController. needsToNotifyUserOfFeatureChange` reads the value directly from `Feature.needsToNotifyUser`.

typfel (changes requested):

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: I think we should set  `needsToNotifyUser ` to `false` only when the user confirms it from the UI.
A case is possible when there will be 2 requests one after another in a short period of time (before the user is informed). The first time it will set the `true` and the second time it will set `false`. So you can lose the "state" of informing the user.

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -43,10 +43,26 @@ public class Feature: ZMManagedObject {
...
     @NSManaged private var nameValue: String
     @NSManaged private var statusValue: String
-    @NSManaged public var configData: Data?
+    @NSManaged var needsToNotifyUser: Bool
 
     @NSManaged public var team: Team?
-
+    
+    @objc
+    public var configData: Data? {
```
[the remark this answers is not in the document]
KaterinaWire: `@NSManaged` not allowed on observing properties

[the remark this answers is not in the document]
typfel: So this approach doesn't work then?

Source/Model/FeatureConfig/AppLock/AppLockController.swift
KaterinaWire: Done

Source/Model/FeatureConfig/AppLock/AppLockController.swift
KaterinaWire: Yes.
Removed

Source/Model/FeatureConfig/Feature.swift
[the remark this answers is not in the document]
KaterinaWire: It works. I have a test to check it:
`testThatItUpdatesNeedsToNotifyUserFlag_IfEnforceAppLockHasBeenChanged() `

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: Yes. We will not directly get or set this value from the UI.
We will set this value from the `didSet` block of the `configData` property. 
And to get this value, `AppLockController` has `needsToNotifyUser`.

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: Yes, good point, this should be applied only for `.applock`. I added a switch.
But I have doubts about the static method:
1.  We will only check for changes in the `enforceAppLock `.
2. I think we should only set `needsToNotifyUser` to `true`. `False` should be set from the UI when the user confirms it.

typfel:
```
var updateConfig: Config {
    didSet {
          guard !needsToNotifyUser else { return }
          
          var notify
          switch name {
                case .appLock:
                notify = AppLock.needsToNotifyUserOnConfigUpdate(old: oldConfig, new: newConfig) 
                default:
                notify = false
          }

          needsToNotifyUser = notify
    }
}

static func needsToNotifyUserOnConfigUpdate(old: Config, new: Config) -> Bool {
     return old.enforceAppLock != new.enforceAppLock
}
```

This is essentially the same, I think it would be nice though if feature specific behaviour is kept in their respective class/struct.

Source/Model/FeatureConfig/AppLock/Feature.AppLock.swift
```diff
@@ -24,10 +24,27 @@ public extension Feature {
...
+                }
+                return feature.needsToNotifyUser
+            }
+            set {
+                guard let context = Self.managedObjectContext else {
+                        return
+                }
+                update(needsToNotifyUser: newValue, in: context)
+            }
+        }
```
johnxnguyen, line 47: Why do we need this setter? Is this how we would set the value to `false` when the user acknowledges the change?

KaterinaWire: I updated this PR after our call

typfel (approved):

Tests/Source/Model/FeatureConfiguration/FeatureTests.swift
```diff
@@ -100,4 +95,44 @@ final class FeatureTests: ZMBaseManagedObjectTest {
         // then
         XCTAssertNotNil(fetchedFeature)
     }
+    
+    func testThatItUpdatesNeedsToNotifyUserFlag_IfEnforceAppLockHasBeenChanged() {
```
johnxnguyen, line 99: We should test also the other way, that enforce app lock was true then changed to false.

johnxnguyen (changes requested):

Tests/Source/Model/FeatureConfiguration/FeatureTests.swift
KaterinaWire: Done

johnxnguyen (approved):

PR #1086, opened 2020-12-08
[Refactor] Keychain migration for applock

KaterinaWire:
### Issues

1. Currently we store a `YES` or `NO` string inside the keychain to represent whether app lock is active or not. The value is stored using a single key. Now that app lock is per-account, we need to update the key to consider the self user,  by including the user id as part of the key.

### Solutions

Store custom passcode and `isActive` status for applock per-account. We also should add a migration step that will check if a value exists in the key chain using the legacy key, and replace it with values stored at the new key(s).


Source/ManagedObjectContext/NSManagedObjectContext+AppLock.swift
```diff
@@ -0,0 +1,34 @@
...
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension NSManagedObjectContext {
+    
+    /// Whether the applock feature is enabled.
+    
+    internal(set) public var appLock: Bool {
```
typfel, line 25: Let's name this so that it's clear hat we are dealing with a boolean value, for example `isAppLockEnabled`

Source/Model/User/ZMUser+Applock.swift
```diff
@@ -0,0 +1,30 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public extension ZMUser {
+    var isAppLockActive: Bool {
+        get {
+            return self.managedObjectContext?.appLock ?? false
```
typfel, line 24:
```suggestion
            return managedObjectContext?.appLock ?? false
```

Source/Model/User/ZMUser+Applock.swift
```diff
@@ -0,0 +1,30 @@
...
+        }
+        set {
+            self.managedObjectContext?.appLock = newValue
```
typfel, line 27:
```suggestion
            managedObjectContext?.appLock = newValue
```

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -235,3 +223,30 @@ public class BiometricsState {
         lastPolicyDomainState = currentPolicyDomainState
     }
 }
+
+// MARK: - Migration rules
+
+extension AppLockController {
```
typfel, line 229: Let's move these migration related methods to a separate struct inside the the `/patches` directory so that it clear that won't be actively in use.

Source/Utilis/Keychain/PasscodeKeychainItem.swift
```diff
@@ -0,0 +1,114 @@
...
+                     kSecValueData: value]
+        }
+
+        return query
+    }
+}
+
+// MARK: - Migration rules
+
+public extension PasscodeKeychainItem {
```
typfel, line 65: Also move this migration related code to a struct in the `/patches` directory.

typfel (changes requested):

Source/ManagedObjectContext/NSManagedObjectContext+AppLock.swift
KaterinaWire: Done

Source/Model/FeatureConfig/AppLock/AppLockController.swift
KaterinaWire: Moved

Source/Utilis/Keychain/PasscodeKeychainItem.swift
KaterinaWire: Moved

typfel (approved):

Patches/PersistedDataPatches+Directory.swift
```diff
@@ -33,7 +33,8 @@ extension PersistedDataPatch {
         PersistedDataPatch(version: "161.0.1", block: TransferStateMigration.migrateLegacyTransferState),
         PersistedDataPatch(version: "167.3.0", block: AvailabilityBehaviourChange.notifyAvailabilityBehaviourChange),
         PersistedDataPatch(version: "198.0.0", block: ZMConversation.introduceParticipantRoles),
-        PersistedDataPatch(version: "220.0.4", block: InvalidConnectionRemoval.removeInvalid)
+        PersistedDataPatch(version: "220.0.4", block: InvalidConnectionRemoval.removeInvalid),
+        PersistedDataPatch(version: "234.0.0", block: TransferApplockKeychain.migrateKeychainItems) //TODO: katerina check the version
```
billypchan, line 37: the version is checked? we can remove the comment.

KaterinaWire: Removed

billypchan (approved):

KaterinaWire:
@typfel @billypchan I moved a part related to the passcode migration to a separate PR, because this question was not yet approved by the product team.

PR #1087, opened 2020-12-09
[Feature] Expose decryption error code on system messages

typfel:
## What's new in this PR?

When we insert a decryption error system message we now store the error code on the system message so we can display it to the user.


typfel:
Failing tests is because I haven't updated DB migration tests

KaterinaWire (approved):

marcomaddy1982 (approved):

PR #1088, opened 2020-12-10
[Feature] Insert session reset system message

typfel:
## What's new in this PR?

Upon receiving a `ClientAction.RESET_ACTION` we now insert a `.sessionReset` system message.


typfel:
Failing tests is because I haven't updated DB migration tests

KaterinaWire (approved):

Source/Model/Conversation/ZMConversation+SystemMessages.swift
```diff
@@ -21,6 +21,14 @@ import Foundation
 
 extension ZMConversation {
     
+    public func appendSessionResetSystemMessage(user: ZMUser, client: UserClient, at timestamp: Date) {
```
KaterinaWire, line 24: Should it be public?

typfel: I made it public because all other append methods in this file are also public. But you're right,  it doesn't really need to be public.

billypchan (approved):

PR #1089, opened 2020-12-11
add isTrust to UserType protocol

billypchan:
## What's new in this PR?

Add `isTrust` to `Usertype` for easier mocking.

### Discussion

Should `ZMSearchUser` `isTrust` always return false?

### DIFF
--- a/Source/Model/User/ZMSearchUser.swift
+++ b/Source/Model/User/ZMSearchUser.swift
@@ -137,7 +137,7 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
     
     /// Whether all user's devices are verified by the selfUser
     public var isTrusted: Bool {
-        return false
+        return user?.isTrusted ?? false
     }
     
     public var teamCreatedBy: UUID? {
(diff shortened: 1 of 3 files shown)


typfel:
>Should ZMSearchUser isTrust always return false?

No, because search user can be backed by real `ZMUser`. We should do this:
```
var isTrusted {
     return user?.isTrusted ?? false
}
```

Source/Model/User/ZMSearchUser.swift
```diff
@@ -143,7 +148,7 @@ public class ZMSearchUser: NSObject, UserType, UserConnectionType {
 
     public var emailAddress: String? {
         get {
-            return user?.emailAddress
+            return (user as? ZMUser)?.emailAddress
```
typfel, line 151: `user` is already a `ZMUser`

typfel (changes requested):

typfel (approved):

David-Henner (approved):

billypchan:
@zenkins release minor

PR #1090, opened 2020-12-16
[Fix] Save needsToNotifyUser for feature

KaterinaWire:
## What's new in this PR?

Add `saveOrRollback()` after `needsToNotifyUser` changing.


Source/Model/FeatureConfig/Feature.swift
```diff
@@ -173,8 +173,20 @@ public class Feature: ZMManagedObject {
...
         return feature
     }
+    
+    func update(needsToNotifyUser: Bool) {
+        guard let context = managedObjectContext else {
+            return
+        }
+        self.needsToNotifyUser = needsToNotifyUser
+        context.saveOrRollback()
+    }
```
johnxnguyen, line 183: I have the feeling it's not good practice for a managed object to trigger a save. We usually see some external object modify a context (by inserting, deleting or mutating managed objects), and then triggering a save.

@typfel what are your thoughts?

typfel: Yes, if this change is coming as an action from the UI we usually require the UI to wrap the change inside a `UserSession.performChanges()` which will trigger a save.

KaterinaWire: I will add `UserSession.performChanges()` in the UI and close this PR

PR #1091, opened 2020-12-16
[Fix] app crash when `service` payload is null

David-Henner:
### Issues

App crashes on launch

### Causes

In `[ZMUser updateWithTransportData:(NSDictionary *) authoritative:(Bool)]` we access the value stored in the dictionary at the `service` key.

The value returned is an instance of `NSNull`.
We expect that value to be an instance of `NSDictionary`.
Before we try to access its values, we check that it isn't `nil`.
However, since it is an instance of `NSNull`, it is not `nil`.
Consequently, we treat it as a `NSDictionary` and the app crashes.

### Solutions

use `optionalDictionaryForKey` to get the dictionary


typfel (approved):

KaterinaWire (approved):

David-Henner:
@zenkins release patch

PR #1092, opened 2020-12-17
[Refactor] Remove useCustomCodeInsteadOfAccountPassword from the Feature.Config

KaterinaWire:
## What's new in this PR?

Remove `useCustomCodeInsteadOfAccountPassword` from the Feature.Config because we don't use it anymore in the UI. The fallback to unlock Wire is always a custom passcode.

### DIFF
--- a/Source/Model/FeatureConfig/AppLock/AppLockController.swift
+++ b/Source/Model/FeatureConfig/AppLock/AppLockController.swift
@@ -147,18 +147,15 @@ public final class AppLockController: AppLockType {
     // MARK: - Types
     
     public struct Config {
-        public let useBiometricsOrAccountPassword: Bool
-        public let useCustomCodeInsteadOfAccountPassword: Bool
+        public let useBiometricsOrCustomPasscode: Bool
         public var forceAppLock: Bool
         public var appLockTimeout: UInt
         public var isAvailable: Bool
         
-        public init(useBiometricsOrAccountPassword: Bool,
-                    useCustomCodeInsteadOfAccountPassword: Bool,
+        public init(useBiometricsOrCustomPasscode: Bool,
                     forceAppLock: Bool,
                     timeOut: UInt) {
-            self.useBiometricsOrAccountPassword = useBiometricsOrAccountPassword
-            self.useCustomCodeInsteadOfAccountPassword = useCustomCodeInsteadOfAccountPassword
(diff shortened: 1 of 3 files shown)


Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -147,18 +147,15 @@ public final class AppLockController: AppLockType {
     // MARK: - Types
     
     public struct Config {
-        public let useBiometricsOrAccountPassword: Bool
-        public let useCustomCodeInsteadOfAccountPassword: Bool
+        public let useBiometricsOrCustomPasscode: Bool
```
johnxnguyen, line 150: Considering that the default chain of authentication methods is first biometrics, then device passcode, then finally custom passcode, this property really means "don't use device passcode".

Probably it is out of the scope of this PR, but I'm curious to know what you think about the following proposal, as an effort to improve this last configurable property.

We could define an option set of `authenticationMethods`, which would contain `biometrics`, `devicePasscode` and `customPasscode`. We could then replace `useBiometricsOrCustomPasscode` with a config property called `allowedAuthenticationMethods`. We can then use this property to decide which authentication methods to try. Any existing configs with `useBiometricsOrCustomPasscode` set to true would have `allowedAuthenticationMethods` set to `biometrics | customPasscode`.

typfel: That's certainly less confusing but do we actually need any configuration at all for this now?

typfel:
Build failed

typfel (changes requested):
Build failed

Source/Model/FeatureConfig/AppLock/AppLockController.swift
johnxnguyen: Unfortunately there's still one white label version that uses this flag, so we have to continue to honor it. But yeah, I'm not sure the effort is worth it.

johnxnguyen (approved):

typfel (approved):

Source/Model/FeatureConfig/AppLock/AppLockController.swift
KaterinaWire: Let's do it in the next PR. I created a Jira ticket for this.

PR #1095, opened 2020-12-18
[Feature] Feature configurations

KaterinaWire:
## What's new in this PR?

This is the feature branch for _feature configurations_. Feature configuration is a mechanism that allows the client to receive configuration payloads from the backend used to adjust the behavior of features on the client. The first feature to use this configuration is the app lock.

In this branch we:
- introduce the `Feature` entity used to store configurations locally.
- introduce `AppLockController` (moved from UI)
- add migration code for handling the custom passcode keychain item

## To do

- [x] In PersistedDataPatches+Directory file, for `TransferApplockKeychain.migrateKeychainItems` put the new version that will be introduce with this PR.


Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -0,0 +1,247 @@
...
+        guard self.weakLAContext == nil else { return }
+        
+        let context: LAContext = LAContext()
+        var error: NSError?
+        
+        self.weakLAContext = context
+        
+        let canEvaluatePolicy = context.canEvaluatePolicy(scenario.policy, error: &error)
+        
+        if scenario.supportsUserFallback && (/*BiometricsState.biometricsChanged(in: context) || */!canEvaluatePolicy) {
```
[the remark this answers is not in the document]
KaterinaWire: This was a temporary comment because I tested it and changed the biometrics state very often.

Patches/PersistedDataPatches+Directory.swift
```diff
@@ -33,7 +33,8 @@ extension PersistedDataPatch {
         PersistedDataPatch(version: "161.0.1", block: TransferStateMigration.migrateLegacyTransferState),
         PersistedDataPatch(version: "167.3.0", block: AvailabilityBehaviourChange.notifyAvailabilityBehaviourChange),
         PersistedDataPatch(version: "198.0.0", block: ZMConversation.introduceParticipantRoles),
-        PersistedDataPatch(version: "220.0.4", block: InvalidConnectionRemoval.removeInvalid)
+        PersistedDataPatch(version: "233.3.0", block: InvalidConnectionRemoval.removeInvalid),
+        PersistedDataPatch(version: "233.3.0", block: TransferApplockKeychain.migrateKeychainItems)
```
johnxnguyen, line 37: Don't forget to update this version number.

johnxnguyen (changes requested):

Patches/TransferApplockKeychain.swift
```diff
@@ -0,0 +1,87 @@
...
+        
+        guard
+            let data = ZMKeychain.data(forAccount: "lockApp"),
+            data.count != 0
+        else {
+            selfUser.isAppLockActive = false
+            return
+        }
+        
+        selfUser.isAppLockActive = String(data: data, encoding: .utf8) == "YES"
```
marcomaddy1982, line 46: First of all, sorry if I missed something from the previous PRs: but why we don't use `true`/`false` instead of `YES` / `NO`?

Using `true`/`false` I guess this method will become something just

```swift
static func migrateIsAppLockActiveState(in moc: NSManagedObjectContext) {
         let selfUser = ZMUser.selfUser(in: moc)
         let data = ZMKeychain.data(forAccount: "lockApp") ?? Data()
         return String(data: data, encoding: .utf8).flatMap(Bool.init) ?? false
}
```

Patches/TransferApplockKeychain.swift
```diff
@@ -0,0 +1,87 @@
...
+
+        guard
+            let passcode = try? Keychain.fetchItem(legacyKeychainItem),
+            !passcode.isEmpty
+        else {
+            return
+        }
+
+        do {
+            for id in ids {
```
marcomaddy1982, line 63: It doesn't make much difference but to be more "swiftly"  here you could use `ids.forEach {  }`. Anyway feel free to use the approach you prefer.

Patches/PersistedDataPatches+Directory.swift
KaterinaWire: The current version is `233.2.0` and it will be a major release. So, I should change to `234.0.0`, right?

Source/Model/FeatureConfig/AppLock/AppLockController+Passcode.swift
```diff
@@ -0,0 +1,42 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+
+public extension AppLockController {
```
marcomaddy1982, line 22: I know that the `AppLockController` is a huge class but honestly I'm not a big fun of having so small category. I would just move this  `extension` in the  `AppLockController` and then use a PragmaMark for it. Anyway, this is just my opinion. Feel free to use the approach you prefer.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -0,0 +1,247 @@
...
+            case .databaseLock:
+                return false
+            }
+        }
+    }
+}
+
+// MARK: - BiometricsState
+
+public class BiometricsState {
```
marcomaddy1982, line 217: Here I would move it out in a separate file.

Source/Model/FeatureConfig/AppLock/AppLockController.PasscodeKeychainItem.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension AppLockController {
+
+    struct PasscodeKeychainItem: KeychainItem {
```
marcomaddy1982, line 23: Sorry if I missed some here too for the previous PRs: but why declaring this inside the `AppLockController`?

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -0,0 +1,247 @@
...
+            zmLog.error("Local authentication error: \(String(describing: error?.localizedDescription))")
+        }
+    }
+    
+    public func persistBiometrics() {
+        BiometricsState.persist()
+    }
+    
+    
+    // MARK: - Types
```
marcomaddy1982, line 162: Here I would separate this types and declare them outside from the `AppLockController` and remake them. As mentioned before this is just my opinion.

marcomaddy1982 (approved):

Patches/PersistedDataPatches+Directory.swift
johnxnguyen: Yes

KaterinaWire: Done

Patches/PersistedDataPatches+Directory.swift
```diff
@@ -33,7 +33,8 @@ extension PersistedDataPatch {
         PersistedDataPatch(version: "161.0.1", block: TransferStateMigration.migrateLegacyTransferState),
         PersistedDataPatch(version: "167.3.0", block: AvailabilityBehaviourChange.notifyAvailabilityBehaviourChange),
         PersistedDataPatch(version: "198.0.0", block: ZMConversation.introduceParticipantRoles),
-        PersistedDataPatch(version: "220.0.4", block: InvalidConnectionRemoval.removeInvalid)
+        PersistedDataPatch(version: "234.0.0", block: InvalidConnectionRemoval.removeInvalid),
```
johnxnguyen, line 36: Did you mean to change this version too?

KaterinaWire: Oh, no!
It was by accident.

johnxnguyen (approved):

KaterinaWire:
@zenkins release major

PR #1097, opened 2020-12-23
[Fix] Send confirmation receipts to composite messages

johnxnguyen:
### Issues

E-alarming messages (send by the e-alarming bot) are not receiving delivery and read receipts when in conversation with just the self user. This is problematic because the bot uses these confirmations to gather metrics for the e-alarming console.

### Causes

1. We always consider a conversation with a bot a group conversation, and delivery receipts are not sent in group conversations.
2. We never send read receipts for the `composite` message type (the type of message used in e-alarming)

### Solutions

1. Remove this restriction, allowing 1:1 conversations with bots.
2. Allow sending read receipts for the `composite` message type.

## Notes

This behavior aligns the iOS app with the behavior on the web app.


billypchan (approved):
We should add a test to confirm this requirement for the future

johnxnguyen:
@billypchan this will be released as a hotfix for the last public release so we shouldn't include `develop`.

Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -203,11 +203,6 @@ extension ZMConversation {
             updateLastModified(timestamp)
         }
         
-        if let sender = message.sender, sender.isSelfUser {
-            // if the message was sent by the self user we don't want to send a lastRead event, since we consider this message to be already read
-            updateLastRead(timestamp, synchronize: false)
-        }
-
```
johnxnguyen, line 210: Deleting this is a fix for an additional issue where a read receipt for a e-alarming message would not be sent if a message button was tapped within 1 or 2 seconds after reading that message. 

The reason is because process of sending delivery receipts happens after an artificial delay from the time that we mark messages as read. After this delay, we eventually try to fetch all unread messages using a predicate, where one of the conditions it that a message is considered unread if its timestamp is after the last read server timestamp of the conversation.

However, if the last read server timestamp is updated within this artificial delay, it can affect which messages are considered unread.

For example, consider the happy path:

1. conversation last read server timestamp is 0.
2. receive message A with timestamp 1
3. open conversation and mark message as read.
4. wait 2 seconds.
5. fetch unread messages, we see A is returned because the timestamp is newer than the last read server timestamp.
6. send read receipt.

Compare this to updating the last read server timestamp between steps 4 and 5:

1. conversation last read server timestamp is 0.
2. receive message A with timestamp 1
3. open conversation and mark message as read.
4. wait 2 seconds, meanwhile set the conversation last read server timestamp to 2.
5. fetch unread messages, we see A is **not** returned because the timestamp 1 is **not** newer than the last read server timestamp 2.
6. don't send read receipt for message A.

The reason why this was happening when we tap the alarming button action as quickly as possible after seeing the alert message for the first time is that doing so would cause the message button to update, which would result in the invocation of this method `updateTimestampsAfterUpdatingMessage(_ message: ZMMessage)`. Since the message button is sent from the self user, we were updating the last read server timestamp with the timestamp of the message button, which came after the original e-alarming message. 

I believe removing this code to update the last read timestamp should be innocuous. @typfel any thoughts?

billypchan: I would suggest adding a test to simulate the case 1 and case 2

billypchan (approved):

billypchan (commented):
@johnxnguyen should we only include the code change for the last hotfix and leave the later changes for the next release?

Source/Model/Conversation/ZMConversation+Timestamps.swift
johnxnguyen: Yes, I'll try add tests now.

johnxnguyen:
> should we only include the code change for the last hotfix and leave the later changes for the next release?

@billypchan I would say the whole PR should go into the next release.

Source/Model/Conversation/ZMConversation+Timestamps.swift
typfel: Yes, this was an optimisation to avoid sending last read updates.

typfel (approved):

johnxnguyen:
@billypchan tests added.

billypchan (approved):

johnxnguyen:
@zenkins release patch

PR #1099, opened 2021-01-04
add senderUser

billypchan:
## What's new in this PR?

rework of https://github.com/wireapp/wire-ios-data-model/pull/1094

### Issues

In UI project, it is not easy to mock a message object without creating a `ZMUser` object.

### Causes

`ZMConversationMessage` has a `sender` property which is `ZMUser` and UI project must create a `ZMUser` to Mock a message.

### Solutions

Add a `senderUser` member with `UserType?` type. 

In https://github.com/wireapp/wire-ios/pull/4798, we can create a mock message with a `senderUser` with `MockUserType`

Adding new property would not affecting data model and downstream components since they do not access it.

When not running the tests,`ZMMessage.senderUser` returns `ZMMessage.sender` in UI Project which would not affect the existing code.

## Discussion

- [ ] a better name of this property?

### DIFF
--- a/Source/Model/Message/ConversationMessage.swift
+++ b/Source/Model/Message/ConversationMessage.swift
@@ -48,7 +48,10 @@ public protocol ZMConversationMessage : NSObjectProtocol {
         
     /// The user who sent the message
     var sender: ZMUser? { get }
-    
+
+    /// Wrapper for UI mocking ZMConversationMessage without creating ZMUser
+    var senderUser: UserType? { get }
+
     /// The timestamp as received by the server
     var serverTimestamp: Date? { get }
     
@@ -195,6 +198,10 @@ extension ZMMessage {
 // MARK:- Conversation Message protocol implementation
 
 extension ZMMessage : ZMConversationMessage {
+    public var senderUser: UserType? {
+        return sender
+    }
+    
     @NSManaged public var linkAttachments: [LinkAttachment]?
     @NSManaged public var needsLinkAttachmentsUpdate: Bool
     @NSManaged public var replies: Set<ZMMessage>


typfel:
Regarding the naming. In the future we would not want to expose the `sender: ZMUser` at all, this would require a lot of changes though. I think `senderUser` is acceptable name during this transition period.

Source/Model/Message/ConversationMessage.swift
```diff
@@ -48,7 +48,10 @@ public protocol ZMConversationMessage : NSObjectProtocol {
         
     /// The user who sent the message
     var sender: ZMUser? { get }
-    
+
+    /// Wrapper for UI mocking ZMConversationMessage without creating ZMUser
```
typfel, line 52: This API should be used everywhere in the UI so I wouldn't document it as a property only used for mocking tests.

I would use  "The user who sent the message" and change the documentation of the existing `sender` property and maybe also mark it as deprecated.
```
     /// The user who sent the message (internal)
     @available(iOS, unavailable, message: "Use `senderUser` instead")
     var sender: ZMUser? { get }
```

typfel (changes requested):

Source/Model/Message/ConversationMessage.swift
billypchan: thanks, updated for your suggestion.

Tests/Source/Model/Conversation/ZMConversationTests+Knock.swift
```diff
@@ -0,0 +1,51 @@
...
+            let selfUser = ZMUser.selfUser(in: self.syncMOC)
+    
+            // when
+            let knock = try? conversation?.appendKnock()
+            let msg = conversation?.lastMessage
+    
+            // then
+            XCTAssertEqual(knock as? ZMMessage, msg)
+            XCTAssertNotNil(knock?.knockMessageData)
+            XCTAssertEqual(knock?.senderUser as? ZMUser, selfUser)
```
johnxnguyen, line 37: You could use `knock?.isUserSend(selfUser)` to avoid the casting here.

Source/Model/Message/ConversationMessage.swift
```diff
@@ -156,8 +160,7 @@ public extension ZMConversationMessage {
     /// Whether the given user is the sender of the message.
 
     func isUserSender(_ user: UserType) -> Bool {
-        guard let zmUser = user as? ZMUser else { return false }
-        return zmUser == sender
+        return user as? ZMUser == senderUser as? ZMUser
```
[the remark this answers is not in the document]
billypchan: restored

Tests/Source/Model/Conversation/ZMConversationTests+Knock.swift
billypchan: fixed

johnxnguyen (approved):

billypchan:
@zenkins release minor

PR #1102, opened 2021-01-06
[Feature] Add method for resolving past decryption failed system messages

typfel:
## What's new in this PR?

Add a new system message type `.decryptionFailedResolved`. When a session is reset we will update all existing `.decryptionFailed` system messages to `.decryptionFailedResolved` to signal to the user that the decryption error has been fixed and there's no need to reset the session anymore.

## NOTE

The failing tests are due to not updated database migration tests and can be disregarded.

### DIFF
--- a/Source/ManagedObjectContext/NSManagedObjectContext+UpdateRequest.swift
+++ b/Source/ManagedObjectContext/NSManagedObjectContext+UpdateRequest.swift
@@ -20,6 +20,12 @@ import Foundation
 
 public extension NSManagedObjectContext {
     
+    /// Execute an NSBatchUpdateRequest and merge the resulting changes into all contexts.
+    ///
+    /// - parameters:
+    ///   - request: NSBatchUpdateRequest to exectute
+    ///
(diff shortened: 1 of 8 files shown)


Source/ManagedObjectContext/NSManagedObjectContext+UpdateRequest.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public extension NSManagedObjectContext {
+    
+    func executeBatchUpdateRequestOrAssert(_ request: NSBatchUpdateRequest) {
```
johnxnguyen, line 23: It would be nice to document this public API.

johnxnguyen (approved):

marcomaddy1982 (approved):

johnxnguyen (approved):

PR #1105, opened 2021-01-07
[Fix] Changed conditions for displaying  custom passcode screen

KaterinaWire:
### Issues
The  custom passcode screen displays for users who should use a device passcode as a fallback.

### Solutions
Add an additional condition.
Only users who has `config.useBiometricsOrCustomPasscode == true` should use a custom passcode as a fallback.


johnxnguyen (commented):
I think you forgot to replace the existing condition (line 143 onwards)

KaterinaWire:
> I think you forgot to replace the existing condition (line 143 onwards)

Fixed

johnxnguyen (changes requested):
Please add a test asserting that:

1. custom passcode is requested if the policy can't be evaluated
2. custom passcode is requested if policy can be evaluated, but biometrics changed

KaterinaWire:
> Please add a test asserting that:
> 
> 1. custom passcode is requested if the policy can't be evaluated
> 2. custom passcode is requested if policy can be evaluated, but biometrics changed

@johnxnguyen Added tests

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
```diff
@@ -192,3 +185,79 @@ final class AppLockControllerTest: ZMBaseManagedObjectTest {
...
+        
+        //when
+        sut.evaluateAuthentication(scenario: scenario, description: "evaluate authentication") { (result, context) in
+            var error: NSError?
+            XCTAssertFalse(context.canEvaluatePolicy(scenario.policy, error: &error))
+            
+            //then
+            XCTAssertEqual(result, .needCustomPasscode)
+        }
+    }
```
johnxnguyen, line 206: I suppose within this method we're using a real `LAContext` to evaluate the policy. I think it would be better to mock this as to make the test more controlled and robust. You can see in https://github.com/wireapp/wire-ios/pull/4732/ that I made a `MockLAContext`, perhaps we can introduce that here.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
```diff
@@ -192,3 +185,79 @@ final class AppLockControllerTest: ZMBaseManagedObjectTest {
     }
 }
 
+// MARK : Evaluate Authentication
+
+extension AppLockControllerTest {
+    
+    func testThatCustomPasscodeIsRequested_IfThePolicyCanNotBeEvaluated() {
+        //given
+        let scenario = AppLockController.AuthenticationScenario.screenLock(requireBiometrics: true)
```
johnxnguyen, line 194: I think we should have a copy of this test when biometrics aren't required.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
```diff
@@ -192,3 +185,79 @@ final class AppLockControllerTest: ZMBaseManagedObjectTest {
...
+            //then
+            XCTAssertEqual(result, .needCustomPasscode)
+        }
+    }
+    
+    func testThatCustomPasscodeIsRequested_IfThePolicyCanBeEvaluated_ButBiometricsHaveChanged() {
+        //given
+        sut = createAppLockController(useBiometricsOrCustomPasscode: true)
+        
+        let scenario = AppLockController.AuthenticationScenario.screenLock(requireBiometrics: false)
```
johnxnguyen, line 212: Does setting `useBiometricsOrCustomPasscode` have any affect on the test? If I remember correctly, the evaluate method only checks the `requireBiometrics` value of the `scenario`.

Also I think you want to have `requireBiometrics` equal to `true`, since the biometric change state is related to this flag.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
```diff
@@ -192,3 +185,79 @@ final class AppLockControllerTest: ZMBaseManagedObjectTest {
...
+        XCTAssertEqual(scenario.policy, .deviceOwnerAuthentication)
+        XCTAssertTrue(scenario.supportsUserFallback)
+        
+        UserDefaults.standard.set(Data(), forKey: "DomainStateKey")
```
johnxnguyen, line 216: I suggest to mock the `biometricsChanged()` method.

johnxnguyen (changes requested):

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: Yes, it would be nice, but a real `LAContext` is created inside the `evaluateAuthentication()` method.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: This method mocks our bundle config. So yes, it has an affect. If `useBiometricsOrCustomPasscode` is false we use a device passcode as a fallback.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: This setting affects only the `scenario.policy`:
```
var policy: LAPolicy {
            switch self {
            case .screenLock(requireBiometrics: let requireBiometrics):
                return requireBiometrics ? .deviceOwnerAuthenticationWithBiometrics : .deviceOwnerAuthentication
            case .databaseLock:
                return .deviceOwnerAuthentication
            }
        }
```

I just used `requireBiometrics = true` because `deviceOwnerAuthenticationWithBiometrics` can't be evaluated by simulator and I could mock the expected behaviour.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -133,7 +133,10 @@ public final class AppLockController: AppLockType {
         
         let canEvaluatePolicy = context.canEvaluatePolicy(scenario.policy, error: &error)
         
-        if scenario.supportsUserFallback && (BiometricsState.biometricsChanged(in: context) || !canEvaluatePolicy) {
+        let biometricsChanged = BiometricsState.biometricsChanged(in: context) && config.useBiometricsOrCustomPasscode
```
typfel, line 136: Isn't `useBiometricsOrCustomPasscode` already part of the `AuthenticationScenario`? Seems like `AuthenticationScenario.supportsUserFallback` could take this into account. 

Then we could simplify this to:

```suggestion
        let biometricsChanged = BiometricsState.biometricsChanged(in: context)
```

typfel (changes requested):

johnxnguyen (commented):
An idea for improving readability of tests: you could extract a lot of boilerplate code into a dedicated `assert` method which takes an `input` and an expected `output` argument. This makes it much easier to understand what is being tested and also makes it easier to set multiple tests (even all input combinations if you want). Something like:

```swift
func testEvaluateScenario() {
  assert(
    input: (scenario: .screenLock(requiresBiometrics: true), canEvaluate: true,  biometricsChanged: true),
    output: .needCustomPasscode
  )

  assert(
    input: (scenario: .screenLock(requiresBiometrics: true), canEvaluate: false,  biometricsChanged: true),
    output: .needCustomPasscode
  )
}

func assert(input: Input, output: AutnethicationResult, file: StaticString = #file, line: Uint = #line) {
 ...
}
```

Source/Model/FeatureConfig/AppLock/AppLockController.swift
johnxnguyen: I think there may be some confusion (on my part) about what exactly `supportsUserFalback` is supposed to mean. I took it to understand that there is a secondary authentication method available. But is it suppose to mean that the fallback must be the custom passcode?

It's confusing because if you have biometrics enabled, you must also have a device pin set (which is also used as a fallback).

I'm thinking now that Jacob's suggest would work with:

```swift
var supportsUserFallback: Bool {
  switch self {
    case let .screenLock(requireBiometrics):
      return requireBiometrics
    case .databaseLock:
      return false
  }
}
```

johnxnguyen: Probably also we should rename `supportsUserFallback` to something like `shouldFallbackToCustomPasscode`.

typfel: >It's confusing because if you have biometrics enabled, you must also have a device pin set (which is also used as a fallback).

But we don't explicitly request the PIN code as fallback do we? It happen automatically by the OS. When EAR is enabled `supportsUserFallback` is always false but you can't still unlock your phone with the PIN if the biometric fails.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
johnxnguyen: It's a bit confusing though, no one would immediately understand why that is the reason `requireBiometrics` is true. I suggest to mock the `LAContext` to avoid such issues.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
johnxnguyen: Yes it's handled by the OS, but the naming `supportsUserFallback` doesn't make that so clear (leading to my confusion).

johnxnguyen: > When EAR is enabled `supportsUserFallback` is always false but you can't still unlock your phone with the PIN if the biometric fails.

Do you mean "... but you **can** still unlock your phone with the PIN""?

If `supportsUserFallback` is false, then it means that we shouldn't allow the user to enter via a custom passcode? I'm wondering what it means if EAR is enabled and `useBiometricsOrCustomPasscode` is `true`. I guess it means that you must use biometrics only and nothing else.

KaterinaWire: `useBiometricsOrCustomPasscode` doesn't affect EAR. 
```
var policy: LAPolicy {
            switch self {
            case .screenLock(requireBiometrics: let requireBiometrics):
                return requireBiometrics ? .deviceOwnerAuthenticationWithBiometrics : .deviceOwnerAuthentication
            case .databaseLock:
                return .deviceOwnerAuthentication
            }
        }
```

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -103,8 +114,10 @@ public final class AppLockController: AppLockType {
         }
     }
     
-    /// a weak reference to LAContext, it should be nil when evaluatePolicy is done.
-    private weak var weakLAContext: LAContext? = nil 
+//    /// a weak reference to LAContext, it should be nil when evaluatePolicy is done.
+//    private weak var weakLAContext: LAContextProtocol? = nil
```
KaterinaWire, line 118: @typfel I'm not sure if we still need to persist `weakLAContext`.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: I added `MockLAContext` and replaced this test

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: Test was replaced

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: Test was replaced

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: `biometricsChanged()` is mocked and test was replaced

KaterinaWire:
> An idea for improving readability of tests: you could extract a lot of boilerplate code into a dedicated `assert` method which takes an `input` and an expected `output` argument. This makes it much easier to understand what is being tested and also makes it easier to set multiple tests (even all input combinations if you want). Something like:
> 
> ```swift
> func testEvaluateScenario() {
>   assert(
>     input: (scenario: .screenLock(requiresBiometrics: true), canEvaluate: true,  biometricsChanged: true),
>     output: .needCustomPasscode
>   )
> 
>   assert(
>     input: (scenario: .screenLock(requiresBiometrics: true), canEvaluate: false,  biometricsChanged: true),
>     output: .needCustomPasscode
>   )
> }
> 
> func assert(input: Input, output: AutnethicationResult, file: StaticString = #file, line: Uint = #line) {
>  ...
> }
> ```

@johnxnguyen Implemented

Source/Model/FeatureConfig/AppLock/AppLockController.swift
KaterinaWire: @typfel @johnxnguyen I replaced `var supportsUserFallback: Bool ` with `var supportedUserFallback: UserFallback`
And 
```
enum UserFallback {
        /// User is allowed to use only a custom passcode as a fallback
        case customPasscode
        /// User is allowed to use only a device passcode as a fallback
        case devicePasscode
        /// User is allowed to use any fallback
        case all
    }
```
Because we have different use cases for `UserFallback`.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
typfel: If nothing references it we don't need it and since it was private this is a safe assumption.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -122,18 +135,20 @@ public final class AppLockController: AppLockType {
     // Creates a new LAContext and evaluates the authentication settings of the user.
     public func evaluateAuthentication(scenario: AuthenticationScenario,
                                        description: String,
-                                       with callback: @escaping (AuthenticationResult, LAContext) -> Void) {
+                                       context: LAContextProtocol = LAContext(),
+                                       with callback: @escaping (AuthenticationResult, LAContextProtocol) -> Void) {
```
typfel, line 139: Just be aware that this is now a breaking change. The context in the closure is calling `unlockDatabase(with context: LAContext)` via the UI project. That path also needs to be updated to use `LAContextProtocol`

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -122,18 +135,20 @@ public final class AppLockController: AppLockType {
...
+//        guard self.weakLAContext == nil else { return }
         
-        let context: LAContext = LAContext()
         var error: NSError?
         
-        self.weakLAContext = context
+//        self.weakLAContext = context
         
         let canEvaluatePolicy = context.canEvaluatePolicy(scenario.policy, error: &error)
+        let shouldUseCustomPasscodeAsFallback = biometricsState.biometricsChanged(in: context) && (scenario.supportedUserFallback == .customPasscode)
```
typfel, line 148: IMO this is easier to read if we rename `shouldUseCustomPasscodeAsFallback` to `biometricsChanged `.

```suggestion
        let biometricsChanged = biometricsState.biometricsChanged(in: context) && (scenario.supportedUserFallback == .customPasscode)
```

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -188,6 +203,15 @@ public final class AppLockController: AppLockType {
...
     }
     
+    enum UserFallback {
+        /// User is allowed to use only a custom passcode as a fallback
+        case customPasscode
+        /// User is allowed to use only a device passcode as a fallback
+        case devicePasscode
+        /// User is allowed to use any fallback
+        case all
+    }
```
johnxnguyen, line 213: Why do we need this? The ui for presenting the entering the device passcode isn't something managed by us. Indeed I noticed that we never check in the evaluate method if `supportedUserFallback == .devicePasscode`.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -201,47 +225,13 @@ public final class AppLockController: AppLockType {
...
-        var supportsUserFallback: Bool {
+        var supportedUserFallback: UserFallback {
             switch self {
-            case .screenLock:
-                return true
+            case .screenLock(requireBiometrics: let requireBiometrics):
+                return requireBiometrics ? .customPasscode : .all
             case .databaseLock:
-                return false
+                return .devicePasscode
```
johnxnguyen, line 233: As far as I understand, the user fallback is only referring to a custom passcode as a secondary way to unlock the app. I believe the only scenario in which we can have such a fallback is the screen lock with biometrics required. We don't support using a device pin and a custom passcode as a fallback. And if we don't have even a device pin, we wouldn't ever use the `evaluate` method because custom passcode verification happens outside of the `AppLockController`.

Source/Model/FeatureConfig/AppLock/BiometricsState.swift
```diff
@@ -0,0 +1,56 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+protocol BiometricsStateProtocol {
+   func biometricsChanged(in context: LAContextProtocol) -> Bool
```
johnxnguyen, line 22: Is there a need to pass in `LAContextProtocol` or can we simply pass `LAContext`? If we're mocking this method, we don't need to use the argument at all and therefore we wouldn't need to rely on the `evaluatedPolicyDomainState`. This means we could remove that requirement from `LAContextProtocol` to keep it simpler.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
```diff
@@ -82,18 +89,18 @@ final class AppLockControllerTest: ZMBaseManagedObjectTest {
...
     func testThatBiometricsStatePersistsState() {
         //given
-        let context = LAContext()
-        var error: NSError?
-        context.canEvaluatePolicy(LAPolicy.deviceOwnerAuthentication, error: &error)
-        _ = BiometricsState.biometricsChanged(in: context)
+        UserDefaults.standard.set(Data(), forKey: "DomainStateKey")
+        var context = MockLAContext(canEvaluate: true)
+        context.evaluatedPolicyDomainState = Data()
+        _ = sut.biometricsState.biometricsChanged(in: context)
```
johnxnguyen, line 100: By setting `Data()` as both the value in user defaults and `context.evaluatedPolicyDomainState`, you won't be able to assert that the persist method does anything; it could not work and the assertion that below would still be true. Better to store some other data in the context to see if that is saved in the user defaults.

Also, it would be good to clean up the value in user defaults at the end of the test.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
```diff
@@ -209,11 +211,6 @@ extension AppLockControllerTest {
             input: (scenario: .screenLock(requireBiometrics: false), canEvaluate: true,  biometricsChanged: false),
             output: .granted
         )
-        
-        assert(
-            input: (scenario: .screenLock(requireBiometrics: false), canEvaluate: false,  biometricsChanged: false),
-            output: .needCustomPasscode
-        )
```
johnxnguyen: We should probably assert that the output is something like `.denied`.

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
```diff
@@ -90,14 +90,16 @@ final class AppLockControllerTest: ZMBaseManagedObjectTest {
...
         //given
-        UserDefaults.standard.set(Data(), forKey: "DomainStateKey")
-        var context = MockLAContext(canEvaluate: true)
-        context.evaluatedPolicyDomainState = Data()
-        _ = sut.biometricsState.biometricsChanged(in: context)
+        let evaluatedPolicyDomainStateData = "test".data(using: .utf8)
+        UserDefaults.standard.set(evaluatedPolicyDomainStateData, forKey: "DomainStateKey")
+        
+        let context = LAContext()
+        XCTAssertTrue(sut.biometricsState.biometricsChanged(in: context))
```
johnxnguyen, line 102: You should also clean up the user defaults value for this test too.

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -225,12 +209,12 @@ public final class AppLockController: AppLockType {
             }
         }
         
-        var supportedUserFallback: UserFallback {
+        var usesCustomPasscodeAsUserFallback: Bool {
```
johnxnguyen, line 212: Suggestion: `usesCustomPasscodeAsFallback`

KaterinaWire: Done

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: Done

typfel (approved):

johnxnguyen (approved):

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTest.swift
KaterinaWire: The output should be `.unavailable`, but the test fails. 
```
if canEvaluatePolicy {
...
} else {
            callback(.unavailable, context)
            zmLog.error("Local authentication error: \(String(describing: error?.localizedDescription))")
}
```
It fails because of `zmLog.error(...)` . Do you know why it happened?
Logs says:
`-[WireDataModelTests.AppLockControllerTest testEvaluateScenario] : failed - Unexpected log error:`

KaterinaWire: As I understood, `evaluateAuthentication(...)` method throws the error (`zmLog.error("Local authentication error: \(String(describing: error?.localizedDescription))")`), which I can't catch from the test.

typfel: Tests will fail if errors are logged while they execute. If the errors are expected the error producing methods need to be wrapped inside a `performIgnoringZMLogError {}` block.

KaterinaWire: Thanks!

KaterinaWire:
@zenkins release major

PR #1107, opened 2021-01-11
[Fix] Don't create connection with self user

typfel:
### Issues

We were getting reports that user's would end up with 1:1 conversations with the self user. 

### Causes

This is likely a side effecting of the long existing issue of `localParticipants` relationship sometimes being reported as empty. If this would happen when receiving a message from the self user in a 1:1 conversation we would try re-establish the connection. This never makes for the self user though since he/she should never have a connection.

This bug was probably introduced with the addition of participant roles because then we started to include the self user in the `localParticipants` relationship  (previously named `activeParticipants`).

### Solutions

- Return early in `addParticipantAndSystemMessageIfMissing()` if it's the self user who's missing.
- Re-apply the patch which cleans up invalid connections with the self user

### DIFF
--- a/Patches/PersistedDataPatches+Directory.swift
+++ b/Patches/PersistedDataPatches+Directory.swift
@@ -34,7 +34,8 @@ extension PersistedDataPatch {
         PersistedDataPatch(version: "167.3.0", block: AvailabilityBehaviourChange.notifyAvailabilityBehaviourChange),
         PersistedDataPatch(version: "198.0.0", block: ZMConversation.introduceParticipantRoles),
         PersistedDataPatch(version: "220.0.4", block: InvalidConnectionRemoval.removeInvalid),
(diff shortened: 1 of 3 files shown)


johnxnguyen:
> ...user's would end up with 1:1 conversations with the self user.

@typfel what does this mean? A user would have a 1:1 conversation with themself?

johnxnguyen (approved):

typfel:
> what does this mean? A user would have a 1:1 conversation with themself?

Exactly, an existing 1:1 conversation would appear to be yourself since `conversation.connection.to` would point to the self user.

typfel:
I think I will also need trigger a slow sync in SE hot fix patch in order to restore any deleted connections.

billypchan (approved):

typfel:
@zenkins release patch

typfel:
@zenkins release patch

PR #1108, opened 2021-01-11
[Feature] Add migration of senderClientID for decryption error system messages

typfel:
## What's new in this PR?

We now set the `senderClientID` when inserting `.decryptionFailed` system messages since we need it to distinguish which client caused the decryption error. We also need to populate this field on any existing `.decryptionFailed` system messages otherwise they wouldn't get updated when the user resolves the decryption error (See https://github.com/wireapp/wire-ios-data-model/pull/1102)

### DIFF
--- a/Patches/MigrateSenderClient.swift
+++ b/Patches/MigrateSenderClient.swift
@@ -1,0 +1,40 @@
+//
+// Wire
+// Copyright (C) 2021 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


marcomaddy1982:
@typfel are the test failing always cause [1102](https://github.com/wireapp/wire-ios-data-model/pull/1102) ?

typfel:
> @typfel are the test failing always cause [1102](https://github.com/wireapp/wire-ios-data-model/pull/1102) ?

Tests are failing because of missing database migration tests

marcomaddy1982 (approved):

johnxnguyen (approved):

PR #1110, opened 2021-01-12
[Revert] Preventing last read update for updated messages

johnxnguyen:
## What's new in this PR?

There are some sync engine test failing due to the change in https://github.com/wireapp/wire-ios-data-model/pull/1097 that removed the optimization to update the last read server timestamp when a message was updated.

I'm reverting these changes now to unblock a release train and will address the fix in a separate PR.

PR #1111, opened 2021-01-12
[Fix] Read receipts not sent for some messages

johnxnguyen:
## What's new in this PR?

An alternate solution for the issue first described in https://github.com/wireapp/wire-ios-data-model/pull/1097 and initially solved in https://github.com/wireapp/wire-ios-data-model/pull/1110 which was later reverted.

### Issues

It can happen that read receipts are not being sent for some messages that expect them. For example, receive a message, but immediately after reading that message, send a new message. If it is sent within 3 seconds, the original message won't receive a read receipt.

### Causes

The method which triggers the generation of read receipts (`enqueueMarkAsReadUpdate()`) uses an artificial delay of 3 seconds to throttle invocations to that method. If the method is called with a timestamp A, and meanwhile that conversation `lastReadServerTimeStamp` is advanced to a value greater than A, then the delayed call to `enqueueMarkAsReadUpdate` would have no effect because it would not find any unread messages within the range `A...lastReadServerTimeStamp` because A is less than `lastReadServerTimeStamp`. This can happen if the self user appends a message in the conversation during this 3 second delay.

The core of the issue here is that last read server timestamp isn't being preserved throughout the 3 second delay because the method used to fetch unread messages (` unreadMessagesIncludingInvisible(until:)`) will always use current value of the last read server timestamp.

### Solutions

Refactor the chain of methods to accept a date range in which we are searching for unread messages. This allows us to "freeze" the conversation's last read server timestamp just before the 3 second delay begins, passing it through the chain of methods until it is finally used in a predicate to fetch the unread messages.

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Timestamps.swift
+++ b/Source/Model/Conversation/ZMConversation+Timestamps.swift
@@ -51,6 +51,10 @@ extension ZMConversation {
         if timestamp > pendingLastReadServerTimestamp {
             pendingLastReadServerTimestamp = timestamp
         }
+
+        if previousLastReadServerTimestamp == nil {
+            previousLastReadServerTimestamp = lastReadServerTimeStamp
+        }
     }
     
     @objc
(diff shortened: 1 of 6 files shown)


Source/Model/Conversation/ZMConversation+Timestamps.swift
```diff
@@ -261,36 +261,49 @@ extension ZMConversation {
...
         
         DispatchQueue.main.asyncAfter(deadline: .now() + lastReadTimestampSaveDelay) { [weak self] in
-            guard currentCount == self?.lastReadTimestampUpdateCounter else { return managedObjectContext.leaveAllGroups(groups) }
-            
-            self?.savePendingLastRead()
+            // Only allow the first dispatched block to execute because it has the oldest value
+            // for `currentLastReadTimestamp`. If we use a later block where the timestamp is newer
+            // (e.g a new message is appended by the self user which update the timestamp), then we would
+            // likely not send read receipts for any messages before that updated timestamp.
+            guard currentCount == 1 else { return managedObjectContext.leaveAllGroups(groups) }
```
typfel, line 291: Looking at this again I now realise that this changes the behaviour quite a bit. Previously this throttling would extend the delay every time `enqueueMarkAsReadUpdate()` is invoked.

```
T0 enqueue #1
T2 enqueue #2
T3 #1 finishes but doesn't execute since #2 was enqueued afterwards
T4 #2 finishes and executes nothing has been enqueued afterwards
```

now
```
T0 enqueue #1
T2 enqueue #2
T3 #1 finishes and executes
T4 #2 finishes but doesn't execute since #1 was enqueued before
```

So new behaviour is that we collect last read updates during a `lastReadTimestampSaveDelay`.

-----

Now this new behaviour isn't broken or anything it just different and we should be aware of it. If we wanted to retain the old behaviour we could store a pending `lastReadServerTimeStamp` similar to the `pendingLastReadServerTimestamp`.

johnxnguyen: Good point. Yes I was thinking this too but found the naming awkward ahaha. I'll go with this approach.

johnxnguyen: Done

typfel (approved):

KaterinaWire (approved):

johnxnguyen:
@zenkins release minor

PR #1113, opened 2021-01-13
[Fix] Display custom passcode screen when no biometrics

KaterinaWire:
### Issues

When user enables applock on a device without biometrics/passcode enabled and navigates out of Wire, user can't unlock the app.

### Solutions

Added an additional condition that allows the user to navigate to a custom passcode screen.


johnxnguyen (approved):

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -154,6 +154,11 @@ public final class AppLockController: AppLockType {
             return
         }
         
+        if case .screenLock = scenario, !canEvaluatePolicy {
```
typfel, line 157: Couldn't the same thing be achieved by making `usesCustomPasscodeAsFallback` always true for the `.screenLock` scenario?

typfel: Do we require users on on the public to assign a custom passcode now? I assume so because otherwise we couldn't require them to input it when they don't have pin code ...

johnxnguyen: > Couldn't the same thing be achieved by making usesCustomPasscodeAsFallback always true for the .screenLock scenario?

This was the behavior before a recent fix, but we need to separate out the case where biometrics is required and biometrics changed.

johnxnguyen: > Do we require users on on the public to assign a custom passcode now? I assume so because otherwise we couldn't require them to input it when they don't have pin code ...

The flow goes like this: 

Do they have biometrics enrolled? Yes -> use biometrics.
Do they have a device passcode set? Yes -> use device passcode.
Otherwise -> set a custom passcode.

It's a little more complicated because if you have biometrics enrolled, then you also have a device passcode set. And then there's the extra requirement where some customers forbid the device passcode as a fallback option, and instead they must enter the custom passcode (as a fallback because they can't use biometrics at that time, or because the biometric state changed).

typfel (approved):

Source/Model/FeatureConfig/AppLock/AppLockController.swift
typfel: Seems complicated, I don't quite follow it but I guess it needs stay like this for now then. Maybe we discuss how to simplify the app lock rules in the future?

johnxnguyen: I agree that we can restructure some code to make it more explicit how this all behaves. Just as some ideas for future reference, I would suggest to extract all code paths relating only to the case where the user only has a custom passcode. In such cases, we have no need of the LocalAuthentication framework and directly navigate to Wire's unlock screen.

johnxnguyen: > Seems complicated, I don't quite follow it but I guess it needs stay like this for now then. Maybe we discuss how to simplify the app lock rules in the future?

Yeah it is but I already have some plans to simplify this.

KaterinaWire:
@zenkins release patch

PR #1114, opened 2021-01-13
Fix: replace UserType properties with ZMConversation parameter to ConversationLike protocol

billypchan:
## What's new in this PR?

Change `UserType` properties with `ZMConversation` parameters to new `ConversationLike` protocol to allow mocking `UserType` without creating a `ZMConversation` instance.

Before this PR if we want to mock a `UserType` method with `ZMConversation` parameter, e.g. 
```
    func canAddUser(to conversation: ZMConversation) -> Bool
```
We have to create an instance of `ZMConversation`, which is not necessary for returning a mock value. After this PR we can just pass a mocked subclass of `ConversationLike` in tests, then return the mocked value in Mocked User's conformed methods. 

Add `hasTeam` property to `UserType`


johnxnguyen (commented):
Could you elaborate what problem this is solving? I don't quite see how making the arguments optional makes it easier to mock.

johnxnguyen:
@billypchan please see the comment on my first review.

johnxnguyen (commented):
I'm concerned that we are making the `UserType` more confusing by making the arguments optional. I believe we should strive to maintain a clear and unambiguous API and that we shouldn't sacrifice this testability.

But that doesn't mean that we shouldn't improve testability, I would just suggest we consider an alternative solution, such as using a `ConversationType` (or perhaps a less confusing name) just like we do for `UserType`. Do you think that be a feasible approach?

billypchan:
@johnxnguyen I agree that `ConversationType` would be our final target, and I am creating a partially done protocol here:

https://github.com/wireapp/wire-ios/pull/4833/files#diff-be275f7def39aa5bedd90ad675663693f8fc3a4c5c07ca73236f2cc0d615e775

The problem currently is there are a lot of methods of `ZMConversation` having parameter type `ZMUser`, we have to change them to `UserType` before making everything in pure abstract interfaces.
I would suggest agree on this PR first, to solve the issue in https://github.com/wireapp/wire-ios/pull/4833. After we collected what we need in `ConversationType`, we can move it back to this framework and update these methods again.

johnxnguyen:
> The problem currently is there are a lot of methods of ZMConversation having parameter type ZMUser, we have to change them to UserType before making everything in pure abstract interfaces.

But do you need to make everything pure abstract interfaces for all of `ZMConversation`. Looking at the `UserType` methods that are affected here, it appears that you only use the properties `isSelfAnActiveMember` and `conversationType`.

typfel:
>But do you need to make everything pure abstract interfaces for all of ZMConversation. Looking at the UserType methods that are affected here, it appears that you only use the properties isSelfAnActiveMember and conversationType.

So this sounds feasible! I would rather have that we have an incomplete `ConversationType` protocol than that we start making argument optional which shouldn't be optional.

typfel (changes requested):
Use `ConversationType` protocol instead of making arguments optional.

Source/Model/Conversation/ConversationType.swift
```diff
@@ -0,0 +1,28 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+
+@objc
+public protocol ConversationType: NSObjectProtocol {
+    var conversationType: ZMConversationType { get }
```
johnxnguyen, line 24: Admittedly it's kind of confusing to have `ConversationType` mean "some conversation like object" and also a `ZMConversationType` mean "the type of a conversation".

I've seen the prefix "Like" to indicate that an object is "like" some abstract version of it. For example:

```swift
protocol ConversationLike {
  var conversationType: ZMConversationType { get }
  var isSelfAnActiveMember: Bool { get }
}

extension ZMConversation: ConversationLike { }
```

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -69,8 +69,8 @@ public extension ZMUser {
     }
     
     @objc(canAddUserToConversation:)
-    func canAddUser(to conversation: ZMConversation?) -> Bool {
-        guard let conversation = conversation,
+    func canAddUser(to conversation: ConversationType) -> Bool {
+        guard let conversation = conversation as? ZMConversation,
```
johnxnguyen, line 73: Why not pass these down without casting? It would save the duplication and it also means you don't have to assume what the concrete type is. If I remember correctly, when accessing a role you need to check that it equals the `ZMConversation` instance, but here you could just use referential equality instead of casting.

johnxnguyen (changes requested):

Source/Model/User/ZMUser+Permissions.swift
billypchan: Change `hasRoleWithAction` to save a casting.

Source/Model/Conversation/ConversationType.swift
billypchan: renamed for your suggestion.

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -191,12 +193,12 @@ public extension ZMUser {
         } else {
             return !isServiceUser // Bots are never guests
                 && ZMUser.selfUser(in: managedObjectContext!).hasTeam // There can't be guests in a team that doesn't exist
-                && conversation.localParticipants.contains(self)
+                && true == (conversation as? ZMConversation)?.localParticipants.contains(self)
```
johnxnguyen, line 196: Is there any way we can avoid this cast? Can we add one more requirement to `ConversationLike` to cover this, such as `isParticipant(user: UserType) -> Bool` ?

billypchan: added the new method for your suggestion. Actually, this checker is call in several places in the UI project and we can replace them.

Source/Model/Conversation/ConversationLike.swift
```diff
@@ -24,6 +24,13 @@ public protocol ConversationLike: NSObjectProtocol {
...
     var isSelfAnActiveMember: Bool { get }
     var teamRemoteIdentifier: UUID? { get }
+    
+    func localParticipantsContain(user: UserType) -> Bool
 }
 
-extension ZMConversation: ConversationLike {}
+extension ZMConversation: ConversationLike {
+    public func localParticipantsContain(user: UserType) -> Bool {
+        guard let user = user as? ZMUser else { return false }
```
johnxnguyen, line 33: Do we ever check if a search user is in the local participants?

billypchan: `localParticapants` should never contians search user since it has the type `Set<ZMUser>`

johnxnguyen: Yeah you're right, was just wondering.

johnxnguyen (approved):

typfel (approved):

billypchan:
@zenkins release major

PR #1115, opened 2021-01-15
[Fix] App lock passcode migration 

johnxnguyen:
### Issues

Patch migration tests were failing on the simulator.

### Causes

The shared container doesn't exist on the simulator and one of the app lock keychain migration tests was relying on it to create the account manager to migrate the custom passcode for all accounts at once.

### Solutions

It's not necessary to migrate all accounts at once, the migration will be performed individually for each account. Therefore, just migrate the passcode for the self user. To do this however, we must not delete the legacy passcode from the keychain.

I also decided not to hard crash if the migration fails. If it does fail, the worst that happens is that the user would need to create a new custom passcode, which isn't so bad.

### DIFF
--- a/Patches/TransferApplockKeychain.swift
+++ b/Patches/TransferApplockKeychain.swift
@@ -22,12 +22,7 @@ struct TransferApplockKeychain {
     
     static func migrateKeychainItems(in moc: NSManagedObjectContext) {
         migrateIsAppLockActiveState(in: moc)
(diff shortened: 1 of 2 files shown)


typfel (approved):

billypchan (approved):

johnxnguyen:
@zenkins release patch

PR #1116, opened 2021-01-15
Move MessageDetailsActionTests from UI project

billypchan:
## What's new in this PR?
Move `ZMConversationMessage` extension methods `canBeLiked`, `isSentBySelfUser`, `areMessageDetailsAvailable` & `areReadReceiptsDetailsAvailable` from UI project to data model since they are pure logics not related to UI.

Move the corresponding test`MessageDetailsActionTests` form UI project also to prevent testing `MessageDetailsActionTests` with creating `ZMConversation` instances, which can be created in `ZMConversationTestsBase`

### Discussion
- [ ] should we move the helper methods and mocking class to Wire testing to prevent duplicated code in UI project?


typfel:
I don't agree that all of these properties are not related to the UI. For example `areMessageDetailsAvailable` is boolean which computes if we can show a details screen for a given message, which to me is UI related logic.

Do have specific reason why you want to move these extensions?

billypchan:
@typfel the reason to open this PR is moving `MessageDetailsActionTests` to data model, to prevent creating `ZMConversation` in UI project tests, which can not be released in Xcode 12.

typfel:
> @typfel the reason to open this PR is moving `MessageDetailsActionTests` to data model, to prevent creating `ZMConversation` in UI project tests, which can not be released in Xcode 12.
> Another solution would be replacing `ZMConversation` with newly created `ConversationLike` inside `ZMConversationMessage` to allow mocking a conversation when creating `ZMConversationMessage` .

Yes, that sounds good!

billypchan:
continue in https://github.com/wireapp/wire-ios/pull/4853

PR #1117, opened 2021-01-18
[Fix] Can't import backups from old versions of Wire

typfel:
### Issues

When trying to import an old Wire build it will sometimes fail with the message:
"Something went wrong: your history couldn't be restored"

### Causes

Since https://github.com/wireapp/wire-ios-data-model/pull/1071 we open the backup db to prepare it for importing before we perform the actual import. When opening an old core data store it might need to be migrated before we can make any changes but we tried to open it with the following options:
```
```

This means that any backed up store which needs migration will fail to import.

### Solutions

Allow migration when opening store.


johnxnguyen (approved):

johnxnguyen:
**FYI** @typfel I updated the PR description because it said `supportsMigration: true`, and not `supportsMigration: false`

KaterinaWire (approved):

typfel:
@zenkins release patch

PR #1121, opened 2021-01-22
[Refactor] App lock controller

johnxnguyen:
## What's new in this PR?

This PR contains extensive refactoring and general clean up.

### Changes

Previously the app lock controller's main responsibilities were to 

- allow authentication with the OS using the LocalAuthentication framework
- answer simple questions such as whether it is active or not
- manage the persistence of the custom passcode

These changes give more responsibility to the app lock controller so that other components don't need to consider them, such as:

- Unlocking the app lock when authentication is successful
- Determining if the timeout is expired
- Expressing its state (locked or unlocked)
- Managing biometric state changes
- Authenticating via custom passcode

Other changes are for:

- Providing consistency in naming (e.g `isForce` and `enforceAppLock`)
- Reducing the ambiguity of names (e.g `requireBiometricOrCustomPasscode`)
- Separating code into different files for improved readability
- Moved nested types to the top level to avoid coupling of `AppLockType` to `AppLockController`
- Add a delegate to communicate when app lock is opened

### Notes

Sorry for the size of this PR, I was only planning a smaller number of changes but it grew unexpectedly!


marcomaddy1982 (approved):

marcomaddy1982:
Seems good for me. Thanks @johnxnguyen 👍

Tests/Source/Model/FeatureConfiguration/AppLock/AppLockControllerTests.swift
```diff
@@ -0,0 +1,431 @@
...
+        // When
+        let result = sut.evaluateAuthentication(customPasscode: "boo!")
+
+        // Then
+        XCTAssertEqual(result, .granted)
+        XCTAssertFalse(sut.isLocked)
+        XCTAssertTrue(mockBiometricsState.didCallPersistState)
+
+        // Clean up
+        try sut.deletePasscode()
```
KaterinaWire, line 290: What if we move this `Clean up passcode` to the `tearDown ()` since we are using it in multiple tests?

johnxnguyen: Ah I remember why I didn't do that, because the `sut` is created at the start of each tests so we need to do the tear down in the test itself.

KaterinaWire (approved):

Source/Model/FeatureConfig/AppLock/AppLockController.swift
```diff
@@ -133,6 +133,7 @@ public final class AppLockController: AppLockType {
     // MARK: - Methods
 
     public func beginTimer() {
+        guard state == .unlocked else { return }
```
johnxnguyen, line 136: @KaterinaWire adding this guard fixes the issue you found where you could enter the app after cancelling authentication. This method is called when the app resigns active, which happens when displaying the TouchID / FaceID UI. Without this guard, calling this method would set the last checkpoint to now. This should only happen when the app is unlocked, since the checkpoint should be the point in time that you left the app in an unlocked state.

KaterinaWire (approved):

PR #1126, opened 2021-01-26
Move Conversation properties to ConversationLike

billypchan:
## What's new in this PR?

Move Conversation form UI project protocol to `ConversationLike`. Put the Swift only property in extension to protocol `SwiftConversationLike` and create a new alias `Conversation` to combine them.

### DIFF
--- a/Source/Model/Conversation/ConversationLike.swift
+++ b/Source/Model/Conversation/ConversationLike.swift
@@ -19,6 +19,8 @@
 
 import Foundation
 
+typealias Conversation = ConversationLike & SwiftConversationLike
+
 @objc
 public protocol ConversationLike: NSObjectProtocol {
     var conversationType: ZMConversationType { get }
@@ -32,6 +34,8 @@ public protocol ConversationLike: NSObjectProtocol {
     var allowGuests: Bool { get }
     var team: Team? { get }
 
+    var isUnderLegalHold: Bool { get }
+    var securityLevel: ZMConversationSecurityLevel { get }
 }
 
 // Since ConversationLike must have @objc signature(@objc UserType has a ConversationLike property), create another protocol to abstract Swift only properties
--- a/Source/Model/Conversation/ConversationLike.swift
+++ b/Source/Model/Conversation/ConversationLike.swift
@@ -26,11 +26,28 @@ public protocol ConversationLike: NSObjectProtocol {
     var teamRemoteIdentifier: UUID? { get }
     
     func localParticipantsContain(user: UserType) -> Bool
+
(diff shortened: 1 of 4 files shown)


Source/Model/Conversation/ConversationLike.swift
```diff
@@ -19,18 +19,40 @@
 
 import Foundation
 
+public typealias Conversation = ConversationLike & SwiftConversationLike
```
billypchan, line 22: I suggest use simply `Conversation` as the protocol name. Since we have name tradition e.g. `ZMConversationMessage` to Message managed object.

johnxnguyen (approved):
It would be nice to document each protocol requirement.

marcomaddy1982 (approved):

Source/Model/Conversation/ConversationLike.swift
```diff
@@ -19,18 +19,40 @@
...
     var conversationType: ZMConversationType { get }
     var isSelfAnActiveMember: Bool { get }
     var teamRemoteIdentifier: UUID? { get }
     
     func localParticipantsContain(user: UserType) -> Bool
+
+    var displayName: String { get }
+    var connectedUserType: UserType? { get }
+    var allowGuests: Bool { get }
+    var team: Team? { get }
```
typfel, line 35: Could we use `TeamType` here?

Source/Model/Conversation/ConversationLike.swift
```diff
@@ -19,18 +19,40 @@
 
 import Foundation
 
+public typealias Conversation = ConversationLike & SwiftConversationLike
+
 @objc
```
typfel, line 24: Question: Why do we need to expose `ConversationLike` to objective-c now that the whole UI project is in Swift?

typfel (changes requested):

Source/Model/Conversation/ConversationLike.swift
billypchan: replaced with `TeamType`

Source/Model/Conversation/ConversationLike.swift
billypchan: `UserType` has to be @objc and has several @objc methods (e.g. `@objc(canAddUserToConversation:)`) and so we still need a @objc `ConversationLike`.

`SwiftConversationLike` is for swift `ZMConversation`'s swift extensions.

typfel: Ok, we should review those `@objc` annotation at some point and see if we can remove them.

PR #1127, opened 2021-01-27
[Fix] Existing system message changing type due to changed enum order

typfel:
### Issues

Existing system message would change type after upgrading to the latest version of Wire

### Causes

New entries were added to the `ZMSystemMessageType` enum which is represented as integer in the database. Adding a new entry to the enum changes the integer associated for all cases after the new entry:

```
enum Foo {
case A = 0
case B = 1
}

enum Foo (edited) {
case A = 0
case C = 1
case B = 2
}
```

### Solutions

Add new entries at the end of the enum.


marcomaddy1982 (approved):

johnxnguyen (approved):

David-Henner (approved):

typfel:
@zenkins release patch

PR #1128, opened 2021-01-28
Add quoteMessage to ZMTextMessageData

billypchan:
## What's new in this PR?

Add `quoteMessage` with type `ZMConversationMessage` to `ZMTextMessageData` to allow mocking `quota` with creating a `ZMMessage` managed object.

### DIFF
--- a/Source/Public/ZMMessage.h
+++ b/Source/Public/ZMMessage.h
@@ -98,9 +98,6 @@ typedef NS_ENUM(int16_t, ZMSystemMessageType) {
 @property (nonatomic, readonly, nullable) NSString *messageText;
 @property (nonatomic, readonly, nullable) LinkMetadata *linkPreview;
 @property (nonatomic, readonly, nonnull) NSArray<Mention *> *mentions;
-
-@property (nonatomic, readonly, nullable) ZMMessage *quote;
-/// Quote message with ZMConversationMessage protocol type, returns casted quote
 @property (nonatomic, readonly, nullable) id<ZMConversationMessage> quoteMessage;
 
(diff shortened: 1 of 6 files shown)


marcomaddy1982 (approved):

Source/Public/ZMMessage.h
```diff
@@ -97,7 +98,10 @@ typedef NS_ENUM(int16_t, ZMSystemMessageType) {
 @property (nonatomic, readonly, nullable) NSString *messageText;
 @property (nonatomic, readonly, nullable) LinkMetadata *linkPreview;
 @property (nonatomic, readonly, nonnull) NSArray<Mention *> *mentions;
+
 @property (nonatomic, readonly, nullable) ZMMessage *quote;
```
typfel, line 102: Can't we remove this? Seems like most references to `quote` in the UI is anyway expecting a `ZMConversationMessage`. 

And if we can remove it you could keep the `quote` name.

typfel (changes requested):

Source/Public/ZMMessage.h
billypchan: We need to reserve `ZMMessage *quote` since `ZMClientMessage` conforms `ZMTextMessageData` and its has ` @NSManaged public var quote: ZMMessage?`

typfel: Ah I see but we could still remove `ZMMessage *quote` from protocol?

billypchan: It is removed, thanks!

typfel (approved):

PR #1129, opened 2021-01-28
[Refactor] Remove unused system messages append methods

typfel:
## What's new in this PR?

Remove methods for inserting .`reactivatedDevice` and `.usingNewDevice` system messages.

See: https://github.com/wireapp/wire-ios-sync-engine/pull/1385


marcomaddy1982 (approved):

David-Henner (approved):

Source/Public/ZMMessage.h
```diff
@@ -75,8 +75,8 @@ typedef NS_ENUM(int16_t, ZMSystemMessageType) {
     ZMSystemMessageTypeDecryptionFailed,
     ZMSystemMessageTypeDecryptionFailed_RemoteIdentityChanged,
     ZMSystemMessageTypeNewConversation,
-    ZMSystemMessageTypeReactivatedDevice,
-    ZMSystemMessageTypeUsingNewDevice,
+    ZMSystemMessageTypeReactivatedDevice __deprecated_enum_msg("Devices can't be reactivated any longer"),
```
David-Henner, line 78: did you mark this as deprecated because the other cases need to keep the same Integer value ?

typfel: Yes exactly and users will still also have these messages in their history so still need to be able to render them.

typfel:
@zenkins release major

PR #1131, opened 2021-02-01
Define AnalyticsLike (Part 1)

johnxnguyen:
## What's new in this PR?

This is a part of an ongoing effort to refactor the analytics system to align it to the implementation spec and improving and cleaning up the code along the way.

### Changes

**Define a new and simpler analytics interface.** It may be extended later if needed and will eventually replace `AnalyticsType`.

**Add type safety.** At the lowest level, events and attributes are stringly typed. By defining types for events, attribute keys and attribute values, we can avoid typo errors that lead to inconsistencies. By defining `AnalyticsAttributeKey` as a struct, we are not limiting the range of possible keys (as would happen with an enum) and can use static constants to improve usability, such as

```swift
extension AnalyticsAttributeKey {

  static let conversationType = AnalyticsAttributeKey(rawValue: "conversation-type")

}
```

### DIFF
--- a/Source/Model/Analytics/AnalyticsLike.swift
+++ b/Source/Model/Analytics/AnalyticsLike.swift
@@ -1,0 +1,86 @@
+//
+// Wire
+// Copyright (C) 2021 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


Source/Model/Analytics/AnalyticsLike.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+}
+
+public typealias AnalyticsAttributes = [AnalyticsAttributeKey: AnalyticsAttributeValue]
+
+public extension AnalyticsAttributes {
+
+    var rawValue: [String: String] {
+        return mapKeys(\.rawValue).mapValues(\.rawValue)
```
KaterinaWire, line 57: I noticed that we also have a 
```
func mapKeysAndValues<NewKey, NewValue>(keysMapping: ((Key) -> NewKey),
                                                   valueMapping: ((Key, Value) -> NewValue?))
```

KaterinaWire (approved):

Source/Model/Analytics/AnalyticsLike.swift
johnxnguyen: Yeah I've seen that too but I wanted to avoid depending on something that I don't need to.

David-Henner (approved):

PR #1133, opened 2021-02-02
[Fix] Avoid fetch request when processing new message event

typfel:
### Issues

When processing an update event for a new message we always perform a fetch request attempting to retrieve the message.

### Causes

When processing a batch of update events we prefetch any messages associated with the update events before we start the processing. New messages will naturally never be prefetched so this optimising strategy only work when process update events updating an already existing message.

### Solutions

Assume we are dealing with a new message if a message is missing from the set of prefetched messages.


Source/Model/Message/ZMMessage.m
```diff
@@ -410,23 +410,43 @@ + (instancetype)fetchMessageWithNonce:(NSUUID *)nonce
...
+        
+        if (nil != prefetchedMessages) {
+            for (ZMMessage *prefetchedMessage in prefetchedMessages) {
+                if ([prefetchedMessage isKindOfClass:[self class]]) {
+                    return prefetchedMessage;
+                }
             }
         }
+        
+        if (assumeMissingIfNotPrefetched) {
```
marcomaddy1982, line 447: Probably we could move this immediately after `if (prefetchResult != nil) {`, so in case `assumeMissingIfNotPrefetched` is `true` we directly return `nil` and we avoid to do the other operation that could be time consuming.

typfel: 👍 

I also removed the `if (prefetchResult != nil)` since it's actually unnecessary.

marcomaddy1982: Question: couldn't if (assumeMissingIfNotPrefetched) { return nil } be done as first check in the method?

typfel: No, we must search through the prefetch results before we assume that the messages doesn't exist.

typfel: Also, I changed the logic so that if no prefetch results are injected the `assumeMissingIfNotPrefetched` is ignored.

marcomaddy1982: Yes ok, now it is clear.

marcomaddy1982 (approved):

KaterinaWire (approved):

typfel:
@KaterinaWire @marcomaddy1982 I found a bug while testing this optimisation.

### Problem
If you receive a file message, two messages would get inserted.

### Cause
File messages are sent via multiple messages:
1. Preview when which is sent while the file is uploaded to the server
2. Final message when the message is uploaded to server

These messages are actually updates to the same message so they share the same message nonce (message ID). Now if these message are processed in the same batch of events they will not be included in the prefetch results which causes us to insert a new message also for second "update" message.

### Solution
Update the prefetch results during the event processing with any newly inserted messages.

KaterinaWire (approved):

KaterinaWire:
> @KaterinaWire @marcomaddy1982 I found a bug while testing this optimisation.
> 
> ### Problem
> If you receive a file message, two messages would get inserted.
> 
> ### Cause
> File messages are sent via multiple messages:
> 
> 1. Preview when which is sent while the file is uploaded to the server
> 2. Final message when the message is uploaded to server
> 
> These messages are actually updates to the same message so they share the same message nonce (message ID). Now if these message are processed in the same batch of events they will not be included in the prefetch results which causes us to insert a new message also for second "update" message.
> 
> ### Solution
> Update the prefetch results during the event processing with any newly inserted messages.

@typfel Did you find it through manual testing? Maybe it makes sense to add some tests for this case?

typfel:
> > @KaterinaWire @marcomaddy1982 I found a bug while testing this optimisation.
> > ### Problem
> > If you receive a file message, two messages would get inserted.
> > ### Cause
> > File messages are sent via multiple messages:
> > 
> > 1. Preview when which is sent while the file is uploaded to the server
> > 2. Final message when the message is uploaded to server
> > 
> > These messages are actually updates to the same message so they share the same message nonce (message ID). Now if these message are processed in the same batch of events they will not be included in the prefetch results which causes us to insert a new message also for second "update" message.
> > ### Solution
> > Update the prefetch results during the event processing with any newly inserted messages.
> 
> @typfel Did you find it through manual testing? Maybe it makes sense to add some tests for this case?

Yes, through manual testing. I'll add a test case.

Tests/Source/Model/Messages/ZMClientMessageTests+Prefetching.swift
```diff
@@ -0,0 +1,89 @@
...
+                                      senderClientID: senderClientID)
+        
+        // when
+        var message: ZMOTRMessage?
+        performPretendingUiMocIsSyncMoc {
+            message = ZMOTRMessage.createOrUpdate(from: event, in: self.uiMOC, prefetchResult: prefetchResults)
+        }
+        
+        // then
+        XCTAssertTrue(existingMessage === message)
```
marcomaddy1982, line 59: Nice having  the test with the identity operator `===` 👍

marcomaddy1982 (approved):

marcomaddy1982:
@typfel after your comment about the bug you found and the tests you added it seems fine for me. Thanks 👍

typfel:
@zenkins release patch

PR #1135, opened 2021-02-03
[Fix] Feature insertion race condition

johnxnguyen:
### Issues

We assert that there can only be a maximum of one instance of `Feature` for any given feature name, however this assertion is failing sometimes.

### Causes

While we have checks to only insert a feature instance if none is present, it is possible that two instance can be created simultaneously from different contexts.

### Solutions

Previously we were trying to create default instances when trying to fetch an instance and finding that it wasn't created yet. This is a **lazy** approach.This over-complicates things and makes it difficult to reason about when we're inserting feature instances and in which context.

Instead I think we should take an **eager** approach and try to create the default instances (if needed) as early as possible so that it is available for use immediately. This simplifies things a lot since we can create a single method to create the default instances which will do all the checking internally. This will be the only way that an instance can be created. Furthermore, we restrict using this method on the sync context only to ensure that two instances can't be created simultaneously.

### DIFF
--- a/Patches/InvalidFeatureRemoval.swift
+++ b/Patches/InvalidFeatureRemoval.swift
@@ -1,0 +1,35 @@
+//
+// Wire
+// Copyright (C) 2021 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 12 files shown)


typfel (approved):

KaterinaWire (approved):

johnxnguyen:
@zenkins release major

PR #1140, opened 2021-02-16
Define RoundedInt (Part 3)

johnxnguyen:
## What's new in this PR?

In order to protect the privacy of BI data, numeric attribute values should be rounded. Previously this was done on the UI side by adding an extension to `Int` to round the value. This is error prone because you need to remember to round the int value.

A more type safe solution is to remove the `Int` conformance of `AnalyticsAttributeValue` and instead define a `RoundedInt` container.

### DIFF
--- a/Source/Model/Analytics/AnalyticsAttributeValue.swift
+++ b/Source/Model/Analytics/AnalyticsAttributeValue.swift
@@ -54,6 +54,14 @@ public struct RoundedInt: AnalyticsAttributeValue {
 
 }
 
+public extension Int {
+
+    func rounded(byFactor factor: Int) -> RoundedInt {
+        return RoundedInt(self, factor: factor)
+    }
+
+}
+
 extension UUID: AnalyticsAttributeValue {
 
     public var analyticsValue: String {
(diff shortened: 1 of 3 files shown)


Source/Model/Analytics/AnalyticsAttributeValue.swift
```diff
@@ -28,18 +28,36 @@ public protocol AnalyticsAttributeValue {
...
+
+    public let analyticsValue: String
+
+    /// Create a rounded integer by a certain factor.
+    ///
+    /// - Parameters:
+    ///     - exactValue: The integer value to round.
+    ///     - factor: Determines how much the value is rounded.
+
+    public init(_ exactValue: Int, factor: Int) {
```
KaterinaWire, line 48: Are you going to make tests for this logic in the UI project?

johnxnguyen: No I should add them here, thanks for the reminder 👍🏻

johnxnguyen: Test added.

David-Henner (approved):

KaterinaWire (approved):

PR #1143, opened 2021-02-22
[Feature] Support for qualified user IDs

typfel:
## What's new in this PR?

Add support for parsing and storing the qualified user ID of users, new endpoints will require the use of qualified IDs so this is a preparation task for using those new end-points.

A qualified ID is the same as the old `id` with the addition of a `domain` field which specifies which domain the user belongs to.


Tests/Source/Model/User/ZMUserTests+Swift.swift
```diff
@@ -956,3 +956,13 @@ extension ZMUserTests_Swift {
     }
 
 }
+
+// MARK: - QualifiedID
+
+extension ZMUserTests_Swift {
+    
+    func testThatUserIsAssignedQualifiedID_WhenAvailable() {
```
marcomaddy1982, line 964: Missing something or we can just delete it?

typfel: Ah yes was planning on adding a swift test but it made little sense since all the setup code was in Objective-C. I'll clean this up.

marcomaddy1982 (approved):

johnxnguyen (approved):

typfel:
@zenkins release major

PR #1144, opened 2021-03-01
[Fix]: Handle gracefully when device runs out of disk space

marcomaddy1982:
### Issues

See the [issue](https://wearezeta.atlassian.net/browse/SQCORE-429)

When there was a problem to load the database in case we run out of disk space (or for any other error), the database was  dropped. This means that the user was losing all the previous conversation

### Solutions

Now a screen similar to the `blacklisted` or `jailbroken` is shown but with a different message. Moreover the alert contains two buttons:
- Device Settings -> Which will show the application settings
- Delete Database -> When the user will choose it another confirmation alert will be show. If the user confirms the database will be removed and the login screen will be shown.


typfel (commented):
Did you investigate if it's possible to add test for checking that the callback gets called when the database can't be loaded?

marcomaddy1982:
> Did you investigate if it's possibility to add test for checking that the callback gets called when the database can't be loaded?

Yes I did. 

@typfel Anyway there are some ongoing discussions with the design team so stuffs could changes in the next day.

typfel:
> > Did you investigate if it's possibility to add test for checking that the callback gets called when the database can't be loaded?
> 
> Yes I did.

And what was the result? Not possible at all or just a lot of work?

marcomaddy1982:
> > > Did you investigate if it's possibility to add test for checking that the callback gets called when the database can't be loaded?
> > 
> > 
> > Yes I did.
> 
> And what was the result? Not possible at all or just a lot of work?

I didn't find any way to test it.

typfel (approved):

KaterinaWire (approved):

Source/ManagedObjectContext/PersistentStorageInitialization.swift
```diff
@@ -57,15 +61,19 @@ extension NSPersistentStoreCoordinator {
...
-            completionHandler(NSPersistentStoreCoordinator(
+
+            guard let persistentStoreCoordinator = NSPersistentStoreCoordinator(
                 storeFile: storeFile,
-                accountIdentifier: accountIdentifier,
+                accountIdentifier: nil,
                 applicationContainer: applicationContainer,
                 model: model,
-                startedMigrationCallback: startedMigrationCallback,
+                startedMigrationCallback: nil,
```
typfel, line 70: Why is this nil now?

typfel, line 70: Why is this nil now?

Source/ManagedObjectContext/PersistentStorageInitialization.swift
```diff
@@ -57,15 +61,19 @@ extension NSPersistentStoreCoordinator {
         PersistentStorageInitialization.executeWhenFileIsAccessible(storeFile) {
             let model = NSManagedObjectModel.loadModel()
             UserClientKeysStore.migrateIfNeeded(accountIdentifier: accountIdentifier, accountDirectory: accountDirectory, applicationContainer: applicationContainer)
-            
-            completionHandler(NSPersistentStoreCoordinator(
+
+            guard let persistentStoreCoordinator = NSPersistentStoreCoordinator(
                 storeFile: storeFile,
-                accountIdentifier: accountIdentifier,
+                accountIdentifier: nil,
```
typfel, line 67: Why is this nil now?

Source/ManagedObjectContext/PersistentStorageInitialization.swift
```diff
@@ -31,14 +31,18 @@ extension NSPersistentStoreCoordinator {
...
+            guard let persistentStoreCoordinator = NSPersistentStoreCoordinator(
                 storeFile: storeFile,
                 accountIdentifier: nil,
                 applicationContainer: applicationContainer,
                 model: model,
                 startedMigrationCallback: nil,
                 databaseLoadingFailureCallBack: nil
-            ))
+            ) else {
+                return
```
typfel, line 42: Returning without calling the completion handler is something which can lead to bugs for the user of this API since it's often not something you consider when using such an API.

Maybe it's better if we bubble any error all the way up to SE by marking this method as throwing instead of having the `databaseLoadingFailureCallBack`?

Source/ManagedObjectContext/NSPersistentStoreCoordinator+Wire.swift
```diff
@@ -81,16 +91,10 @@ extension NSPersistentStoreCoordinator {
     }
     
     /// Adds the persistent store
-    fileprivate func addPersistentStoreWithNoMigration(at url: URL,
-                                                       databaseLoadingFailureCallBack: (() -> Void)?) {
+    fileprivate func addPersistentStoreWithNoMigration(at url: URL) throws {
         let options = NSPersistentStoreCoordinator.persistentStoreOptions(supportsMigration: false)
         let addStore = { _ = try self.addPersistentStore(ofType: NSSQLiteStoreType, configurationName: nil, at: url, options: options) }
```
typfel, line 96: No need for this closure now that only use it once.

Source/ManagedObjectContext/NSPersistentStoreCoordinator+Wire.swift
```diff
@@ -47,22 +47,32 @@ extension NSPersistentStoreCoordinator {
...
+
+        do {
+            if NSPersistentStoreCoordinator.shouldMigrateStoreToNewModelVersion(at: storeURL,
+                                                                                model: model) {
+                startedMigrationCallback()
+                try self.migrateAndAddPersistentStore(url: storeURL)
+            } else {
+                try self.addPersistentStoreWithNoMigration(at: storeURL)
+            }
+            return true
```
typfel, line 72: Returning true / false to signal success or failure is often confusing. I'd suggest letting the error bubble up instead.

typfel (changes requested):

Source/ManagedObjectContext/PersistentStorageInitialization.swift
marcomaddy1982: ops my bad, I don't know what's happened!! Same next comment!!!

Source/ManagedObjectContext/PersistentStorageInitialization.swift
marcomaddy1982: I understand you suggestion. Anyway I have a question: So you mean also that I should not have the `optional` `convenience init?(` on the `NSPersistentStoreCoordinator` and return in any case the persistentStoreCoordinator?

The reason why I adopted that approach is because when the database is deleted and the  `completionHandler(persistentStoreCoordinator)` is executed also a crash is happening.

typfel: Or maybe this is not working to due to it being asynchronous? I haven't checked it, was just an idea.

marcomaddy1982: > I'm suggesting to make the init() throwing instead of optional (and every method above). The catching of the error would then happen SE.

Ok this is a clear. Thanks. And if I understood correctly in the `catch`  together with the `databaseLoadingFailureCallBack()` I should execute the `completionHandler(persistentStoreCoordinator)` which before leads to have a crash.

> Or maybe this is not working to due to it being asynchronous? I haven't checked it, was just an idea.

This is a good point.

typfel: Hmm but the only reason why this is asynchronous is due to it being wrapped inside `executeWhenFileIsAccessible`.  But maybe we can remove it now after https://github.com/wireapp/wire-ios/pull/4794 which doesn't even create the SessionManaged until the file system is accessible.

marcomaddy1982: Ok I can have a look.

Source/ManagedObjectContext/NSPersistentStoreCoordinator+Wire.swift
marcomaddy1982: Done

Source/ManagedObjectContext/PersistentStorageInitialization.swift
```diff
@@ -49,40 +52,25 @@ extension NSPersistentStoreCoordinator {
...
-    fileprivate static func executeWhenFileIsAccessible(_ file: URL, usingBlock block: @escaping () -> Void) {
-        // We need to handle the case when the database file is encrypted by iOS and user never entered the passcode
-        // We use default core data protection mode NSFileProtectionCompleteUntilFirstUserAuthentication
-        let storageInitialization = PersistentStorageInitialization()
-        
-        storageInitialization.applicationProtectedDataDidBecomeAvailableObserver = FileManager.default.executeWhenFileSystemIsAccessible { 
-            storageInitialization.applicationProtectedDataDidBecomeAvailableObserver = nil
-            block()
+                startedMigrationCallback: startedMigrationCallback)
+            completionHandler(persistentStoreCoordinator)
```
typfel, line 70: Since this method is synchronous now you can return `persistentStoreCoordinator` instead using a completionHandler. You would then also need to cascade the `throws` further up.

typfel (changes requested):

Source/ManagedObjectContext/PersistentStorageInitialization.swift
marcomaddy1982: All done!!!

KaterinaWire (approved):

typfel (approved):

marcomaddy1982:
@zenkins release major

PR #1148, opened 2021-03-11
[Fix] Use optional getter for time field since it's not always present

typfel:
### Issues

Warning were logged when attempting to fetch the timestamp form events which doesn't include the time field.

### Causes

We were using the the non-optional method `[NSDictionary dateFor:]` for the time field is not always guaranteed to be present.

### Solutions

Use the optional getter `[NSDictionary optionalDateFor:]`.


johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1152, opened 2021-03-24
[Fix] Restore logic for only migrating when model version changed

typfel:
### Issues

After https://github.com/wireapp/wire-ios-data-model/pull/1144 we now show the DB migration screen on every app launch.

### Causes

We removed a rule for only migrating the database when the model version has changed since that path killing the app on any error.

### Solutions

The path for killing the app no longer exists after some further refactoring so we can restore the old behaviour of only attempting a migration when the model version changed.


marcomaddy1982 (approved):

johnxnguyen (approved):

typfel:
@zenkins release patch

PR #1153, opened 2021-03-25
[Feature] Add isFederated on the UserType

typfel:
## What's new in this PR?

- Add isFederated on the UserType, which indicates if another user comes from a different domain.
- Support qualified IDs in search users


Source/Model/User/ZMSearchUser.swift
```diff
@@ -165,7 +166,7 @@ public class ZMSearchUser: NSObject, UserType {
     
     public var domain: String? {
         get {
-            return user?.domain
+            return user != nil ? user?.domain : internalDomain
```
johnxnguyen, line 169: Can this be simplified to  `return user?.domain ?? internalDomain` ?

Source/Model/User/UserType+Federation.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+extension UserType {
+
+    func isFederating(with otherUser: UserType) -> Bool {
```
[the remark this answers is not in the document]
johnxnguyen: Federation allows users that belong to different backend instances to communicate with each other.

The verb `federate` means (with respect to a number of organisations) to form or be formed into a single centralized unit. I believe the intention here is to express that two users are federating together, rather than one user federating another user, which doesn't make any sense. So I would say that the use of "with" is appropriate.

johnxnguyen (approved):

Source/Model/User/UserType+Federation.swift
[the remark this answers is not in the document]
David-Henner: Thanks for the explanation! I agree then that in this case, using `with` is totally correct

David-Henner (approved):

Source/Model/User/ZMSearchUser.swift
typfel: Yes, in this case for sure. It's not completely equivalent though. I was following the pattern of the other properties when implementing this.

I'll review and see if I can do this for all properties.

typfel: Done

typfel:
@zenkins release major

PR #1156, opened 2021-04-29
[Feature] Add a new state to the ExternalParticipantsState for a conversation

KaterinaWire:
## What's new in this PR?

As part of the UI improvement, we should update the "External users" banner and display an additional case where external partners are present in the conversation. For this reason, we need to extend the existing  `ExternalParticipantsState`. 

https://wearezeta.atlassian.net/browse/SQSERVICES-394


typfel (approved):

johnxnguyen (approved):

KaterinaWire:
@zenkins release minor

PR #1157, opened 2021-05-10
[Fix] Conversation search

KaterinaWire:
### Issues
The search result for the query `9:3` will be empty, even if there's a conversation named `Feature [9:30]`.

https://wearezeta.atlassian.net/browse/SQSERVICES-65

### Causes
For the conversation search, we use a predicate that compares normalized conversation names (without punctuation characters) to query strings that may contain punctuation characters.

### Solutions
Fix a predicate that uses query strings without punctuation characters.


billypchan (approved):

KaterinaWire:
@zenkins release patch

PR #1159, opened 2021-05-22
[Feature] Update client capabilities

KaterinaWire:
## What's new in this PR?

All updated clients that can support the new BE end-points, related to legal hold should send
`"capabilities": [ "legalhold-implicit-consent"]` to BE. To track and notify clients when they need to update capabilities, a new `UserClient` property has been introduced in this PR.

## Notes
https://wearezeta.atlassian.net/browse/SQSERVICES-417


Source/Model/UserClient/UserClient.swift
```diff
@@ -748,3 +755,16 @@ extension UserClient {
...
+// MARK: - Capabilities
+extension UserClient {
+
+    public static func updateCapabilities(_ context: NSManagedObjectContext) {
+        guard let selfClient = ZMUser.selfUser(in: context).selfClient() else { return }
+
+        selfClient.needsToUpdateCapabilities = true
+        selfClient.setLocallyModifiedKeys(Set(arrayLiteral: ZMUserClientNeedsToUpdateCapabilitiesKey))
+
+        context.enqueueDelayedSave()
```
johnxnguyen, line 767: I suggest to rename this to something like `triggerSelfClientCapabilityUpdate`.

KaterinaWire: Done

johnxnguyen (approved):

typfel (approved):

PR #1160, opened 2021-05-24
[Feature] Add forced connection block status

KaterinaWire:
## What's new in this PR?

BE will add new connection `"missing-legalhold-consent"` status. It means that the connection is blocked due to missing legal hold consent.

## Notes
https://wearezeta.atlassian.net/browse/SQSERVICES-429


Source/Model/User/UserType.swift
```diff
@@ -221,6 +221,9 @@ public protocol UserConnections {
     
     ///  Whether the user has been blocked by the self user
     var isBlocked: Bool { get }
+
+    ///  Whether the user has been forced blocked due to missing legal hold consent
+    var isMissingLegalholdConsent: Bool { get }
```
johnxnguyen, line 226: I suggest to not mention force blocking in the comment, but instead just describe what the consent is for. Something like "Whether the user consents to being exposed to legal hold".

The reason being is the abstraction should describe the 'what' and not the 'how'. In the future I see this property being backed not by the connect status but instead by a consent checkbox. What do you think?

Source/Model/User/ZMSearchUser.swift
```diff
@@ -292,6 +292,10 @@ public class ZMSearchUser: NSObject, UserType {
         return user?.isBlocked == true
     }
 
+    public var isMissingLegalholdConsent: Bool {
+        return user?.isMissingLegalholdConsent == true
+    }
```
johnxnguyen, line 297: If there is no `user` then `isMissingLegalholdConsent` will be `true`. I think it safer to always assume no consent when in doubt. You could do this explicitly with a guard, or you can invert the flag to mean `consentsToLegalHoldExposure`. Personally I prefer boolean flags to have a single positive semantic (`isSomething == true`), rather than a double negative (`isNotSomething == false`).

Source/Model/User/UserType.swift
johnxnguyen: Are we spelling legal hold as `LegalHold` or `Legalhold`? I see it differently across platforms, but we should be consistent in our own code.

johnxnguyen: Also, the user type is used in SE tests, so you'll need to update conformance there.

KaterinaWire: I will check and rename `LegalHold` in SE and UI.

KaterinaWire: I agree that the`consentsToLegalHoldExposure` property name is clearer as we are going to set this property from the settings in the future. 
I updated this PR.

Source/Model/User/ZMSearchUser.swift
KaterinaWire: I renamed it and added a guard

johnxnguyen (approved):

Source/Model/User/ZMUser.m
```diff
@@ -871,20 +871,36 @@ @implementation ZMUser (Connections)
...
 - (BOOL)isBlocked
 {
-    return self.connection != nil && (self.connection.status == ZMConnectionStatusBlocked || self.connection.status == ZMConnectionStatusBlockedMissingLegalholdConsent);
+    switch (self.blockStateReason) {
+        case ZMBlockStateReasonBlocked:
+        case ZMBlockStateReasonBlockedMissingLegalholdConsent:
+            return YES;
+        case ZMBlockStateReasonNone:
+            return NO;
+    }
```
johnxnguyen, line 880: I think it would be safer to only consider the existence, i.e `return self.blockStateReason != ZMBlockStateReasonNone;`

johnxnguyen: And, you can move this implementation to an extension of `UserConnections` so you don't have to reimplement it for every conforming type.

Source/Public/ZMUser.h
```diff
@@ -30,6 +30,12 @@
 
 extern NSString * _Nonnull const ZMPersistedClientIdKey;
 
+typedef NS_ENUM(int16_t, ZMBlockStateReason) {
+    ZMBlockStateReasonNone = 0,
+    ZMBlockStateReasonBlocked, ///< We have blocked this user
+    ZMBlockStateReasonBlockedMissingLegalholdConsent, ///< The user is blocked due to legal hold missing consent
+};
```
johnxnguyen, line 37: It's a little bit confusing because the type name suggests that these are all reasons, but `ZMBlockStateReasonBlocked` isn't a reason. I understand you need this case because `isBlocked` depends in it to work. But perhaps naming this type something like `ZMBlockState` is more suitable.

Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -1300,12 +1300,12 @@ -(NSString *)managedByString:(ZMUser *)user {
...
-
-
-
-@implementation ZMUserTests (Connections)
+//@end
+//
+//
+//
+//
+//@implementation ZMUserTests (Connections)
```
johnxnguyen: To delete?

Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -1843,11 +1844,11 @@ - (void)testThatItCanNotCancelIgnoredConnectionRequest
...
-@end
-
-
-
-@implementation ZMUserTests (Validation)
+//@end
+//
+//
+//
+//@implementation ZMUserTests (Validation)
```
johnxnguyen: To delete?

Tests/Source/Model/User/ZMUserTests.m
```diff
@@ -2379,11 +2380,11 @@ - (void)testThatItValidatesAValidPassword
...
-@end
-
-
-
-@implementation ZMUserTests (KeyValueObserving)
+//@end
+//
+//
+//
+//@implementation ZMUserTests (KeyValueObserving)
```
johnxnguyen: To delete?

KaterinaWire: Removed

Tests/Source/Model/User/ZMUserTests.m
KaterinaWire: Removed

Source/Model/User/ZMUser.m
KaterinaWire: What do you mean to move to `UserConnections`?

Source/Public/ZMUser.h
KaterinaWire: Done

PR #1168, opened 2021-06-21
[Feature] Support for qualified conversations

typfel:
## What's new in this PR?

- Add `domain` to the `Conversation` entity
- Update API for fetching and creating users to optionally include the domain

The domain `parameter` is included when processing events and responses which include a qualified ID. Since we still needs to be backwards compatible with old servers the domain will remain optional for now. 

The API for creating a new user/conversation has been changed so that it returns a non-optional `ZMUser` or `ZMConversation`.


Tests/Source/Model/ZMManagedObjectFetchingTests.swift
```diff
@@ -0,0 +1,179 @@
...
+        XCTAssertEqual(user, fetched)
+    }
+
+    func testItFetchesEntityByRemoteIdentifier_WhenObjectIsNotRegisteredInContext() throws {
+        // given
+        let uuid = UUID()
+        let user = ZMUser.insertNewObject(in: mocs.viewContext)
+        user.remoteIdentifier = uuid
+        try mocs.viewContext.save()
+        mocs.viewContext.refresh(user, mergeChanges: false)
```
johnxnguyen, line 57: Why do these two lines make the object non registered?

Source/Model/Conversation/ZMConversation+Creation.swift
```diff
@@ -20,6 +20,34 @@ import Foundation
 
 
 extension ZMConversation {
+
+    @objc public static func fetchOrCreate(with remoteIdentifier: UUID,
+                                           domain: String?,
+                                           in context: NSManagedObjectContext) -> ZMConversation {
+        var created: Bool = false
```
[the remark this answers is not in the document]
typfel: No, I'm just declaring it to fill in all the parameters. The point of this method is expose a version without the `created` parameter since it's not relevant in most cases.

Same thing this applies for the ZMUser variant.

Source/Model/User/ZMUser+Create.swift
```diff
@@ -19,6 +19,34 @@
...
+        } else {
+            created.pointee = true
+            let user = ZMUser.insertNewObject(in: context)
+            user.remoteIdentifier = remoteIdentifier
+            user.domain = domain
+            return user
+        }
+    }
+
+    // TODO jacob either we delete the duplicate merging or we update it to handle domains
```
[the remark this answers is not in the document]
typfel: So the problem duplicate merging is that with introduction of a domain it's valid for two user to exist with the same UUID given that they belong to different domains. Keeping this method as it is would potentially incorrectly delete users since it doesn't take the domain into account.

Now this method introduced at some point to workaround a bug where a database some how ended up multiple users with the same UUID which should never happen in first place.

Tests/Source/Model/ZMManagedObjectFetchingTests.swift
typfel: Calling `refresh(user, mergeChanges: false)` turns an object into fault and it will be reloaded from the disk the next time it's accessed. When we fetching an object we first look into `NSManagedObjectContext.registeredObjects` to see if we find a non-faulted object there before perform a fetch request. I'm also fairly sure that refresh(mergeChanges:false)` removes an objects from `registeredObjects`.

https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506224-refresh

Source/Model/Conversation/ZMConversation+Creation.swift
```diff
@@ -20,6 +20,34 @@ import Foundation
...
+    }
+
+    @objc public static func fetchOrCreate(with remoteIdentifier: UUID,
+                                           domain: String?,
+                                           in context: NSManagedObjectContext,
+                                           created: UnsafeMutablePointer<Bool>) -> ZMConversation {
+        // We must only ever call this on the sync context. Otherwise, there's a race condition
+        // where the UI and sync contexts could both insert the same user (same UUID) and we'd end up
+        // having two duplicates of that user, and we'd have a really hard time recovering from that.
+        require(context.zm_isSyncContext, "Users are only allowed to be created on sync context")
```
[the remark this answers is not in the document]
typfel: Updated the docs

Source/Model/User/ZMUser+Create.swift
```diff
@@ -19,6 +19,34 @@
...
+        return fetchOrCreate(with: remoteIdentifier, domain: domain, in: context, created: &created)
+    }
+
+    @objc static func fetchOrCreate(with remoteIdentifier: UUID,
+                                    domain: String?,
+                                    in context: NSManagedObjectContext,
+                                    created: UnsafeMutablePointer<Bool>) -> ZMUser {
+        // We must only ever call this on the sync context. Otherwise, there's a race condition
+        // where the UI and sync contexts could both insert the same user (same UUID) and we'd end up
+        // having two duplicates of that user, and we'd have a really hard time recovering from that.
```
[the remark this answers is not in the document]
typfel: Updated the docs

johnxnguyen (approved):

Source/Model/Connection/ZMConnection.m
```diff
@@ -260,14 +260,7 @@ + (instancetype)connectionWithUserUUID:(NSUUID *)UUID inContext:(NSManagedObject
...
         ZMConnection *connection = [ZMConnection insertNewObjectInManagedObjectContext:moc];
-        ZMUser *user = [ZMUser userWithRemoteID:UUID createIfNeeded:NO inContext:moc];
-        
-        if (user == nil) {
-            user = [ZMUser userWithRemoteID:UUID createIfNeeded:YES inContext:moc];
-            user.needsToBeUpdatedFromBackend = YES;
-        }
-        
-        connection.to = user;
+        connection.to = [ZMUser fetchOrCreateWith:UUID domain:nil in:moc];;
```
David-Henner, line 263: Do we need to set `needsToBeUpdateFromBackend` in this case ?

Source/Model/Conversation/ZMConversation+Creation.swift
```diff
@@ -20,6 +20,52 @@ import Foundation
...
+    ///     - remoteIdentifier: UUID assigned to the conversation.
+    ///     - domain: domain assigned to the conversation.
+    ///     - context: `NSManagedObjectContext` on which to fetch or create the conversation.
+    ///                NOTE that this **must** be the sync context.
+    ///     - created: Will be set `true` if a new user was created.
+
+    @objc public static func fetchOrCreate(with remoteIdentifier: UUID,
+                                           domain: String?,
+                                           in context: NSManagedObjectContext,
+                                           created: UnsafeMutablePointer<Bool>) -> ZMConversation {
```
David-Henner, line 52: I know this might be inconvenient to change since the API calls have already been updated to this new API, but if it's imperative for the sync context to be passed here, would it make sense to change the function's signature to 
```
fetchOrCreate(with remoteIdentifier: UUID,
              domain: String?,
              syncContext: NSManagedObjectContext,
              created: UnsafeMutablePointer<Bool>)
```

Source/Model/ZMManagedObject.m
```diff
@@ -346,6 +351,36 @@ + (instancetype)fetchObjectWithRemoteIdentifier:(NSUUID *)uuid inManagedObjectCo
...
 }
 
++ (instancetype)internalFetchObjectWithRemoteIdentifier:(NSUUID *)uuid
+                                         domain:(NSString *)domain
+                         inManagedObjectContext:(NSManagedObjectContext *)moc;
+{
+    // Executing a fetch request is quite expensive, because it will _always_ (1) round trip through
+    // (1) the persistent store coordinator and the SQLite engine, and (2) touch the file system.
+    // Looping through all objects in the context is way cheaper, because it does not involve (1)
+    // taking any locks, nor (2) touching the file system.
```
David-Henner, line 361: looks like the `(1)` in the second line shouldn't be there

Source/Model/Conversation/ZMConversation+Creation.swift
typfel: IMO this doesn't give that much benefit of the current safe guards (documentation and runtime assertion).

Something I'd like to do later is to introduce wrappers around the `NSManagedObjectContext` which specify the type of the context  like this: 
```
struct SyncContext {
   var underlayingContext: NSManagedObjectContext
}
```

Then we could enforce these rules via the type system.

Source/Model/Connection/ZMConnection.m
typfel: No, it was redundant since we set `needsToBeUpdatedFromBackend =  true` in `awakeFromInsert`, which means it will be set for all newly inserted/created users.

PR #1169, opened 2021-06-22
Fix: Designated initializer cannot be declared in an extension of 'Mention'

billypchan:
### Issues

When data model is build as a XCFramework and link with other frameworks, compile error occurs.

### Causes

`Designated initializer cannot be declared in an extension of 'Mention'` and similar error for `Reaction`

### Solutions

User static method instead of `init` in extensions. 

Add `typealias WireProtosMention = WireProtos.Mention` to prevent linker ambiguity issue when linking this framework.

### DIFF
--- a/Source/Utilis/Protos/GenericMessage+Helper.swift
+++ b/Source/Utilis/Protos/GenericMessage+Helper.swift
@@ -358,7 +358,7 @@ extension Text {
     public init(content: String, mentions: [Mention] = [], linkPreviews: [LinkMetadata] = [], replyingTo: ZMOTRMessage? = nil) {
         self = Text.with {
             $0.content = content
-            $0.mentions = mentions.compactMap { WireProtos.Mention($0) }
+            $0.mentions = mentions.compactMap { WireProtos.Mention.createMention($0) }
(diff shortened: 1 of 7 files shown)


KaterinaWire (approved):

Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -513,12 +513,12 @@ extension External {
 }
 
 // MARK: - Mention
-
-public extension WireProtos.Mention {
-    init?(_ mention: WireDataModel.Mention) {
+public typealias WireProtosMention = WireProtos.Mention
```
David-Henner, line 516: could we have a comment there explaining why we need this typealias?

billypchan:
```suggestion
///When other frameworks links to data modal XCFramework, the generated swift interface file omit the namespace and gives linker error. Added this typealias solves the issue.
public typealias WireProtosMention = WireProtos.Mention
```

David-Henner (approved):

billypchan:
@zenkins release major

PR #1171, opened 2021-06-28
[Feature] Support for qualified message endpoint 

typfel:
## What's new in this PR?

- Update after a namespace was added to the `otr.proto` file which added `Proteus_` to all protobuf structs used to communicate with the backend API.
- Add support for building `Proteus_QualifiedNewOtrMessage` messages  which is the protobuf message used with the new qualified message endpoint.

## Dependencies

- [x] https://github.com/wireapp/wire-ios-protos/pull/47


Source/Model/Message/ZMClientMessage+Encryption.swift
```diff
@@ -88,17 +98,35 @@ extension ZMAssetClientMessage: EncryptedPayloadGenerator {
...
     public var debugInfo: String {
         return "\(String(describing: underlyingMessage))"
     }
     
 }
 
+extension GenericMessage {
+
+
+    public func encryptForProteus(for recipients: [ZMUser: Set<UserClient>],
```
KaterinaWire, line 115: What is this method for?

KaterinaWire, line 115: What is this method for?

johnxnguyen (approved):

Source/Model/Message/ZMClientMessage+Encryption.swift
typfel: Nothing it should be deleted :)

KaterinaWire (approved):

PR #1172, opened 2021-07-01
Add ProtosReactionFactory to fix WireProtos.Reaction extension not generated correct in Swift interface

billypchan:
### Issues

When testing conversion to XCFramework, `WireProtos.Reaction.createReaction` can not be found. Similar PR: https://github.com/wireapp/wire-ios-data-model/pull/1170

### Causes

generated Swift interface can not find method `createReaction` under `WireProtos.Reaction`, seems it has ambiguity between `WireProtos.Reaction` & `WireDataModel.Reaction`.

### Solutions

Create a container enum `ProtosReactionFactory` wrapper to solve the namespace issue.

### DIFF
--- a/Source/Utilis/Protos/GenericMessage+Helper.swift
+++ b/Source/Utilis/Protos/GenericMessage+Helper.swift
@@ -410,6 +410,13 @@ extension WireProtos.Reaction {
     }
 }
 
+public enum ProtosReactionFactory {
+    public static func createReaction(emoji: String, messageID: UUID) -> WireProtos.Reaction {
+        return WireProtos.Reaction.createReaction(emoji: emoji,
+                                                  messageID: messageID)
+    }
+}
+
 // MARK: - LastRead
 
 extension LastRead {


typfel:
Do you have link describing these namespacing issues with XCFramework? It's not so cool to have create these wrappers and other changes for something which should actually not be an issue.

billypchan:
@typfel the below snapshot of sync engine project show the issue:

Screenshot 2021-07-01 at 17 37 45

and I checked the generated interface file does not have the namespace `WireProtos`. The other solution would be rename `WireDataModel.Reaction` but I would suggest this solution for minimal changes.

billypchan:
another choice would be copy and paste the `createReaction` method to SE, but it may cause issues for repeated implementation.

typfel:
> and I checked the generated interface file does not have the namespace `WireProtos`. The other solution would be rename `WireDataModel.Reaction` but I would suggest this solution for minimal changes.

How can I reproduce this issue on my setup? Do I need to run Xcode 13?

billypchan:
@typfel it only occurs when linking a XCFramework(Xcode 12.4)

typfel:
> @typfel it only occurs when linking a XCFramework(Xcode 12.4)

Ok cool, could you provide instructions for how reproduce it? I'd also like to have a look at the error.

typfel (approved):

KaterinaWire (approved):

billypchan:
@zenkins release minor

PR #1176, opened 2021-07-09
Fix: UTType init returns nil on M1 Mac

billypchan:
### Issues

When running test on M1 Mac iOS simulator, `UTType(mimeType: )` always return nil and cause `hasRasterImage` always false

### Causes

Unknown

### Solutions

As a workaround, check `mimeType` is SVG to return it is a raster image or not.

related PR:
https://github.com/wireapp/wire-ios-images/pull/43

### DIFF
--- a/Source/Model/Message/RasterImages+Protobuf.swift
+++ b/Source/Model/Message/RasterImages+Protobuf.swift
@@ -22,6 +22,7 @@ public extension WireProtos.Asset.Original {
             return false
         }
         
+        // FUTUREWORK remove once arm64 simulator support have been added JIRA ticket: SQPIT-583
         #if targetEnvironment(simulator)
         if let utType = UTType(mimeType: mimeType) {
             return utType.isSVG == false
--- a/Source/Model/Message/RasterImages+Protobuf.swift
+++ b/Source/Model/Message/RasterImages+Protobuf.swift
@@ -18,8 +18,21 @@
 
 public extension WireProtos.Asset.Original {
     var hasRasterImage: Bool {
-        guard case .image? = self.metaData,
-            UTType(mimeType: mimeType)?.isSVG == false else { return false }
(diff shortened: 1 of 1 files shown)


David-Henner (approved):

billypchan:
I checked that this issue is related to a bug of arm64 simulator. But I would keep this temporary fix til arm64 simulator frameworks are ready.

typfel:
@billypchan I don't so this issue wasn't resolved by https://github.com/wireapp/wire-ios-images/pull/43 ?

billypchan:
@typfel that PR was solved the tests failures on `wire-ios-images` and this PR fixes the new failure found on `wire-ios`.

Source/Model/Message/RasterImages+Protobuf.swift
```diff
@@ -18,8 +18,21 @@
 
 public extension WireProtos.Asset.Original {
     var hasRasterImage: Bool {
-        guard case .image? = self.metaData,
-            UTType(mimeType: mimeType)?.isSVG == false else { return false }
+        guard case .image? = self.metaData else {
+            return false
+        }
+        
+        #if targetEnvironment(simulator)
```
typfel, line 25: So the root cause issue for this problem is missing arm64 simulator support?

Then let's add a
// FUTUREWORK remove once arm64 simulator support have been added (<link to JIRA ticket>)

typfel (changes requested):

Source/Model/Message/RasterImages+Protobuf.swift
billypchan:
```suggestion
        // FUTUREWORK remove once arm64 simulator support have been added JIRA ticket: SQPIT-583
        #if targetEnvironment(simulator)
```

typfel (approved):

PR #1178, opened 2021-07-15
Fix: ZMUpdateEvent stub not work on XCFramework

billypchan:
### Issues

After using XCFramework, it is no longer allow stub `ZMUpdateEvent` `messageNonce` property. After remove that stub, some tests crashes

### Causes

`messageNonce` no longer allow  be marked @objc since it is in extension 
After using XCFramework, `ZMUpdateEvent` stub no longer works and the tests calls original code

### Solutions

~disable those tests and refactor them without stub later.~ Add simulator only guard in `updateEventParticipantsRemovedReason` to prevent accessing `ZMUpdateEvent.payload` stub


Source/Model/Message/ZMSystemMessage+ChildMessages.swift
```diff
@@ -22,7 +22,15 @@ extension ZMSystemMessage {
     
     @objc
     static func updateEventParticipantsRemovedReason(_ updateEvent: ZMUpdateEvent) -> ZMParticipantsRemovedReason {
-            return updateEvent.participantsRemovedReason
+        
+        // fix: OCMock no longer work after XCFramework is used
```
typfel, line 26: Create a ticket for removing the usage of OCMock and link it here.

typfel: Or remove the OCMock usage in this PR doesn't look like it's that much work.

billypchan: Other tests are still depends on OCMock, e.g. `testThatItSavesTheConversationTitleInConversationNameChangeSystemMessage` still calls objc code. I would prefer keep this minimal change and remove ocmock later.

billypchan: > Create a ticket for removing the usage of OCMock and link it here.

https://wearezeta.atlassian.net/browse/SQPIT-601

typfel, line 26:
```suggestion
        // FUTUREWORK: OCMock no longer work after XCFramework is used. We should remove the OCMock dependency in the tests SQPIT-601
```

typfel (approved):

David-Henner (approved):

PR #1182, opened 2021-07-22
[Feature] Add file sharing feature

KaterinaWire:
## What's new in this PR?

- added new `fileSharing` feature;

## Notes

https://wearezeta.atlassian.net/browse/SQSERVICES-611


Source/Model/User/UserType.swift
```diff
@@ -56,7 +56,10 @@ public protocol UserType: NSObjectProtocol, UserConnections {
     
     /// Whether the PDF digial signature is enable
     var hasDigitalSignatureEnabled: Bool { get }
-
+    
+    /// Whether the user can share files
+    var canShareFiles: Bool { get }
```
johnxnguyen, line 61: Two things here that I see as a little problematic:

This particular feature config is not only sharing files with others (sending) but also about other's sharing files with the self user (receiving). I think the name needs to make this clearer, because for me it seems to only be concerned about sending.

On the one hand I see that it would be convenient to make this a property requirement on `UserType`, since we could get a reference to the self user easily and check this value. But this property would only be relevant for the self user and no other users. This is not clear and will lead to confusion. In https://github.com/wireapp/wire-ios-data-model/pull/1183/ we also remove the relationship to team, which you rely on to get this value. I think we could find a better way to propagate this value to the needed call sites.

Source/Model/FeatureConfig/FeatureLike.swift
```diff
@@ -37,15 +37,15 @@ public protocol FeatureLike: Codable {
 
     /// The object used to configure the feature.
 
-    var config: Config { get }
+    var config: Config? { get }
```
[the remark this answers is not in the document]
KaterinaWire: Merged that PR

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -162,6 +163,14 @@ public class Feature: ZMManagedObject {
                    config: defaultConfigData,
                    team: team,
                    context: context)
+        case .fileSharing:
+            let defaultInstance = Feature.FileSharing()
```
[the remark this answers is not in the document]
KaterinaWire: Merged that PR

Source/Model/User/UserType.swift
KaterinaWire: Yes, probably `UserType` isn't the best place.
I decided to propagate it in SessionManager as we did it with AppLock (the solution is in PR for SE)

David-Henner (approved):

Source/Model/FeatureConfig/FileSharing/Feature.FileSharing.swift
```diff
@@ -0,0 +1,38 @@
+//
+// Wire
+// Copyright (C) 2020 Wire Swiss GmbH
```
billypchan, line 3:
```suggestion
// Copyright (C) 2021 Wire Swiss GmbH
```

billypchan (approved):

PR #1183, opened 2021-07-23
[Refactor] Feature configs

johnxnguyen:
## What's new in this PR?

This refactor is intended to simplify the transition to "team-agnostic" feature configs. Currently, feature configs fetched from the backend are always bound to a team because a team id is required in the endpoint. Backend is introducing an additional endpoint for each feature config that does not require a team id (they are able to infer it). 

Since we were binding feature configs to the team in the API, we were also reflecting this in the client code by creating a relationship between the `Feature` and `Team` Core Data entities. It was written in such a way that feature configs could be accessed by traversing the relationship from the self user team.

This refactoring... 
- removes the relationship and any associated of the team and feature completely. 
- Access to the features is provided by the new `FeatureService`, which takes over the responsibilities previously performed by the `Team` extension.
- The `FeatureLike` protocol has been removed, it was very awkward due to the associated type, which made it very hard to express a feature without a configuration (and just a status).
- Previously the default instances for a particular `Feature` were created on demand (i.e first access but no instance found). This also felt a bit awkward and I've opted to create all default instances up front when initializing the sync context.
- The app lock controller now has two config structs, a `Config` for what comes from the backend, and a `LegacyConfig` that comes from the compile time `session_manager.json` file. The latter is now optional and will override the backend config, so it should only be present for customers that actually use it.

Needs releases with:

- [ ] https://github.com/wireapp/wire-ios-build-configuration/pull/55
- [ ] https://github.com/wireapp/wire-ios-bk-configuration/pull/28
- [ ] https://github.com/wireapp/wire-ios-ey-configuration/pull/56
- [ ] https://github.com/wireapp/wire-ios-build-assets/pull/39


Resources/zmessaging.xcdatamodeld/.xccurrentversion
```diff
@@ -3,6 +3,6 @@
 <plist version="1.0">
 <dict>
 	<key>_XCCurrentVersionName</key>
-	<string>zmessaging2.92.0.xcdatamodel</string>
+	<string>zmessaging2.93.0.xcdatamodel</string>
```
KaterinaWire, line 6: What has changed in the data model? It's difficult to read the file changes.

Source/Model/FeatureConfig/FeatureService.swift
```diff
@@ -0,0 +1,102 @@
...
+    // MARK: - Life cycle
+
+    public init(context: NSManagedObjectContext) {
+        self.context = context
+    }
+
+    // MARK: - Accessors
+
+    public func fetchAppLock() -> Feature.AppLock {
+        let feature = Feature.fetch(name: .appLock, context: context)!
```
KaterinaWire, line 47: Is it safe to have a force unwrapping?

Resources/zmessaging.xcdatamodeld/.xccurrentversion
johnxnguyen: The relationship between `Team` and `Feature` has been removed.

Source/Model/FeatureConfig/FeatureService.swift
johnxnguyen: Yes because we create default instances when the `CoreDataStack` is setup, but I was going to add an assertion failure to give a better description, but I forgot. I'll add that.

Source/Model/FeatureConfig/FeatureService.swift
```diff
@@ -0,0 +1,102 @@
...
+
+    // MARK: - Life cycle
+
+    public init(context: NSManagedObjectContext) {
+        self.context = context
+    }
+
+    // MARK: - Accessors
+
+    public func fetchAppLock() -> Feature.AppLock {
```
KaterinaWire, line 46: Should we create similar public methods for all future features?

johnxnguyen: Yes

David-Henner (approved):

KaterinaWire (approved):

johnxnguyen:
@zenkins release major

PR #1186, opened 2021-07-28
Remove OCMock in ZMMessageTests

billypchan:
## What's new in this PR?

Create `protocol UpdateEvent` to prevent using `OCMock` in `ZMMessageTests`


billypchan:
continue in https://github.com/wireapp/wire-ios-data-model/pull/1188

PR #1188, opened 2021-07-28
create UpdateEvent protocol

billypchan:
## What's new in this PR?

Create `UpdateEvent` to allow mocking `ZMUpdateEvent` properties without using `OCMock`.


KaterinaWire:
@billypchan How did you decide in which methods the `ZMUpdateEvent` should be converted to an `UpdateEvent`, and where not?

billypchan:
@KaterinaWire this PR is for remove the `OCMock` for `ZMUpdateEvent` so I refactor the mocked properties only.

typfel (changes requested):
Do we really need to mock `ZMUpdateEvent`? It's fairly simple class which holds a payload. Another approach would be to create `ZMUpdateEvent` with the expected payload in the tests.

billypchan:
@typfel The mock of `ZMUpdateEvent` to replace `OCMock` stub is necessary for testing. The properties of `ZMUpdateEvent` `timestamp`, `senderUUID` and `conversationUUID` are depends on other properties (`isTransient`, `type`, `payload`).

typfel:
> @typfel The mock of `ZMUpdateEvent` to replace `OCMock` stub is necessary for testing. The properties of `ZMUpdateEvent` `timestamp`, `senderUUID` and `conversationUUID` are depends on other properties (`isTransient`, `type`, `payload`).

Yes, `timestamp` depend on `isTransient` and `payload` but that doesn't prevent you from writing tests. You need to provide a valid payload for the given scenario you want to test. All those properties can supplied in the existing initialiser and we already have existing tests which does this.

PR #1191, opened 2021-08-04
[Feature] Add isRestricted property to ZMConversationMessage protocol

KaterinaWire:
## What's new in this PR?

Add `isRestricted` property to `ZMConversationMessage` which we can use and mock in the UI project.

### DIFF
--- a/Source/Model/Message/ConversationMessage.swift
+++ b/Source/Model/Message/ConversationMessage.swift
@@ -154,6 +154,9 @@ public protocol ZMConversationMessage : NSObjectProtocol {
     var needsLinkAttachmentsUpdate: Bool { get set }
     
     var isSilenced: Bool { get }
+
+    /// Whether the message can not be received or shared.
+    var isRestricted: Bool { get }
 }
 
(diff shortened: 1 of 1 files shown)


johnxnguyen (approved):

Source/Model/Message/ConversationMessage.swift
```diff
@@ -271,6 +274,18 @@ extension ZMMessage : ZMConversationMessage {
     public var isSilenced: Bool {
         return conversation?.isMessageSilenced(nil, senderID: sender?.remoteIdentifier) ?? true
     }
+
+    public var isRestricted: Bool {
+        guard (self.isFile || self.isImage),
+            let managedObjectContext = self.managedObjectContext else {
+            return false
+        }
```
David-Henner, line 282:
```suggestion
        guard 
            (self.isFile || self.isImage),
            let managedObjectContext = self.managedObjectContext 
        else { return false }
```

David-Henner (approved):

PR #1201, opened 2021-08-11
[Feature] Parse qualified IDs when available

typfel:
## What's new in this PR?

Parse the domain when processing update events.


typfel:
replaced by https://github.com/wireapp/wire-ios-data-model/pull/1204

PR #1228, opened 2021-09-02
[Fix] Don't notify users when the default feature status has changed

KaterinaWire:
### Issues
Every time the user logs in, they are notified that the conference calling feature is now enabled.

### Causes
The first time a user logs in, we add a default value for conference calling, which is disabled. After that, we initiate to fetch this value from the BE, and if it is enabled, it means that, according to the implementation, the user should be informed about the changes.

### Solutions
Add a new property, `hasInitialDefault`, which will indicate that the current value is the default and we don't have to notify the user of the changes.


Source/Model/FeatureConfig/Feature.swift
```diff
@@ -51,14 +51,15 @@ public class Feature: ZMManagedObject {
     @NSManaged private var statusValue: String
     @NSManaged private var configData: Data?
     @NSManaged public var needsToNotifyUser: Bool
+    @NSManaged public var hasInitialDefault: Bool
```
johnxnguyen, line 54: This doesn't need to be public.

johnxnguyen (commented):
It would also be good to have a test (that no notification is sent after changing the initial default) so it is clear what the purpose of the flag is.

KaterinaWire:
> It would also be good to have a test (that no notification is sent after changing the initial default) so it is clear what the purpose of the flag is.

Added test

KaterinaWire:
@johnxnguyen I need to keep `hasInitialDefault` as public, since in Request strategy, after receiving a response, we update the status and config. And this is a moment when `hasInitialDefault` should be set to false.
https://github.com/wireapp/wire-ios-request-strategy/pull/365

johnxnguyen:
> I need to keep hasInitialDefault as public, since in Request strategy, after receiving a response, we update the status and config. And this is a moment when hasInitialDefault should be set to false.

@KaterinaWire can you set `hasInitialDefault = false` in the `updateOrCreate` method after we make changes to an existing instance?

Source/Model/FeatureConfig/Feature.swift
```diff
@@ -51,14 +51,15 @@ public class Feature: ZMManagedObject {
...
     
     public var config: Data? {
         get {
             return configData
         }
 
         set {
-            if !statusValue.isEmpty {
+            if !statusValue.isEmpty && !hasInitialDefault {
```
johnxnguyen, line 62: I think it would be helpful to have a comment here to explain why we need both of these conditions. Or, you could refactor it out to a private computed property and use the name to describe the purpose of the condition. This way you can reuse the property in the other setter.

Tests/Source/Model/FeatureConfiguration/FeatureTests.swift
```diff
@@ -130,11 +134,50 @@ final class FeatureTests: ZMBaseManagedObjectTest {
...
+                feature.status = .enabled
+            }
+        }
+
+        // then
+        XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 0.5))
+        XCTAssertTrue(testObserver.changes.isEmpty)
+    }
+
+    class TestObserver: NSObject {
```
johnxnguyen, line 169: I guess this can be private.

KaterinaWire: Done

Source/Model/FeatureConfig/Feature.swift
KaterinaWire: Good idea! 
I added :
```
private var hasBeenUpdatedFromBackend: Bool {
        return !statusValue.isEmpty && !hasInitialDefault
    }
```

KaterinaWire:
> > I need to keep hasInitialDefault as public, since in Request strategy, after receiving a response, we update the status and config. And this is a moment when hasInitialDefault should be set to false.
> 
> @KaterinaWire can you set `hasInitialDefault = false` in the `updateOrCreate` method after we make changes to an existing instance?

In the Request strategy, when we receive a response from BE, we set the status and config without using the `updateOrCreate` method.

johnxnguyen:
@KaterinaWire would it be possible to flip `hasInitialDefault = false` at the end of the setters? You'd also need to set 
`hasInitialDefault = true` at the end of the create block (currently it's before the change blocks is executed). This would keep this logic private.

KaterinaWire:
> @KaterinaWire would it be possible to flip `hasInitialDefault = false` at the end of the setters? You'd also need to set
> `hasInitialDefault = true` at the end of the create block (currently it's before the change blocks is executed). This would keep this logic private.

Yes, this approach works. I think we can only set `hasInitialDefault = false` in the status setter since every feature has a status property.

KaterinaWire:
> @KaterinaWire would it be possible to flip `hasInitialDefault = false` at the end of the setters? You'd also need to set
> `hasInitialDefault = true` at the end of the create block (currently it's before the change blocks is executed). This would keep this logic private.

Yes, this approach works.

johnxnguyen (approved):

typfel (approved):

PR #1229, opened 2021-09-03
Self deleting messages (part 1)

johnxnguyen:
## What's new in this PR?

**Jira:** https://wearezeta.atlassian.net/browse/SQSERVICES-541

### Context

This PR contains some refactoring regarding how the different possible message destruction timeouts for a conversation are exposed. Previously there were only 2 different timeouts per conversation, the "local" timeout only affects the self users, and the "synced" timeout affects all participants of the group and is synchronized through the backend. 

Both of these timeouts are stored on the `ZMConversation` entity and both have a value simultaneously. The active timeout (i.e the one that is actually used) depends on which of these values is set. In short, use the synced timeout, otherwise the local timeout, otherwise none. This logic is encapsulated within the `messageDestructionTimeout` property which is a unified accessor for both kinds of timeouts (modeled with `MessageDestructionTimeout`).

### Problem

Now there is a third kind of timeout set by the team admin which affects the whole team, and although this timeout has the highest precedence (i.e it overrides all other timeouts), it doesn't negate the possibility to configure the synced timeout for the conversation. (The idea here is that even though I may send messages with a 10 second timeout set by my team admin, as a group admin I may still set a 5 minute timeout for the group which affects other participants). 

The current implementation of the single accessor `messageDestructionTimeout` seems to be designed to expose only a single timeout for the ui layer, so it's not possible to use a team timeout and still set a group timeout.

### Solution

This PR contains changes to make this possible, by:

- replacing the `messageDestructionTimeout` accessor with an API that allows fetching and setting any of the three possible timeouts separately
- adding API for the "active" timeout, which contains the logic that determines which of the three possible timeouts to use

Also included is some general refactoring, such as renaming types, moving types into separate files, and other miscellaneous cleaning.


Source/Model/Conversation/Self deleting messages/ZMConversation+SelfDeletingMessages.swift
```diff
@@ -0,0 +1,127 @@
+//
+// Wire
+// Copyright (C) 2016 Wire Swiss GmbH
```
KaterinaWire, line 3:
```suggestion
// Copyright (C) 2021 Wire Swiss GmbH
```

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -107,6 +107,13 @@ public extension ZMUser {
     
     @objc(canModifyEphemeralSettingsInConversation:)
     func canModifyEphemeralSettings(in conversation: ConversationLike) -> Bool {
+        guard
+            let featureService = managedObjectContext.map(FeatureService.init),
+            featureService.fetchSelfDeletingMesssages().status == .enabled
+        else {
```
KaterinaWire, line 113: Should the user be allowed to modify settings if `enforcedTimeoutSeconds` > 0?

Tests/Source/Model/Conversation/ZMConversationTests+Ephemeral.swift
```diff
@@ -112,39 +103,90 @@ class ZMConversationTests_Ephemeral : BaseZMMessageTests {
...
+            featureService.storeSelfDeletingMessages(.init(status: .enabled, config: .init(enforcedTimeoutSeconds: 300)))
+        }
+
+        syncMOC.performGroupedBlockAndWait {
+            let feature = featureService.fetchSelfDeletingMesssages()
+            XCTAssertEqual(feature.status, .enabled)
+            XCTAssertEqual(feature.config.enforcedTimeoutSeconds, 300)
+        }
+
+        syncMOC.performGroupedBlockAndWait {
```
KaterinaWire, line 181: Is it necessary to put the below code in a separate block?

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
```diff
@@ -0,0 +1,188 @@
...
+            .fiveMinutes,
+            .oneHour,
+            .oneDay,
+            .oneWeek,
+            .fourWeeks
+        ]
+    }
+
+    var displayString: String? {
+        guard .none != self else { return NSLocalizedString("input.ephemeral.timeout.none", comment: "") }
```
billypchan, line 109: Is this string included in UI project? DM does not have string resource. I would recommend return a enum and let UI project handle localization.

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
```diff
@@ -0,0 +1,188 @@
...
+            return String(Int(rawValue / .oneDay))
+        }
+
+        if isWeeks {
+            return String(Int(rawValue / .oneWeek))
+        }
+
+        if isYears {
+            return String(Int(rawValue / .oneYear))
+        }
```
billypchan, line 136: I would suggest use switch to save some code and improve performance

```
    switch rawValue {
    case (.zero)..<(.oneMinute):
        return String(Int(rawValue))
    case (.oneMinute)..<(.oneHour):
        return String(Int(rawValue / .oneMinute))
:
:
    }
```

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
```diff
@@ -0,0 +1,188 @@
...
+    return (.oneWeek)..<(.oneYear) ~= rawValue
+    }
+
+    var isYears: Bool {
+        return rawValue >= .oneYear
+    }
+
+}
+
+fileprivate let longStyleFormatter: DateComponentsFormatter = {
```
billypchan, line 167: make this `static` and put as a member/`DateComponentsFormatter` of a class would be easier to manage

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
```diff
@@ -0,0 +1,188 @@
...
+extension TimeInterval {
+
+    static let tenSeconds: Self = 10
+    static let oneMinute: Self = 60
+    static let fiveMinutes: Self = 300
+    static let oneHour: Self = 3600
+    static let oneDay: Self = 86400
+    static let oneWeek: Self = 604800
+    static let fourWeeks: Self = 2419200
+    static let oneYear: Self = 31557600
```
billypchan, line 186: Do we need to consider leap year case?

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
johnxnguyen: Yes this is referring to a string in the ui project, it was already in data model I just moved it from one place to another. I tried to move all this code to the ui project but it got too messy so I would prefer to do it in a separate PR.

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
johnxnguyen: Yes, I already tried that, but we actually need the `isSeconds`, `isMinutes`, ... properties for other code, so I decided to keep it as it is for now. I can try to refactor it again separately.

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
johnxnguyen: Ah I guess that was the reason why it was previous computed after adding a year. I can revert that part of the code.

Source/Model/User/ZMUser+Permissions.swift
johnxnguyen: Yes, because this permission concerns the group timeout, so as long as ephemeral messages are enabled they can change the value of the timeout. Note though that if `enforcedTimeoutSeconds > 0`, then they will send a message with the forced timeout, regardless of what the conversation timeout it.

Tests/Source/Model/Conversation/ZMConversationTests+Ephemeral.swift
johnxnguyen: Nope! Fixed.

Source/Model/Conversation/Self deleting messages/ZMConversation+SelfDeletingMessages.swift
johnxnguyen: Fixed

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
johnxnguyen: Fixed

Source/Model/Conversation/Self deleting messages/MessageDestructionTimeoutValue.swift
johnxnguyen: Moved it to a private extension.

billypchan (approved):

KaterinaWire (approved):

PR #1237, opened 2021-09-20
[Fix] Add additional verification to message sending

KaterinaWire:
## What's new in this PR?

This fix is an extra layer of checking that the user cannot send a file (files or images) message if file sharing feature status is disable. With this fix, we can ensure that if the user finds a way to add a file from the UI, we won't send this message.


typfel (approved):

David-Henner (approved):

KaterinaWire:
@zenkins release patch

PR #1236, opened 2021-09-20
[Feature] Integrate new conversation request strategy

typfel:
## What's new in this PR?

Remove the methods for parsing transport payloads on the `ZMConversation`, that responsibility has now been taken over by the new `ConversationRequestStrategy` in `wire-ios-request-strategy`. Also add new public methods for updating the conversation state which are used by the conversation strategy.

PR #1239, opened 2021-09-28
[Feature] Add actions for adding / removing participants

typfel:
## What's new in this PR?

Introduce `EntityAction` and `EntityActionHandler` as new pattern for adding actions to entities which returns a result or failure.

This new pattern is used to re-implement support for adding / removing participants.

## Notes

 - `QualifiedID` was moved here from the wire-ios-request-strategy.
 - This PR needs to be rebased once https://github.com/wireapp/wire-ios-data-model/pull/1236 is merged

### DIFF
--- a/Source/Model/Conversation/ZMConversation+Participants.swift
+++ b/Source/Model/Conversation/ZMConversation+Participants.swift
@@ -19,8 +19,54 @@
 import Foundation
 import WireProtos
 
+public enum ConversationRemoveParticipantError: Error {
+   case unknown,
+        invalidOperation,
+        conversationNotFound
+}
+
+public enum ConversationAddParticipantsError: Error {
+   case unknown,
+        invalidOperation,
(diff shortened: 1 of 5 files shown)


Source/Model/Conversation/ZMConversation+Participants.swift
```diff
@@ -19,8 +19,54 @@
...
+   case unknown,
+        invalidOperation,
+        accessDenied,
+        notConnectedToUser,
+        conversationNotFound,
+        tooManyMembers,
+        missingLegalHoldConsent
+}
+
+public class AddParticipantAction: EntityAction {
```
David-Henner, line 38: should it be named `AddParticipantsAction` if it takes multiple users?

David-Henner (approved):

billypchan (approved):

PR #1240, opened 2021-09-29
[Fix] The external participants state in conversation

KaterinaWire:
### Issues
In a group conversation created by a member of another team, the banner with the external participants state doesn't show the presence of guests in the conversation.

### Causes
We've displayed a banner with information about guests based on their role in the current conversation. Since the conversation was created by a member of another team, `selfUser` was considered a guest and we did not display a banner.

### Solutions
Display a banner with information about guests if there are non-team member participants.


KaterinaWire:
@typfel This issue is discussed [here](https://wearezeta.atlassian.net/browse/SQSERVICES-582). So, we shouldn't show the guest icon for the selfUser if he/she is a guest in the current conversation, but not a guest in the team.

David-Henner (approved):

PR #1248, opened 2021-10-21
Fix: creating duplicate conversations/users when domain is empty string

typfel:
### Issues

It's been observed that app starts to crash while processing events.

### Causes

The app crashes due to sanity check that we don't have multiple conversations with the same ID.

The code path which causes duplicate conversation to inserted:
```
    static func updateConversation(withLastReadFromSelfConversation lastRead: LastRead, inContext moc: NSManagedObjectContext) {
        let newTimeStamp = Double(integerLiteral: lastRead.lastReadTimestamp)
        let timestamp = Date(timeIntervalSince1970: newTimeStamp/1000)
        guard let conversationID = UUID(uuidString: lastRead.conversationID) else {
            return
        }
        let conversation = ZMConversation.fetchOrCreate(with: conversationID, domain: lastRead.qualifiedConversationID.domain, in: moc)
        conversation.updateLastRead(timestamp, synchronize: false)
    }
 ```

If the `lastRead` protobuf object doesn't contain a `qualifiedConversationID` it will default to an empty string. An empty domain is treated as different from the local domain and new conversation is therefore inserted. Later if fetch is being made without a domain parameter it will find both conversations.

### Solutions

Treat empty domain as a nil domain.

### Future improvements

- Make it impossible to to fetch conversations/users without specifying a domain.

### DIFF
--- a/Source/Model/ZMManagedObject.m
+++ b/Source/Model/ZMManagedObject.m
@@ -365,7 +365,7 @@ static NSString * const KeysForCachedValuesKey = @"ZMKeysForCachedValues";
     Require(entity != nil);
 
     NSString *key = [self remoteIdentifierDataKey];
-    NSString *domainKey = [self remoteIdentifierDataKey];
+    NSString *domainKey = [self domainKey];
     NSData *data = uuid.data;
     for (NSManagedObject *mo in moc.registeredObjects) {
         if (!mo.isFault && mo.entity == entity &&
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: Source/Model/ZMManagedObject.m [hunk 1]
```diff
@@ -365,7 +365,7 @@ + (instancetype)internalFetchObjectWithRemoteIdentifier:(NSUUID *)uuid
     Require(entity != nil);
 
     NSString *key = [self remoteIdentifierDataKey];
-    NSString *domainKey = [self remoteIdentifierDataKey];
+    NSString *domainKey = [self domainKey];
```
typfel, line 368: Noticed that this was using the wrong key causing us to do fetch requests unnecessarily.

PR #1252, opened 2021-10-28
feat: Include federated users in externalParticipantsState SQCORE-1126

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCORE-1126" title="SQCORE-1126" target="_blank">BugSQCORE-1126</a>  [iOS] Distinguish "federated users are present" from guests
  </td></table>
  <br />
 

## What's new in this PR?

Introduce `visibleRemotes` to  the `ExternalParticipantsState` option set which is included when a conversation contain federated users. The `visibleRemotes` logic is the same as of `visibleGuests`, that is it's only included for users with a team in groups.


typfel:
release major

billypchan:
release major

PR #1254, opened 2021-11-03
Fix: Handle conflicting domains FS-119

typfel:
### Issues

There are instances of Wire servers which share the same database but run on different domains. This causes the iOS client to stop receiving  and sending messages in some cases.

### Causes

In this multi domain setup it's possible to login to account A1 on server S1 (domain: s1.com) and on server S2 (domain s2.com) since they both share the same database. When this user sends a message from S1 the qualified ID of the sender will be (uuid: 123, domain: s1.com) and when it's sent from S2 the qualified ID will be (uuid: 123, domain: s2.com).

The iOS client will treat these qualified IDs as coming from different user since the domain is different and thus insert a user object for each one of them. This causes problems since they are actually referring to the same user on the backend.

### Solutions

- Ignore the domain when creating users and conversations if federation is not enabled.
- Delete previously created duplicate users and conversations with a domain not equal to self user domain.

### DIFF
--- a/Patches/InvalidDomainRemoval.swift
+++ b/Patches/InvalidDomainRemoval.swift
@@ -1,0 +1,54 @@
+//
+// Wire
+// Copyright (C) 2021 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
+import Foundation
+
+enum InvalidDomainRemoval {
+
+    /// We had a situation where we were creating duplicate users and conversations where the UUID would
+    /// be duplicated but the domain would be different.
+    ///
+    /// We need to delete all duplicated users and converations which doesn't have a domain equal to the
+    /// self user domain.
+    static func removeDuplicatedEntitiesWithInvalidDomain(in moc: NSManagedObjectContext) {
+        guard let selfDomain = ZMUser.selfUser(in: moc).domain else {
+            return
+        }
+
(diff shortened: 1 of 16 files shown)


Patches/InvalidDomainRemoval.swift
```diff
@@ -0,0 +1,54 @@
...
+import Foundation
+
+enum InvalidDomainRemoval {
+
+    /// We had a situation where we were creating duplicate users and conversations where the UUID would
+    /// be duplicated but the domain would be different.
+    ///
+    /// We need to delete all duplicated users and converations which doesn't have a domain equal to the
+    /// self user domain.
+    static func removeDuplicatedEntitiesWithInvalidDomain(in moc: NSManagedObjectContext) {
```
David-Henner, line 29: should we make sure this is only called when federation is disabled? Otherwise we run the risk of deleting false duplicates. e.g: two different users with same identifier on different federated backends

Tests/Source/Model/Conversation/ZMConversationTests.m
```diff
@@ -387,6 +387,40 @@ - (void)testThatItTreatsEmptyDomainAsNil
...
 
+- (void)testThatItIgnoresDomainWhenFederationIsDisabled
+{
+    // given
+    NSUUID *uuid = [NSUUID createUUID];
+
+    [self.syncMOC performBlockAndWait:^{
+        // when
+        self.syncMOC.zm_isFederationEnabled = NO;
+        ZMConversation *found = [ZMConversation fetchOrCreateWith:uuid domain:@"a.com" in:self.syncMOC];
```
David-Henner, line 398: is the conversation meant to be created here? If so, should it be named `found` ? It gives the impression it's being fetched. If it's fetched then nothing to worry about

billypchan (approved):

Patches/InvalidDomainRemoval.swift
typfel: 👍  Yes, that would safer. 

However we don't know if federation is enabled or not at this point. We needs to refactor the code for that. I'll create follow up ticket for this. Since federation is not deployed anymore yet in production we still have time fix this later.

Tests/Source/Model/Conversation/ZMConversationTests.m
typfel: Copy & paste :)

Yes, the conversation is actually created in this case and `created` or `conversation` would be a better name.  I'll change the name.

typfel: Updated

David-Henner (approved):

PR #1259, opened 2021-11-05
Fix: App crashes in [ZMConversation refreshObjectsThatAreNotNeededInSyncContext:self]

typfel:
### Issues

App crashes in `[ZMConversation refreshObjectsThatAreNotNeededInSyncContext:self]`

### Causes

When fetching the `localParticipants` the app crashes if a `ParticipantRole` doesn't have user since it's declared as non-optional. However when deleting users there's no current rule to delete the corresponding `ParticipantRole` the user can actually be nil.

### Solutions

Declare `user` as optional.


David-Henner (approved):

billypchan (approved):

typfel:
release major

PR #1269, opened 2021-11-09
feat: check domain when checking if users are on the same team FS-35

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-35" title="FS-35" target="_blank">TaskFS-35</a>  [Clients] When checking if a user is in the same team as the user, consider the domain
  </td></table>
  <br />
 

## What's new in this PR?

https://wearezeta.atlassian.net/browse/FS-35

Consider the domain when checking if users are on the same team. 

Provided that `canAccessCompanyInformation` is used whenever we're checking if users are on the same team, checking that domains are equal ensures we won't mistake users as team members when team ids are the same but domains differ (in federation context).


Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -169,11 +169,11 @@ public extension ZMUser {
...
             let selfUserTeamID = self.team?.remoteIdentifier
-            else {
-                return false
+        else {
+            return false
         }
-        
-        return selfUserTeamID == otherUserTeamID
+
+        return selfUserTeamID == otherUserTeamID && domain == otherUser.domain
```
typfel, line 176: Since we can't be guaranteed that domain is always set for all users I think this will fail in some cases. 

For example when the `selfUser` has a domain but it's `nil` for the `otherUser`. In this case we would incorrectly return false.

A safer approach would be to change the statement to:
```
selfUserTeamID == otherUserTeamID && !otherUser.isFederated
```

typfel (changes requested):

KaterinaWire (approved):

typfel (approved):

PR #1272, opened 2021-11-12
[Fix] App crashes after update

KaterinaWire:
### Issues
There's a crash when we try to update the public version from TestFlight.

### Causes
We create default feature configs in a sync context, but fetch them in a view context. This could be the cause of the crash.

### Solutions
Create default feature configs in the view context and save the context.


billypchan (approved):

typfel (approved):

KaterinaWire:
release patch

PR #1273, opened 2021-11-12
[Chore] Add public logs

KaterinaWire:
## What's new in this PR?

This is the second attempt to fix a crash in the public app. I have reverted some of the changes that were made recently.


billypchan (approved):

typfel (commented):
Revert looks good but we still have a double performGroupedBlockAndWait

typfel (approved):

PR #1274, opened 2021-11-12
feat(conversation-mention): Mention in conversation #FS-86

sb88k:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-86" title="FS-86" target="_blank">TaskFS-86</a>  [iOS] As a user, I can use @mentions in conversations with federated users
  </td></table>
  <br />
 

## What's new in this PR?

Mention feature in federated conversation

### Attachments
Screenshot 2021-11-12 at 17 53 41

Screenshot 2021-11-12 at 17 53 41


Source/Utilis/Protos/GenericMessage+Helper.swift
```diff
@@ -599,6 +599,14 @@ public extension WireDataModel.Mention {
             $0.start = Int32(range.location)
             $0.length = Int32(range.length)
             $0.userID = userID
+            
+            if user.isFederated,
```
typfel, line 603: We want to include the domain for all users if available so I'd suggest to remove `user.isFederated`

typfel (changes requested):

typfel (approved):

David-Henner (approved):

PR #1275, opened 2021-11-15
feat: Broadcast availability to same backend users only #FS-101

sb88k:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-101" title="FS-101" target="_blank">StoryFS-101</a>  [iOS] Disable sending status across backends
  </td></table>
  <br />
 

## What's new in this PR?

Removing recipients from other backend when broadcasting availability change


Source/Model/User/ZMUser+Availability.swift
```diff
@@ -91,6 +91,9 @@ extension ZMUser {
             .prefix(remainingSlots)
 
         recipients.formUnion(teamUsers)
+        
+        guard let domain = ZMUser.selfUser(in: context).domain else { return recipients }
+        recipients = recipients.filter { $0.domain == domain }
```
typfel, line 96:
```suggestion
        recipients = recipients.filter { !$0.isFederated }
```

typfel (changes requested):

Source/Model/User/ZMUser+Availability.swift
typfel: This is shorter and handles edge cases when the domain is `nil` on the user.

typfel (approved):

David-Henner (approved):

sb88k:
release minor

sb88k:
release minor

PR #1276, opened 2021-11-16
[Chore] Add public logs

KaterinaWire:
## What's new in this PR?

To be able to reproduce the crash using TestFlight build, there are more public logs that can help us.

### Note
This should be reverted after debugging.

### DIFF
--- a/Source/ManagedObjectContext/CoreDataStack.swift
+++ b/Source/ManagedObjectContext/CoreDataStack.swift
@@ -308,7 +308,6 @@ public class CoreDataStack: NSObject, ContextProvider {
             context.mergePolicy = NSMergePolicy(merge: .mergeByPropertyObjectTrumpMergePolicyType)
 
             FeatureService(context: context).createDefaultConfigsIfNeeded()
-            context.saveOrRollback()
         }
 
         // this will be done async, not to block the UI thread, but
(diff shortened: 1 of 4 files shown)


Source/ManagedObjectContext/CoreDataStack.swift
```diff
@@ -308,7 +310,12 @@ public class CoreDataStack: NSObject, ContextProvider {
             context.mergePolicy = NSMergePolicy(merge: .mergeByPropertyObjectTrumpMergePolicyType)
 
             FeatureService(context: context).createDefaultConfigsIfNeeded()
-            context.saveOrRollback()
+
+            do {
+                try context.save()
+            } catch {
+                zmLog.safePublic("Failed to save context: \(error.localizedDescription)")
+            }
```
KaterinaWire, line 318: @johnxnguyen  It was a temporary solution to save the context here. We save it in the updateOrCreate () method. Why do you want to keep it?

johnxnguyen: ah, in that case we can remove it. As long as we have one place where we can log the context save.

johnxnguyen: Removed

johnxnguyen (approved):

typfel (approved):

KaterinaWire:
release patch

PR #1277, opened 2021-11-16
[Fix] Remove forced unwrapping

KaterinaWire:
## What's new in this PR?

Another attempt to figure out the reason of the crash is to remove forced unwrapping from the` fetchFeatureName()` methods. This will probably bring the crash to another part of the code and may help us.


johnxnguyen (approved):

typfel (approved):

KaterinaWire:
release patch

PR #1288, opened 2021-11-18
feat: consider domain when checking conversation team - FS-108

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-108" title="FS-108" target="_blank">TaskFS-108</a>  [iOS] [Clients] When checking a conversation belongs to your team, consider the domain
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

When checking if a conversation belongs to a user's team, we need to consider the domain. If the domain is different, the conversation is not a team conversation.

### Solutions

Extended `ZMConversation` with a helper method `isFederating(with: UserType)` and used that whenever we check if the conversation is a team conversation.

### Testing

Wrote new unit tests for:
- `ZMConversation+Team`
- `ZMConversation+Federation`
- `isGuest(in: ConversationLike)` method

----

----

----
##### References


Source/Model/Conversation/ZMConversation+Participants.swift
```diff
@@ -389,18 +389,14 @@ extension ZMConversation {
...
-        {
-            self.needsToDownloadRoles = true
+        guard
+            conversationType == .group,
+            !isTeamConversation
+        else { return }
+
+        // if there are no roles with actions
+        if nonTeamRoles.isEmpty || !nonTeamRoles.contains(where: { !$0.actions.isEmpty }) {
+            needsToDownloadRoles = true
```
David-Henner, line 399: this will be executed if the conversation doesn't belong to the same team or if the conversation belongs to another domain. If it shouldn't be executed when the conversation is federated, then we need to change the guard to return early in such cases.

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -184,10 +187,11 @@ public extension ZMUser {
             // We store the teamRemoteIdentifier of the team to check if we don't have a local team,
             // but received a teamId in the conversation payload, which means we are a guest in the conversation.
 
-            if conversation.creator == self {
+            // Federated users and conversation creators are never guests
+            if conversation.creator == self || conversation.isFederating(with: self) {
```
typfel, line 191: So here we are saying that self user is not a guest in a conversation if the conversation belongs do a different domain.

I think that's wrong we should consider the self user to be a guest.

Source/Model/User/ZMUser+Permissions.swift
```diff
@@ -184,10 +187,11 @@ public extension ZMUser {
...
                 return false
             }
-            
+
             if let team = team {
                 // If the self user belongs to a team he/she's a guest in every non team conversation
                 return conversation.teamRemoteIdentifier != team.remoteIdentifier
```
typfel, line 197: I think here we need a something like:
```
return conversation.teamRemoteIdentifier != team.remoteIdentifier || conversation.isFederating(with: self))
```

typfel (changes requested):

Source/Model/User/ZMUser+Permissions.swift
David-Henner: I see your point, my bad. I think I applied the "federated users are never guests" rule to the conversation, as in "users in federated conversations are never guests"

Source/Model/User/ZMUser+Permissions.swift
David-Henner: Yes, seems right.

Source/Model/Conversation/ZMConversation+Participants.swift
typfel: Yes makes sense. Federation conversation are by design never team conversations from the perspective of the self user since teams can't contain federated users.

typfel (approved):

sb88k (approved):

PR #1291, opened 2021-11-22
fix: use private class for fetching framework bundle SQSERVICES-1031

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1031" title="SQSERVICES-1031" target="_blank">BugSQSERVICES-1031</a>  [iOS] Release build crashes
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

The database patches should only we run when the app has upgraded and uses a new data model version, but we are observing that they're running all the time, even on a fresh install.

### Causes

The current version of the framework is fetched from the framework bundle, which is used to determine whether the patches should be run. For some reason, when this code is run in the testflight app, the Bundle we retrieve is actually the main app bundle, and the version that we retrieve is actually the app's short version, which is currently 3.93. This means that effectively all patches are run, because we expect a much larger data model version.

### Solutions

The bundle is fetched by passing in _ZMUser.self_  in an initializer. This should work, but for some reason it's not, and I wonder if the bundle code gets confused with this type because it is so ubiquitous and has extensions in the main app bundle. To test this, I am passing in a private class `BundleAnchor`. If this still doesn't work, then it seems like a bug with the Bundle initializer. If that's the case, then I will try fetch the bundle explicitly using the bundle identifier.

### Testing

This can only be tested by running this code in the testflight app.

----

----

----
##### References

### DIFF
--- a/Patches/PersistedDataPatches.swift
+++ b/Patches/PersistedDataPatches.swift
@@ -39,7 +39,7 @@ public struct PersistedDataPatch {
     public static func applyAll(in moc: NSManagedObjectContext, fromVersion: String? = nil, patches: [PersistedDataPatch]? = nil) {
         zmLog.safePublic("Beginning patches...")
 
-        guard let currentVersion = Bundle(for: ZMUser.self).infoDictionary!["CFBundleShortVersionString"] as? String else {
+        guard let currentVersion = Bundle(for: BundleAnchor.self).infoDictionary!["CFBundleShortVersionString"] as? String else {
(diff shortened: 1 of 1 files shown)


KaterinaWire (approved):

typfel (approved):

johnxnguyen:
release patch

PR #1299, opened 2021-12-06
refactor: Convert ZMConversation Objc to Swift #SQCORE-1158

sb88k:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQCORE-1158" title="SQCORE-1158" target="_blank">SubtaskSQCORE-1158</a>  Convert ZMConversation.m to Swift
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Convert ZMConversation entity code from Objc to Swift.


Tests/Source/Utils/RolesMigrationTests.swift
```diff
@@ -64,7 +64,7 @@ class RolesMigrationTests: DiskDatabaseTest {
         oneToOneConvo.needsToDownloadRoles = false
         oneToOneConvo.needsToBeUpdatedFromBackend = false
         
-        let selfConvo = ZMConversation.selfConversation(in: self.moc)
+        let selfConvo = ZMConversation.selfConversation(in: self.moc) 
         selfConvo.conversationType = .self
         selfConvo.addParticipantAndUpdateConversationState(user: selfUser, role: nil)
         selfConvo.userDefinedName = "Self"
```
sb88k, line 70: @typfel 
This test case is still failing due to the missing self conversation.
And I think the base class is not creating the self conversation.

I ran this with objc code and `selfConvo` here is pointing `0x00000000`, so I guess the asserts related to `selfConvo` below are not really working.
Also, two test cases in sync engine, `testThatEventsAreForwardedToAllEventConsumers_WhenProcessed` and `testThatItProcessBufferedEvents_WhenSyncCompletes`, are failing because of the same reason.

so, my assumption is that the self conversation should be created along with some object together, but I could not find any clue.

If we change `selfConversation(in:)` function to return optional ZMConversation, it will cause a lot of changes in all over the places in the consumers, also unexpected behaviors.

typfel: When you run the real application we always have a self conversation because it's created by the BE which we then synchronise.

I agree that changing  `selfConversation(in:)` to return an optional ZMConversation would cause a lot of changes.

I think we should fix the tests by explicitly creating the self conversation in the `DiskDatabaseTests` since they seems to expect it to be there.

Source/Model/Conversation/ZMConversation.swift
```diff
@@ -0,0 +1,540 @@
...
+        clearedTimeStamp = lastServerTimeStamp // This will delete all messages
+        lastReadServerTimeStamp = lastServerTimeStamp
+    }
+    
+    public func revealClearedConversation() {
+        isArchived = false
+    }
+    
+    @objc(appendMessage:)
+    public func append(_ newMessage: ZMMessage?) {
```
typfel, line 331: I agree that we should change the signatures of public methods but I think we should take second pass after this PR is merged and see how we can clean up the API.

Here I'd say it's pretty clear that `newMessage` shouldn't be optional since we crash the app if it is.

typfel (approved):
Good work! 🚀 

Also before we merge this we should create playground build and run the automation regression run.

Source/Model/Conversation/ZMConversation+Type.swift
```diff
@@ -0,0 +1,73 @@
...
+    
+    public var conversationType: ZMConversationType {
+        get {
+            guard var internalType = ZMConversationType(rawValue: internalConversationType) else { return .invalid }
+            
+            if internalType == .group,
+               teamRemoteIdentifier != nil,
+               (userDefinedName ?? "").isEmpty,
+               localParticipantRoles.count == 2,
+               localParticipantsExcludingSelf.count == 1 {
```
johnxnguyen, line 61: If I remember correctly there was a comment here describing that the order of these checks affects performance. I think it would be a good idea to preserve that comment because it's not obvious that changing the order of these calls would be more expensive.

johnxnguyen (changes requested):

Source/Model/Conversation/ZMConversation+Type.swift
sb88k: I will take a look 👍

johnxnguyen (approved):

PR #1328, opened 2022-01-24
fix: only use domain if federation is enabled FS-357

johnxnguyen:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-357" title="FS-357" target="_blank">BugFS-357</a>  [iOS] 1:1 or group calling is not working between iOS & WEB on staging (non federated)
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

Some upstream code assumes that if a user has a domain, then the domain should be used. This is the case with calling code in the sync engine, which is leading to calling issues in non federated environments.

### Causes

The calling code needs to construct AVS identifiers from the users id and possibly the domain. In the case of staging, the domain exists and so the calling code will create a fully qualified AVS identifier. However, the other clients don't do this, so calls between these clients are breaking due to mismatched identifiers.

### Solutions

Instead of assuming federation is enabled based on the existence of a domain, we should instead (and we mostly do) rely on the internal feature flag that enables / disables federation apis. But as an extra safeguard, we can also return `nil` for user domains if the feature flag is disabled.

----

----

----
##### References


David-Henner (approved):

Source/Model/User/ZMUser.swift
```diff
@@ -223,7 +223,33 @@ extension ZMUser {
...
+    /// Note: If the federation feature flag is not set (i.e `context.zm_isFederationEnabled` is `false`),
+    /// then `nil` will always be returned. This is to ensure that we only use the domain if needed since
+    /// some code assumes that if a domain exists then the app should use federated apis.
+
+    public var domain: String? {
+        get {
+            // Only consider the domain if the app needs to federate.
+            if let context = managedObjectContext, !context.zm_isFederationEnabled {
+                return nil
+            }
```
sb88k, line 237: If MOC is nil, would it be okay to return the primitive domain below instead of nil?

if so, I would go
```suggestion
            guard managedObjectContext?.zm_isFederationEnabled == true else { return nil }
```

sb88k (approved):

Source/Model/User/ZMUser.swift
johnxnguyen: If there is no context, then it would mean probably mean that the object is deleted. As far as I know we don't delete use objects, we just mark them as deleted, so it shouldn't happen. But I think I was trying to account for this edge case anyway. I have no strong opinion about it, so happy to commit your suggestion.

johnxnguyen:
release patch

zenkins:
Hello, @johnxnguyen! Merging PR for command **release patch** for PR #**1328**. Will create a **patch** release after merged.
Next version is 283.3.2.

PR #1333, opened 2022-01-31
fix: SwiftLint warnings for multiple rules

agisilaos:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

With this PR I fix warnings and errors related to the SwiftLint rules below:

- **Computed Accessors Order Violation**: Computed properties should declare first the getter and then the setter. `(computed_accessors_order)`
- **Operator Function Whitespace Violation**: Operators should be surrounded by a single whitespace when defining them. `(operator_whitespace)`
- **Returning Whitespace Violation**: Return arrow and return type should be separated by a single space or on a separate line. `(return_arrow_whitespace)`
**-Implicit Getter Violation**: Computed read-only properties should avoid using the get keyword. `(implicit_getter)`
- Vertical Parameter Alignment Violation: Function parameters should be aligned vertically if they're in multiple lines in a declaration. `(vertical_parameter_alignment)`
- **Redundant String Enum Value Violation**: String enum values can be omitted when they are equal to the enumcase name. `(redundant_string_enum_value)`
- **For Where Violation**: `where` clauses are preferred over a single `if` inside a `for`. `(for_where)`
- **Unused Optional Binding Violation**: Prefer `!= nil` over `let _ =` `(unused_optional_binding)`
- **Orphaned Doc Comment Violation**: A doc comment should be attached to a declaration. `(orphaned_doc_comment)`
- **Colon Spacing Violation**: Colons should be next to the identifier when specifying a type and next to the key in dictionary literals. `(colon)`
- **Trailing Semicolon Violation**: Lines should not have trailing semicolons. `(trailing_semicolon)`
- **Vertical Whitespace Violation**: Limit vertical whitespace to a single empty line. Currently 3. `(vertical_whitespace)`
- **Trailing Whitespace Violation**: Lines should not have trailing whitespace. `(trailing_whitespace)`
- **NSObject Prefer isEqual Violation**: NSObject subclasses should implement isEqual instead of ==.

----
##### References


sb88k (approved):

Source/Model/Label/Label.swift
```diff
@@ -109,11 +109,10 @@ public class Label: ZMManagedObject, LabelType {
             fatal("Label entity not registered in managed object model")
         }
 
-        for managedObject in context.registeredObjects {
-            if managedObject.entity == entity, !managedObject.isFault {
+        for managedObject in context.registeredObjects where managedObject.entity == entity && !managedObject.isFault {
                 guard let label = managedObject as? Label, label.kind == .favorite else { continue }
                 return label
-            }
+
```
johnxnguyen, line 115: The indentation in the for loop seems off now.

Source/Model/Message/ConversationMessage.swift
```diff
@@ -325,10 +325,8 @@ extension ZMMessage {
 
     @objc public var usersReaction: [String: [UserType]] {
         var result = [String: [ZMUser]]()
-        for reaction in reactions {
-            if reaction.users.count > 0 {
+        for reaction in reactions where reaction.users.count > 0 {
```
johnxnguyen, line 328: Also here the indentation in the loop needs fixing.

Source/Model/User/ZMUser+RichProfile.swift
```diff
@@ -30,10 +30,6 @@ public enum ZMUserKeys {
         self.value = value
     }
 
-    public static func == (lhs: UserRichProfileField, rhs: UserRichProfileField) -> Bool {
-        return lhs.type == rhs.type && lhs.value == rhs.value
-    }
-
```
johnxnguyen, line 36: What is the lint rule that caused this change?

Tests/Source/Model/Conversation/ZMConversationTests.swift
```diff
@@ -87,7 +87,7 @@ extension ZMConversationTests {
         let result = uiMOC.executeFetchRequestOrAssert(request)
 
         XCTAssertEqual(result.count, 1)
-        /// TODO
+        // TODO
```
johnxnguyen, line 90: You can just delete this, it doesn't tell us what is to be done.

Source/Model/User/ZMUser+RichProfile.swift
agisilaos: @johnxnguyen NSObject Prefer isEqual: https://realm.github.io/SwiftLint/nsobject_prefer_isequal.html

johnxnguyen: ok, the question is whether by removing this function the behaviour of `isEqual` stays the same.

agisilaos: I think it does?! cc: @sb88k I remember we had a discussion about this

```swift
public override func isEqual(_ object: Any?) -> Bool {
         guard let other = object as? UserRichProfileField else { return false }
         return self == other
```

sb88k: As SwiftLint suggests, if == operator is replaced by `isEqual`, it works as expected.

```swift
public override func isEqual(_ object: Any?) -> Bool {
     guard let other = object as? UserRichProfileField else { return false }
     return self.type == other.type && self.value == other.value
}
```

sb88k (commented):
Looks like there is some mis-replacements.

Source/Model/User/ZMUser+RichProfile.swift
agisilaos: Ahhh, I see! Good catch. Thank you, I will update it.

agisilaos: Done!

johnxnguyen (approved):

sb88k (approved):

PR #1342, opened 2022-02-10
fix: Introduce mapping for accessRole and accessRoleV2 for the separation of access controls for guests and services SQSERVICES-1052

agisilaos:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/SQSERVICES-1052" title="SQSERVICES-1052" target="_blank">StorySQSERVICES-1052</a>  [iOS] Separate access controls for guests and services
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR, introduces the mapping between the old access role and the new one. This is an essential part for the separation of access controls for guests and services. The reason why is because backend and clients should implement the same logic (send both fields, accept either, and translate everything to the new values). The mapping will help us achieve just that.

These two methods are going to be used across our frameworks.

### TODO:

- [x] Add Unit Tests

----
##### References

### DIFF
--- a/Source/Model/Conversation/ZMConversation+AccessMode.swift
+++ b/Source/Model/Conversation/ZMConversation+AccessMode.swift
@@ -80,6 +80,46 @@ public enum ConversationAccessRole: String {
     case activated = "activated"
     /// Any user can join.
     case nonActivated = "non_activated"
+    case `private` = "private"
+
+    static func fromAccessRoleV2(_ accessRoles: Set<ConversationAccessRoleV2>) -> ConversationAccessRole {
+        switch accessRoles {
+        case []:
+            return .private
+        case [.teamMember]:
+            return .team
+        case [.teamMember, .nonTeamMember]:
+            return .activated
+        case [.teamMember, .nonTeamMember, .guest]:
+            return .nonActivated
+        case [.teamMember, .nonTeamMember, .guest, .service]:
+            return .nonActivated
+        case [.teamMember, .nonTeamMember, .service]:
+            return activated
+        case [.teamMember, .guest]:
+            return.nonActivated
+        case [.teamMember, .guest, .service]:
+            return .nonActivated
+        case [.teamMember, .service]:
+            return .activated
+        case [.nonTeamMember]:
+            return .activated
+        case [.nonTeamMember, .guest]:
+            return .nonActivated
+        case [.nonTeamMember, .guest, .service]:
+            return nonActivated
+        case [.nonTeamMember, .service]:
+            return activated
+        case [.guest]:
+            return nonActivated
+        case [.guest, .service]:
+            return nonActivated
+        case [.service]:
+            return activated
+        default:
+            return .team
+        }
+    }
 }
 
 /// The issue:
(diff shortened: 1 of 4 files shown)


Source/Model/Conversation/ZMConversation+AccessMode.swift
```diff
@@ -80,6 +80,46 @@ public enum ConversationAccessRole: String {
     case activated = "activated"
     /// Any user can join.
     case nonActivated = "non_activated"
+    case `private` = "private"
```
johnxnguyen, line 83: Is there any doc comment we can write for private?

Source/Model/Conversation/ZMConversation+AccessMode.swift
```diff
@@ -80,6 +80,46 @@ public enum ConversationAccessRole: String {
...
+
+    static func fromAccessRoleV2(_ accessRoles: Set<ConversationAccessRoleV2>) -> ConversationAccessRole {
```
johnxnguyen, line 85: Will you need this function in request strategy? If so, it needs to be public.

agisilaos: Yep we will need it in RS. Will do.

Source/Model/Conversation/ZMConversation+AccessMode.swift
```diff
@@ -105,6 +145,19 @@ public enum ConversationAccessRoleV2: String {
     case guest = "guest"
     /// A service pseudo-user, aka a non-human bot.
     case service = "service"
+
+    static func fromLegacyAccessRole(_ accessRole: ConversationAccessRole) -> Set<Self> {
```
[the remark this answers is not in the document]
agisilaos: Yes

Source/Model/Conversation/ZMConversation+AccessMode.swift
agisilaos: Couldn't find something.

johnxnguyen: Does someone in the CER squad know what it's for?

agisilaos: Just got an answer from them. Will update the doc

Source/Model/Conversation/ZMConversation+AccessMode.swift
```diff
@@ -80,6 +80,47 @@ public enum ConversationAccessRole: String {
...
+            return activated
+        case [.guest]:
+            return nonActivated
+        case [.guest, .service]:
+            return nonActivated
+        case [.service]:
+            return activated
+        default:
+            return .team
+        }
```
sb88k, line 122: It looks like this can be simplified..like
if empty { private }
else if contains only team member { team } 
else if contains guest { not activated }
else { activated }

I am not sure which way has better readability, just a little suggestion 😄

johnxnguyen (approved):

sb88k (approved):

PR #1362, opened 2022-05-09
feat: Check if federation is enabled before fetching a conversation

KaterinaWire:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

Since we use fetch(with remoteIdentifier: UUID, domain: String?, in context: NSManagedObjectContext) -> ZMConversation? method in other frameworks, it makes sense to check if federation is enabled inside it.

----

----

----
##### References


David-Henner (approved):

Tests/Source/Model/ZMManagedObjectFetchingTests.swift
```diff
@@ -26,6 +27,7 @@ class ZMManagedObjectFetchingTests: DatabaseBaseTest {
         super.setUp()
         self.mocs = self.createStorageStackAndWaitForCompletion()
         XCTAssert(waitForAllGroupsToBeEmpty(withTimeout: 1))
+        APIVersion.isFederationEnabled = true
```
johnxnguyen, line 30: It would be good to reset the value back to false in the tear down method.

johnxnguyen (changes requested):
It would also be good to have the same logic when fetching `ZMUser`.

Tests/Source/Model/ZMManagedObjectFetchingTests.swift
KaterinaWire: Done

KaterinaWire:
> It would also be good to have the same logic when fetching `ZMUser`.

ZMUser.fetchOrCreate(with remoteIdentifier: UUID,...) method uses the same method:
```
@objc static func fetchOrCreate(with remoteIdentifier: UUID,
                                    domain: String?,
                                    in context: NSManagedObjectContext,
                                    created: UnsafeMutablePointer<Bool>) -> ZMUser {
        // We must only ever call this on the sync context. Otherwise, there's a race condition
        // where the UI and sync contexts could both insert the same user (same UUID) and we'd end up
        // having two duplicates of that user, and we'd have a really hard time recovering from that.
        require(context.zm_isSyncContext, "Users are only allowed to be created on sync context")

        let domain: String? = APIVersion.isFederationEnabled ? domain : nil

        if let user = fetch(with: remoteIdentifier, domain: domain, in: context) {
            return user
        } else {
            created.pointee = true
            let user = ZMUser.insertNewObject(in: context)
            user.remoteIdentifier = remoteIdentifier
            user.domain = domain?.selfOrNilIfEmpty
            return user
        }
    }
```
The changes in this PR are related to the `ZMManagedObject`  extension. Sorry, the PR description didn't explain that this is also related to `ZMUser`.

johnxnguyen (approved):

KaterinaWire:
release patch

zenkins:
Hello, @KaterinaWire! Merging PR for command **release patch** for PR #**1362**. Will create a **patch** release after merged.
Next version is 287.2.1.

PR #1363, opened 2022-05-17
chore: remove redundant push token code

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

We've changed the way we manage push tokens in the request strategy, namely we've moved away from the upstream object sync to using an action based sync. Because of this, we don't need to rely on the `UserClient` properties `pushToken` and `legacyPushToken` for the `PushTokenStrategy` to work. In fact, we don't even need to store the push token in the client, since it only affects the self client, as if can be stored in a single location such as the user defaults.

For this reason, we are removing the push token storage entirely from the user client, and stripping away the unnecessary properties and methods of `PushToken` that are no longer needed. **However we do not delete the `pushToken` property from the database, as we need it to migrate the token to user defaults.** We can remove it after some time.

----

----

----
##### References


Source/Model/UserClient/UserClient.swift
```diff
@@ -119,28 +126,14 @@ public class UserClient: ZMManagedObject, UserClientType {
...
-                primitiveLegacyPushToken = try? JSONEncoder().encode(newValue)
-                self.didChangeValue(forKey: Keys.LegacyPushToken)
-                setLocallyModifiedKeys([Keys.LegacyPushToken])
-            }
-        }
+    /// Fetches and removes the old push token from the self client.
+    ///
+    /// - returns: the legacy push token if it exists.
+
+    public func retrieveLegacyPushToken() -> PushToken? {
```
KaterinaWire, line 133: We should also have a public setter to nil

KaterinaWire (approved):

KaterinaWire:
release major

KaterinaWire:
release major

zenkins:
Hello, @KaterinaWire! Merging PR for command **release major** for PR #**1363**. Will create a **major** release after merged.
Next version is 288.0.0.

PR #1379, opened 2022-07-05
feat(mls): add helpers for bytes conversions

David-Henner:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR adds helpers for conversions between the byte buffers used by the core crypto library and Swift types used in our project.

### Testing

- [ ] Added tests for back and forth conversions of `String` and `UUID` types

### Notes 

We may want to support more types

### DIFF
--- a/Source/Utilis/Bytes.swift
+++ b/Source/Utilis/Bytes.swift
@@ -1,0 +1,55 @@
+// Wire
+// Copyright (C) 2022 Wire Swiss GmbH
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+
+public typealias Bytes = [UInt8]
+
+public protocol BytesConvertible {
+    var bytes: Bytes { get }
+    static func from(bytes: Bytes) -> Self
+}
+
+extension BytesConvertible {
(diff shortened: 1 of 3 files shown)


Tests/Source/Utils/BytesConversionTests.swift
```diff
@@ -0,0 +1,46 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+import Foundation
+@testable import WireDataModel
+
+class BytesConversionTests: XCTestCase {
```
johnxnguyen, line 22: You could improve these tests by having separate tests for each conversation direction. For instance, string > bytes, bytes > string.

Source/Utilis/Bytes.swift
```diff
@@ -0,0 +1,55 @@
...
+    public static func from(bytes: Bytes) -> Self {
+        return bytes.data.object()
+    }
+}
+
+extension Data {
+    var bytes: Bytes { Bytes(self) }
+
+    func object<T: BytesConvertible>() -> T {
+        self.withUnsafeBytes { $0.load(as: T.self) }
```
johnxnguyen, line 41: How does this work? Can it fail somehow? For instance, if I have some random Data and try to load it as an instance of a String?

johnxnguyen: What is the main use case of initializing from bytes? Do you need to convert bytes returned from CoreCrypto to other objects?

David-Henner: > What is the main use case of initializing from bytes? Do you need to convert bytes returned from CoreCrypto to other objects?

That's the only use case I have in mind

David-Henner: > Ideally we want something like `init?(bytes: Bytes)` as the requirement. But this means we'd need to be able to catch exceptions so we can handle it gracefully by failing the initializer.

I'm not sure how to catch the exceptions since none are being thrown

David-Henner: We could restrict this to be used only within the `MLSController` to interact with core crypto, that would help reduce the risks of misuse. But we still need a way to either catch exceptions or validate the bytes

PR #1381, opened 2022-07-06
chore: support development API versions FS-709

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

## Problem

New API versions start in a development phase until the changes for that version have been finalized, up which it will move to the set of stable versions. API v2 is now in development, and we'd like to start using as we work to support it, but there is currently no way we can do this.

## Cause

`APIVersion` only keeps track of the officially `supported` API versions and doesn't know about development versions.

## Solution

- Keep track of all versions, both production and development.
- Allow setting a `preferredVersion`, which not only allows us to specify a development version, but also **any** of the production versions (not just the latest).
- Calculate the `current` version by choosing first the preferred version if it exists, or else the highest production version.

#### Test Coverage

- Unit tests for setting versions
- Unit tests for calculating current version

----

----

----
##### References


Source/Utilis/APIVersion.swift
```diff
@@ -21,10 +21,15 @@ import Foundation
 
 extension APIVersion {
 
-    private enum Keys: String {
-        case apiVersion = "currentAPIVersion"
```
johnxnguyen, line 25: It shouldn't be a problem to discard this stored value because we negation the api versions frequently.

agisilaos (approved):

ghuMustafa (approved):

PR #1392, opened 2022-07-28
feat(mls): add support for decrypting messages FS-650

David-Henner:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-650" title="FS-650" target="_blank">TaskFS-650</a>  [iOS] Receive an MLS message
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

This PR introduces support for decrypting MLS messages.

### Changes

- Added `decrypt` method to `MLSController`
- Added support to init a `GenericMessage` from an update event of type `.conversationMLSMessageAdd`

### Testing

- Added unit tests for the `decrypt` method
- Added mock of `decrypt` method
- Introduced `XCTestCase+ErrorAssertion` extension to help simplify errors and throws assertions 

----

----

----
##### References

### DIFF
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -30,7 +30,7 @@ public protocol MLSControllerProtocol {
     @discardableResult
     func processWelcomeMessage(welcomeMessage: String) throws -> MLSGroupID
 
-    func decrypt(message: String, for conversation: ZMConversation) throws -> Data
+    func decrypt(message: String, for groupID: MLSGroupID) throws -> Data?
 
 }
 
@@ -139,7 +139,6 @@ public final class MLSController: MLSControllerProtocol {
             logger.error("failed to claim key packages: \(String(describing: error))")
             throw MLSGroupCreationError.failedToClaimKeyPackages
         }
-
     }
 
     @available(iOS 15, *)
@@ -340,46 +339,35 @@ public final class MLSController: MLSControllerProtocol {
 
         case failedToConvertMessageToBytes
         case failedToDecryptMessage
-        case noGroupID
 
     }
 
-    /// Decrypts an MLS message for the given conversation
+    /// Decrypts an MLS message for the given group
     ///
     /// - Parameters:
     ///   - message: a base64 encoded encrypted message
-    ///   - conversation: the conversation representing the MLS group
+    ///   - groupID: the id of the MLS group
     ///
-    /// - Returns: The data representing the decrypted message bytes
+    /// - Returns:
+    ///   The data representing the decrypted message bytes.
+    ///   May be nil if the message was a handshake message, in which case it is safe to ignore.
     ///
     /// - Throws: `MLSMessageDecryptionError` if the message could not be decrypted
 
-    public func decrypt(message: String, for conversation: ZMConversation) throws -> Data {
-
-        guard let groupID = conversation.mlsGroupID else {
-            throw MLSMessageDecryptionError.noGroupID
-        }
+    public func decrypt(message: String, for groupID: MLSGroupID) throws -> Data? {
 
         guard let messageBytes = message.base64EncodedBytes else {
             throw MLSMessageDecryptionError.failedToConvertMessageToBytes
         }
 
-        guard let decryptedData = try decrypt(bytes: messageBytes, for: groupID) else {
-            throw MLSMessageDecryptionError.failedToDecryptMessage
-        }
-
-        return decryptedData
-    }
-
-    private func decrypt(bytes: Bytes, for groupID: MLSGroupID) throws -> Data? {
         do {
             let decryptedMessageBytes = try coreCrypto.wire_decryptMessage(
                 conversationId: groupID.bytes,
-                payload: bytes
+                payload: messageBytes
             )
             return decryptedMessageBytes?.data
         } catch {
-            logger.error("failed to decrypt message: \(String(describing: error))")
+            logger.warn("failed to decrypt message: \(String(describing: error))")
             throw MLSMessageDecryptionError.failedToDecryptMessage
         }
     }
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -30,6 +30,8 @@ public protocol MLSControllerProtocol {
     @discardableResult
     func processWelcomeMessage(welcomeMessage: String) throws -> MLSGroupID
 
+    func decrypt(message: String, for conversation: ZMConversation) throws -> Data
+
 }
 
 public final class MLSController: MLSControllerProtocol {
@@ -107,7 +109,6 @@ public final class MLSController: MLSControllerProtocol {
 
     @available(iOS 15, *)
     public func createGroup(for groupID: MLSGroupID, with users: [MLSUser]) async throws {
-        guard let context = context else { return }
 
         guard !users.isEmpty else {
             throw MLSGroupCreationError.noParticipantsToAdd
@@ -272,7 +273,7 @@ public final class MLSController: MLSControllerProtocol {
 
     private func generateKeyPackages(amountRequested: UInt32) throws -> [String] {
 
-        var keyPackages = [[UInt8]]()
+        var keyPackages = [Bytes]()
 
         do {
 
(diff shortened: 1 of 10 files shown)


Tests/MLS/MLSControllerTests.swift
```diff
@@ -43,9 +49,73 @@ class MLSControllerTests: ZMConversationTestsBase {
...
+            }
+        }
+    }
+
+    func test_Decrypt_ThrowsFailedToDecryptMessage() {
+        syncMOC.performAndWait {
+            // Given
+            let message = Data([1, 2, 3]).base64EncodedString()
+            self.conversation.mlsGroupID = self.groupID
+            self.mockCoreCrypto.mockDecryptMessage = .some(.none)
```
David-Henner, line 86: By setting the mock value to `.some(.none)` we rely on the `decrypt` method throwing an error after get a `nil` return value from the core crypto mock. This means we're not testing the reaction of the `decrypt` method to a throw from core crypto. 

I think this is due to the way that `MockCoreCrypto` throws errors when mocks are not set. 
```swift
return try XCTUnwrap(mockDecryptMessage, "return value not mocked")
```

When we set the mock to `.none` and try running the test, we get this error

Screenshot 2022-07-28 at 09 30 27

Below, we observe that an error is caught and `.failedToDecryptMessage` is thrown

Screenshot 2022-07-28 at 09 31 39

We also observe than the new error thrown is equal to the expected error. (no screenshot, but I saw it)

### Conclusion
This seems to be due to `XCTUnwrap` failing, and thus causing the test to fail, even though our assertions are successful. 
We could rewrite the `MockCoreCrypto` class to unwrap and throw a custom error instead of using `XCTUnwrap`

```swift
guard let mock = mockDecryptMessage else { throw MockError.unmockedMethodCalled }
return mock
```

MLS/MLSController.swift
```diff
@@ -335,6 +334,56 @@ public final class MLSController: MLSControllerProtocol {
...
+    ///
+    /// - Parameters:
+    ///   - message: a base64 encoded encrypted message
+    ///   - conversation: the conversation representing the MLS group
+    ///
+    /// - Returns: The data representing the decrypted message bytes
+    ///
+    /// - Throws: `MLSMessageDecryptionError` if the message could not be decrypted
+
+    public func decrypt(message: String, for conversation: ZMConversation) throws -> Data {
```
johnxnguyen, line 357: We started to pass in the group id as an argument rather than the whole conversation, since that is typically the only part of the conversation we need. You don't have to do it but it might be nice to have a consistent approach.

MLS/MLSController.swift
```diff
@@ -335,6 +334,56 @@ public final class MLSController: MLSControllerProtocol {
...
+
+    private func decrypt(bytes: Bytes, for groupID: MLSGroupID) throws -> Data? {
+        do {
+            let decryptedMessageBytes = try coreCrypto.wire_decryptMessage(
+                conversationId: groupID.bytes,
+                payload: bytes
+            )
+            return decryptedMessageBytes?.data
+        } catch {
+            logger.error("failed to decrypt message: \(String(describing: error))")
```
johnxnguyen, line 382: FYI logging errors can cause your tests to fail. We've been using `warn` instead.

Tests/MLS/MLSControllerTests.swift
johnxnguyen: `XCTUnwrap` is actually useful in this case as it shows that the test failed due to an incorrect test environment (no mock value provided).

If you want to simulate Core Crypto throwing an error when the decrypt method is called, you could modify the mock class to accept a mock error. For example, here's a mock for creating a conversation:

```swift
var mockCreateConversationError: CryptoError?

func wire_createConversation(conversationId: ConversationId, config: ConversationConfiguration) throws {
    calls.createConversation.append((conversationId, config))

    if let error = mockCreateConversationError {
        throw error
    }
}
```

The `CryptoError`  is from CoreCrypto, so you can just pick whatever would be most suitable. For instance, this is the test that uses this error:

```swift
func test_CreateGroup_ThrowsError() throws {
        // Given
        let groupID = MLSGroupID(Data([1, 2, 3]))
        mockCoreCrypto.mockCreateConversationError = CryptoError.MalformedIdentifier(message: "bad id")

        // When
        XCTAssertThrowsError(try sut.createGroup(for: groupID)) { error in
            switch error {
            case MLSController.MLSGroupCreationError.failedToCreateGroup:
                break

            default:
                XCTFail("Unexpected error: \(String(describing: error))")
            }
        }

        // Then
        let createConversationCalls = mockCoreCrypto.calls.createConversation
        XCTAssertEqual(createConversationCalls.count, 1)
        XCTAssertEqual(createConversationCalls[0].0, groupID.bytes)
        XCTAssertEqual(createConversationCalls[0].1, ConversationConfiguration(ciphersuite: .mls128Dhkemx25519Aes128gcmSha256Ed25519))
    }
```

David-Henner: Did you have to make `CryptoError` conform to the `Error` protocol in order for this to work?

MLS/MLSController.swift
```diff
@@ -335,6 +334,56 @@ public final class MLSController: MLSControllerProtocol {
...
+            throw MLSMessageDecryptionError.noGroupID
+        }
+
+        guard let messageBytes = message.base64EncodedBytes else {
+            throw MLSMessageDecryptionError.failedToConvertMessageToBytes
+        }
+
+        guard let decryptedData = try decrypt(bytes: messageBytes, for: groupID) else {
+            throw MLSMessageDecryptionError.failedToDecryptMessage
+        }
```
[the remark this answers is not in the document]
David-Henner: It's a good question, I don't know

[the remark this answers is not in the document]
David-Henner: Right, so I checked the use case and that's what it says. I'll update accordingly. Nice catch :)

Tests/MLS/MLSControllerTests.swift
johnxnguyen: yes

johnxnguyen (approved):

MLS/MLSController.swift
```diff
@@ -107,7 +109,6 @@ public final class MLSController: MLSControllerProtocol {
 
     @available(iOS 15, *)
```
ghuMustafa, line 110: we might removed this check as update to iOS 13

Tests/MLS/MLSControllerTests.swift
```diff
@@ -46,6 +47,77 @@ class MLSControllerTests: ZMConversationTestsBase {
...
+
+            let decryptMessageCalls = self.mockCoreCrypto.calls.decryptMessage
+            XCTAssertEqual(decryptMessageCalls.first?.0, self.groupID.bytes)
+            XCTAssertEqual(decryptMessageCalls.first?.1, messageBytes)
+        }
+    }
+
     // MARK: - Create group
 
     @available(iOS 15, *)
```
ghuMustafa, line 123: Here as well

MLS/MLSController.swift
```diff
@@ -138,7 +139,6 @@ public final class MLSController: MLSControllerProtocol {
             logger.error("failed to claim key packages: \(String(describing: error))")
             throw MLSGroupCreationError.failedToClaimKeyPackages
         }
-
     }
 
     @available(iOS 15, *)
```
ghuMustafa, line 144: here

Tests/Source/Helper/XCTestCase+ErrorAssertion.swift
```diff
@@ -0,0 +1,50 @@
...
+
+import Foundation
+import XCTest
+
+extension XCTestCase {
+    typealias AsyncThrowingBlock = () async throws -> Void
+    typealias ThrowingBlock = () throws -> Void
+    typealias EquatableError = Error & Equatable
+
+    @available(iOS 13, *)
```
ghuMustafa, line 27: same here

ghuMustafa (approved):

PR #1394, opened 2022-08-02
feat: mls remove user from mls group FS-546

ghuMustafa:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-546" title="FS-546" target="_blank">TaskFS-546</a>  [iOS] Admin removes a user from an MLS group
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- Added `removeMembersFromConversation` method in `MLSMLSController`.
- Added the call of in `ZMConversation+Participants` and handled the case for proteus and mls.

### Testing

- Added tests for `removeMembersFromConversation` in `MLSMLSController`.
----

----

----
##### References

### DIFF
--- a/MLS/MLSClientID.swift
+++ b/MLS/MLSClientID.swift
@@ -20,7 +20,7 @@ import Foundation
 
 /// An ID representing a identifying a single user client.
 
-struct MLSClientID: Equatable {
+public struct MLSClientID: Equatable {
 
     // MARK: - Properties
 
@@ -59,6 +59,11 @@ struct MLSClientID: Equatable {
         self.clientID = clientID.lowercased()
         self.domain = domain.lowercased()
         self.string = "\(self.userID):\(self.clientID)@\(self.domain)"
+    }
+
+    
+    public var bytes: Bytes? {
+        string.base64EncodedString?.base64EncodedBytes
     }
 
 }
--- a/MLS/MLSClientID.swift
+++ b/MLS/MLSClientID.swift
@@ -61,9 +61,4 @@ public struct MLSClientID: Equatable {
         self.string = "\(self.userID):\(self.clientID)@\(self.domain)"
     }
 
-    
-    public var bytes: Bytes? {
-        string.base64EncodedString?.base64EncodedBytes
-    }
-
 }
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -33,7 +33,7 @@ public protocol MLSControllerProtocol {
 
     func addMembersToConversation(with users: [MLSUser], for groupID: MLSGroupID) async throws
 
-    func removeMembersFromConversation(with clientIds: [ClientId], for groupID: MLSGroupID) async throws
+    func removeMembersFromConversation(with clientIds: [MLSClientID], for groupID: MLSGroupID) async throws
 }
 
 public final class MLSController: MLSControllerProtocol {
@@ -239,13 +239,15 @@ public final class MLSController: MLSControllerProtocol {
     }
 
     public func removeMembersFromConversation(
-        with clientIds: [ClientId],
+        with clientIds: [MLSClientID],
         for groupID: MLSGroupID
     ) async throws {
 
         guard !clientIds.isEmpty else {
             throw MLSRemoveParticipantsError.noClientsToRemove
         }
+        
+        let clientIds =  clientIds.compactMap { $0.bytes }
 
         let messageToSend = try removeMembers(id: groupID, clientIds: clientIds)
 
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -33,6 +33,7 @@ public protocol MLSControllerProtocol {
 
     func addMembersToConversation(with users: [MLSUser], for groupID: MLSGroupID) async throws
 
+    func removeMembersFromConversation(with clientIds: [ClientId], for groupID: MLSGroupID) async throws
 }
 
 public final class MLSController: MLSControllerProtocol {
@@ -225,6 +226,47 @@ public final class MLSController: MLSControllerProtocol {
         } catch let error {
             logger.warn("failed to add members: \(String(describing: error))")
             throw MLSGroupCreationError.failedToAddMembers
+        }
+    }
+
+    // MARK: - Remove participants from mls group
+
+    enum MLSRemoveParticipantsError: Error {
+
+        case failedToRemoveMembers
+        case noClientsToRemove
+
+    }
+
+    public func removeMembersFromConversation(
+        with clientIds: [ClientId],
+        for groupID: MLSGroupID
+    ) async throws {
+
+        guard !clientIds.isEmpty else {
+            throw MLSRemoveParticipantsError.noClientsToRemove
+        }
+
+        let messageToSend = try removeMembers(id: groupID, clientIds: clientIds)
+
+        guard let messageToSend = messageToSend else { return }
+        try await sendMessage(messageToSend)
+
+    }
+
+    private func removeMembers(
+        id: MLSGroupID,
+        clientIds: [ClientId]
+    ) throws -> Bytes? {
+
+        do {
+            return try coreCrypto.wire_removeClientsFromConversation(
+                conversationId: id.bytes,
+                clients: clientIds
+            )
+        } catch let error {
+            logger.error("failed to remove members: \(String(describing: error))")
+            throw MLSRemoveParticipantsError.failedToRemoveMembers
         }
     }
 
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -247,7 +247,7 @@ public final class MLSController: MLSControllerProtocol {
             throw MLSRemoveParticipantsError.noClientsToRemove
         }
         
-        let clientIds =  clientIds.compactMap { $0.bytes }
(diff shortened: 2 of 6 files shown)


Source/Model/Conversation/ZMConversation+Participants.swift
```diff
@@ -200,9 +201,46 @@ extension ZMConversation {
...
+        case .proteus:
+            var action = RemoveParticipantAction(user: user, conversation: self)
+            action.onResult(resultHandler: completion)
+            action.send(in: context.notificationContext)
+
+        case .mls:
+            guard
+                let mlsController = context.mlsController,
+                let groupID = mlsGroupID?.base64EncodedString,
+                let mlsGroupID = MLSGroupID(base64Encoded: groupID)
```
johnxnguyen, line 215: `mlsGroupID` should already be of type `MLSGroupID`. So you only need to unwrap it from the conversation:

`guard let groupID = mlsGroupID else { ... )`

MLS/MLSController.swift
```diff
@@ -33,6 +33,7 @@ public protocol MLSControllerProtocol {
 
     func addMembersToConversation(with users: [MLSUser], for groupID: MLSGroupID) async throws
 
+    func removeMembersFromConversation(with clientIds: [ClientId], for groupID: MLSGroupID) async throws
```
johnxnguyen, line 36: Instead of passing in `ClientId` which is just an alias for bytes, we could pass in the more descriptive `[MLSClientID]`, then extract the bytes. This is in line with the responsibility of the `MLSController` to be a high level interface to the low level `CoreCrypto`.

Source/Model/Conversation/ZMConversation+Participants.swift
```diff
@@ -200,9 +201,46 @@ extension ZMConversation {
...
+            Task {
+                do {
+                    try await mlsController.removeMembersFromConversation(with: mlsClientIDs, for: mlsGroupID)
+
+                    context.perform {
+                        completion(.success(()))
+                    }
+
+                } catch {
+                    Logging.eventProcessing.error("Failed to remove member to mls group: \(String(describing: error))")
```
johnxnguyen, line 235: Better to `warn` than `error` to avoid test issues.

Tests/MLS/MLSControllerTests.swift
```diff
@@ -359,4 +359,98 @@ class MLSControllerTests: ZMConversationTestsBase {
...
+
+        // Mock sending message.
+        mockActionsProvider.sendMessageMocks.append({ message in
+            XCTAssertEqual(message, Data([0, 0, 0, 0]))
+
+            // Mock update event for member join
+            let payload: NSDictionary = [
+                "type": "team.member-join",
+                "data": message
+            ]
```
johnxnguyen, line 383: `team.member-join` isn't right. It should be `conversation.member-leave`.

Tests/MLS/MLSControllerTests.swift
```diff
@@ -359,4 +359,98 @@ class MLSControllerTests: ZMConversationTestsBase {
...
+
+            let updateEvent = ZMUpdateEvent(fromEventStreamPayload: payload, uuid: nil)!
+
+            self.mockConversationEventProcessor.processConversationEvents([updateEvent])
+
+            let processConversationEventsCalls = self.mockConversationEventProcessor.calls.processConversationEvents
+            XCTAssertEqual(processConversationEventsCalls.count, 1)
```
johnxnguyen, line 390: You shouldn't be calling the mock method directly in your test, it's something that will be called in the `MLSController` when we send the message. You do, however, want to assert that it was called afterwards in `THEN`.

Tests/MLS/MLSControllerTests.swift
```diff
@@ -359,4 +359,98 @@ class MLSControllerTests: ZMConversationTestsBase {
...
+            XCTFail("Unexpected error: \(String(describing: error))")
+        }
+
+        let removeMembersFromConversationCalls = mockCoreCrypto.calls.removeClientsFromConversation
+        XCTAssertEqual(removeMembersFromConversationCalls.count, 1)
+        XCTAssertEqual(removeMembersFromConversationCalls[0].0, mlsGroupID.bytes)
+        XCTAssertEqual(removeMembersFromConversationCalls[0].1, [mlsClientID])
+    }
+
+    func test_AddingMembersToConversation_ThrowsNoClientsToRemove() async {
```
johnxnguyen, line 409: The test name says "adding", but the test is for "removing.

Tests/MLS/MLSControllerTests.swift
```diff
@@ -359,4 +359,98 @@ class MLSControllerTests: ZMConversationTestsBase {
...
+            case MLSController.MLSRemoveParticipantsError.noClientsToRemove:
+                break
+
+            default:
+                XCTFail("Unexpected error: \(String(describing: error))")
+            }
+        }
+    }
+
+    func test_AddingMembersToConversation_FailsToSendHandShakeMessage() async {
```
johnxnguyen, line 429: The test name says "adding", but the test is for "removing.

MLS/MLSController.swift
ghuMustafa: Totally agree with it, as here we can avoid our high level interface dependancy to low level, Thanks.

Source/Model/Conversation/ZMConversation+Participants.swift
ghuMustafa: Yap I have changed it.

Source/Model/Conversation/ZMConversation+Participants.swift
ghuMustafa: Changed.

Tests/MLS/MLSControllerTests.swift
ghuMustafa: Yes it should be conversation.member-leave it occurred while copy past from add participants method, Thanks for identifying.

Tests/MLS/MLSControllerTests.swift
ghuMustafa: Yes, actually it should not be called, I have removed the call and make an assert after the removeMembersFromConversation is called.

Tests/MLS/MLSControllerTests.swift
ghuMustafa: updated.

Tests/MLS/MLSControllerTests.swift
ghuMustafa: updated.

MLS/MLSClientID.swift
```diff
@@ -61,4 +61,9 @@ struct MLSClientID: Equatable {
         self.string = "\(self.userID):\(self.clientID)@\(self.domain)"
     }
 
+    
+    public var bytes: Bytes? {
+        string.base64EncodedString?.base64EncodedBytes
+    }
```
johnxnguyen, line 67: Is this necessary? Maybe we can just convert to bytes where we need it?

MLS/MLSController.swift
```diff
@@ -228,6 +229,49 @@ public final class MLSController: MLSControllerProtocol {
...
+        for groupID: MLSGroupID
+    ) async throws {
+
+        guard !clientIds.isEmpty else {
+            throw MLSRemoveParticipantsError.noClientsToRemove
+        }
+        
+        let clientIds =  clientIds.compactMap { $0.bytes }
+
+        let messageToSend = try removeMembers(id: groupID, clientIds: clientIds)
```
johnxnguyen, line 252: What exactly do we need to pass in as client ids?  Raw bytes of the "user:client@domain" id? Or base64 encoded bytes?

Source/Utilis/String+Bytes.swift
```diff
@@ -24,4 +24,8 @@ extension String {
         return Bytes(base64Encoded: self)
     }
 
+    public var base64EncodedString: String? {
+        data(using: .utf8)?.base64EncodedString()
+    }
```
johnxnguyen, line 29: Why is this needed?

MLS/MLSController.swift
ghuMustafa: The Core Crypto expects an array of Raw bytes and ClientId represents `typealias ClientId = [UInt8]` and Here we make raw bytes of `MLSClientID` from String, and to make raw bytes I'm using this method `data(using: .utf8)?.base64EncodedString()`. is it the correct understanding?

MLS/MLSClientID.swift
ghuMustafa: when I tried to create bytes with following init method with plan string it return nil as it expects an base64Encoded string.

>    init?(base64Encoded: String) {
        guard let bytes = Data(base64Encoded: base64Encoded)?.bytes else {
            return nil
        }
        self = bytes
    }

ghuMustafa: Yes I got it as we just needed to pass raw data so we no need to convert to base64string.

Source/Utilis/String+Bytes.swift
ghuMustafa: I just removed the it as we need to pass the Raw bytes instead of base64 encoded bytes.

johnxnguyen (approved):

PR #1399, opened 2022-08-11
feat(mls): replenish key packages FS-793

ghuMustafa:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-793" title="FS-793" target="_blank">TaskFS-793</a>  [iOS] Replenish key packages
  </td></table>
  <br />
 

----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

- In this ticket when we call method upload key packages we check if the amount of the remaining key packages is less then 50 % we try to replenish remaining amount of key packages.
- Modified the upload key packages calls with async await.

### Testing

----

----

----
##### References


MLS/MLSController.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
 
 public protocol MLSControllerProtocol {
 
-    func uploadKeyPackagesIfNeeded()
+    func uploadKeyPackagesIfNeeded(for clientID: String, with context: NotificationContext) async throws
```
johnxnguyen, line 23: Where is this being called?

johnxnguyen: Also I don't think you need to pass in the context here, you can take it from the `context.notificationContext` in the mls controller.

MLS/MLSController.swift
```diff
@@ -580,3 +598,35 @@ public protocol ConversationEventProcessorProtocol {
...
+            if let difference = Calendar.current.dateComponents([.hour], from: storedDate, to: Date()).hour,
+               difference > 24 {
+                return true
+
+            } else {
+                return false
+            }
+        }
+    }
+}
```
johnxnguyen, line 632: I think it could be confusing to have a getter with additional logic as it could happen you set the the value to one value and then get a different value.

I suggest split the logic across two properties. One stores a `Date` timestamp for the last check, another computes if more than 24 hours has passed since the last check.

MLS/MLSController.swift
ghuMustafa: It's bing called from sync engine `UserClientRequestStrategy` after the MLS Public Keys uploaded,
https://github.com/wireapp/wire-ios-sync-engine/pull/1775/files

When I try access context from MLS Controller its crashes while getting clientID 
`let clientID = ZMUser.selfUser(in: context).selfClient()?.remoteIdentifier`
because of calling from Task { } thread.

MLS/MLSController.swift
```diff
@@ -284,43 +285,42 @@ public final class MLSController: MLSControllerProtocol {
...
-        // TODO: Here goes the logic to determine how check to remaining key packages and re filling the new key packages after calculating number of welcome messages it receives by the client.
+    public func uploadKeyPackagesIfNeeded(
+        for clientID: String,
+        with context: NotificationContext
+    ) async throws {
 
-        /// For now temporarily we generate and upload at most 100 new key packages
+        // TODO: Get actual key packages count from CoreCrypto once updated to latest. For now using a mock value of 50.
+        let clientValidKeyPackagesCount  = 50
+        let remainingKeyPackagesCountCheck = clientValidKeyPackagesCount < targetUnclaimedKeyPackageCount / 2
```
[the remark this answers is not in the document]
ghuMustafa: Thanks I will rename them

MLS/MLSController.swift
```diff
@@ -284,43 +285,42 @@ public final class MLSController: MLSControllerProtocol {
...
-
-        // TODO: Here goes the logic to determine how check to remaining key packages and re filling the new key packages after calculating number of welcome messages it receives by the client.
+    public func uploadKeyPackagesIfNeeded(
+        for clientID: String,
+        with context: NotificationContext
+    ) async throws {
 
-        /// For now temporarily we generate and upload at most 100 new key packages
+        // TODO: Get actual key packages count from CoreCrypto once updated to latest. For now using a mock value of 50.
+        let clientValidKeyPackagesCount  = 50
```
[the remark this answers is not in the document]
ghuMustafa: Yes thats more descriptive.

David-Henner (approved):

MLS/MLSController.swift
```diff
@@ -607,26 +609,22 @@ private extension UserDefaults {
         static let keyPackageQueriedTime = "keyPackageQueriedTime"
     }
 
-    var has24HoursPassedSinceLastKeyPackage: Bool {
-
-        set {
-            set(Date(), forKey: Keys.keyPackageQueriedTime)
-        }
+    var lastKeyPackageStoredDate: Date? {
```
johnxnguyen, line 612: Perhaps something like `lastKeyPackageCountDate`

MLS/MLSController.swift
johnxnguyen: In addition to uploading initial key packages, we should also try to check if we need to do this periodically. At the moment, we only do it when creating a new client, which only happens once. (maybe we we're also doing it when the mls controller initializes?).

Perhaps we can call this method after processing a welcome message, since a welcome message usually means that a key package was consumed. Note that we shouldn't actually upload new key packages after every welcome message received, but we just want to check if we need to upload more (which is what the method is doing).

johnxnguyen (changes requested):

MLS/MLSController.swift
johnxnguyen: > When I try access context from MLS Controller its crashes while getting clientID
let clientID = ZMUser.selfUser(in: context).selfClient()?.remoteIdentifier
because of calling from Task { } thread.

I was referring to the `NotificationContext`. I don't think you need to pass that as an argument, you should be able to get a reference to the notification context from the `context` inside the mls controller. So I suggest to remove it as an argument.

ghuMustafa: Yes I have update the arguments for uploadKeyPackages removed the `context` and `clientID` dependancy and also handling the async calls inside of `Task` in the method implementation.

Tests/MLS/MockMLSController.swift
```diff
@@ -47,8 +47,8 @@ class MockMLSController: MLSControllerProtocol {
 
     // MARK: - Methods
 
-    func uploadKeyPackagesIfNeeded(for clientID: String, with context: NotificationContext) async throws {
-        calls.uploadKeyPackagesIfNeeded.append((clientID, context))
+    func uploadKeyPackagesIfNeeded() {
+
```
johnxnguyen, line 52: don't forget to add 

```swift
calls.uploadKeyPackagesIfNeeded.append(())
```

MLS/MLSController.swift
```diff
@@ -287,42 +291,48 @@ public final class MLSController: MLSControllerProtocol {
...
-                )
+            userDefaults.lastKeyPackageCountDate = Date()
 
-            } catch {
-                self.logger.error("failed to generate new key packages: \(String(describing: error))")
-            }
+            guard unclaimedKeyPackageCount <= targetUnclaimedKeyPackageCount / 2 else { return }
+
+            let amount = UInt32(targetUnclaimedKeyPackageCount - unclaimedKeyPackageCount)
+            let keyPackages = try generateKeyPackages(amountRequested: amount)
```
johnxnguyen, line 318: According to the updated use case, we should always request the target number of key packages.

MLS/MLSController.swift
```diff
@@ -324,44 +329,77 @@ public final class MLSController: MLSControllerProtocol {
...
+            guard lastCheckWasMoreThan24Hours || shouldCountRemainingKeyPackages else {
+                logger.info("last check was recent and there are enough unclaimed key packages. not uploading.")
+                return false
             }
+
+            return true
+
+        } catch let error {
+            logger.warn("failed to get valid key packages count with error: \(String(describing: error))")
+            return false // maybe return true if we want to query the backend even though core crypto threw an error
```
johnxnguyen, line 382: It probably is a good idea to return true here, to be on the safer side.

johnxnguyen (commented):
Looks good. I think we need a test that we check to upload key packages after processing a welcome message. What do you think?

David-Henner:
> Looks good. I think we need a test that we check to upload key packages after processing a welcome message. What do you think?

Yes, good point. I added a simple test to verify this. I also noticed there are no tests for processing the welcome message, so they will need to be added at some point

MLS/MLSController.swift
David-Henner: done

MLS/MLSController.swift
```diff
@@ -324,44 +329,77 @@ public final class MLSController: MLSControllerProtocol {
...
+        }
 
-         countUnclaimedKeyPackages(clientID: clientID, context: context.notificationContext) { unclaimedKeyPackageCount in
-             self.logger.info("there are \(unclaimedKeyPackageCount) uncliamed key packages")
+        guard let clientID = ZMUser.selfUser(in: context).selfClient()?.remoteIdentifier else {
+            return logWarn(abortedWithReason: "failed to get client ID")
+        }
 
-             guard unclaimedKeyPackageCount <= self.targetUnclaimedKeyPackageCount / 2 else { return }
+        Task { () -> Void in
```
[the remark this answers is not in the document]
David-Henner: We don't

```swift
// doing this made the Task return value ambiguous for the compiler
return Logging.mls.warn("...")
```

PR #1405, opened 2022-09-09
feat: commit pending proposals FS-703

typfel:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-703" title="FS-703" target="_blank">TaskFS-703</a>  [iOS] Automatically Commit PendingProposals
  </td></table>
  <br />
 

----

- The **PR Title**
  - [ ] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [ ] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
----

# What's new in this PR?

Commit incoming pending proposals as described here: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/601522340/Use+Case+Committing+Pending+Valid+Proposals+MLS

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

#### How to Test

_Briefly describe how this change was tested and if applicable the exact steps taken to verify that it works as expected._

### Attachments (Optional)

_Attachments like images, videos, etc. (drag and drop in the text box)_

----

----

----
##### References

### DIFF
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -559,7 +559,7 @@ public final class MLSController: MLSControllerProtocol {
         let unmutableGroupsWithPendingCommits = groupsWithPendingCommits!
         Task {
             for (groupID, timestamp) in unmutableGroupsWithPendingCommits {
-                if timestamp.compare(Date()) != .orderedDescending {
+                if timestamp.isInThePast {
                     logger.info("commit scheduled in the past, committing...")
                     try await commitPendingProposals(in: groupID)
                 } else {
@@ -665,6 +665,13 @@ public struct MLSGroup: Equatable, Hashable {
 }
 
 // MARK: - Helper Extensions
+
+private extension Date {
+
+    var isInThePast: Bool {
+        return compare(Date()) != .orderedDescending
+    }
+}
 
 private extension String {
 
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -529,7 +529,6 @@ public final class MLSController: MLSControllerProtocol {
     /// - Parameters:
     ///   - message: a base64 encoded encrypted message
     ///   - groupID: the id of the MLS group
-    ///   - timestamp: timestamp of the event which delivered the message
     ///
     /// - Returns:
     ///   The data representing the decrypted message bytes.
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -582,6 +582,10 @@ public final class MLSController: MLSControllerProtocol {
     }
 
     func commitPendingProposals(in groupID: MLSGroupID) async throws {
+        guard let context = context else {
+            return
+        }
+
         logger.info("committing pending proposals in: \(groupID)")
 
         let commitBundle: CommitBundle?
@@ -602,8 +606,8 @@ public final class MLSController: MLSControllerProtocol {
             }
         }
 
-        context?.performAndWait {
-            let conversation = ZMConversation.fetch(with: groupID, in: context!)
+        context.performAndWait {
+            let conversation = ZMConversation.fetch(with: groupID, in: context)
             conversation?.commitPendingProposalDate = nil
         }
     }
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -593,8 +593,8 @@ public final class MLSController: MLSControllerProtocol {
         }
 
         context?.performAndWait {
-            let converation = ZMConversation.fetch(with: groupID, in: context!)
-            converation?.commitPendingProposalTimestamp = nil
+            let conversation = ZMConversation.fetch(with: groupID, in: context!)
+            conversation?.commitPendingProposalTimestamp = nil
         }
     }
 
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -66,6 +66,12 @@ class DummyCoreCryptoCallbacks: CoreCryptoCallbacks {
     }
 }
 
+public protocol MLSControllerDelegate: AnyObject {
+
+    func mlsControllerDidCommitPendingProposal(groupID: MLSGroupID)
+
+}
+
 public final class MLSController: MLSControllerProtocol {
 
     // MARK: - Properties
@@ -77,9 +83,12 @@ public final class MLSController: MLSControllerProtocol {
     private var groupsPendingJoin = Set<MLSGroup>()
 
     var backendPublicKeys = BackendMLSPublicKeys()
+    var pendingProposalCommitTimers = [MLSGroupID: Timer]()
 
     let targetUnclaimedKeyPackageCount = 100
     let actionsProvider: MLSActionsProviderProtocol
+
+    weak var delegate: MLSControllerDelegate?
 
     // MARK: - Life cycle
 
@@ -609,9 +618,13 @@ public final class MLSController: MLSControllerProtocol {
                 logger.info("commit scheduled in the past, committing...")
                 try await commitPendingProposals(in: groupID)
             } else {
-                logger.info("commit scheduled in the future, waiting...")
-                try await Task.sleep(nanoseconds: timestamp.timeIntervalSinceNow.nanoseconds)
-                try await commitPendingProposals(in: groupID)
+                // Wrap in another task so we don't block the loop.
+                Task {
+                    logger.info("commit scheduled in the future, waiting...")
+                    try await Task.sleep(nanoseconds: timestamp.timeIntervalSinceNow.nanoseconds)
+                    logger.info("scheduled commit is ready, committing...")
+                    try await commitPendingProposals(in: groupID)
+                }
             }
         }
     }
@@ -648,10 +661,6 @@ public final class MLSController: MLSControllerProtocol {
 
         logger.info("committing pending proposals in: \(groupID)")
 
-        defer {
-            clearPendingPropsalCommitDate(for: groupID)
-        }
-
         do {
             // TODO wire_commitPendingProposals will return an optional CommitBundle soon in the case
             // when there are no pending proposals, then we could return early on an error.
@@ -662,10 +671,15 @@ public final class MLSController: MLSControllerProtocol {
                 try await sendWelcomeMessage(welcome)
             }
 
+            clearPendingPropsalCommitDate(for: groupID)
+
         } catch {
             logger.info("failed to commit pending proposals in \(groupID): \(String(describing: error))")
+            clearPendingPropsalCommitDate(for: groupID)
             throw MLSCommitPendingProposalsError.failedToCommitPendingProposals
         }
+
+        delegate?.mlsControllerDidCommitPendingProposal(groupID: groupID)
     }
 
     private func clearPendingPropsalCommitDate(for groupID: MLSGroupID) {
--- a/MLS/MLSController.swift
+++ b/MLS/MLSController.swift
@@ -545,7 +545,7 @@ public final class MLSController: MLSControllerProtocol {
             groupsWithPendingCommits = conversations.compactMap { conversation in
                 if
                     let groupID = conversation.mlsGroupID,
-                    let timestamp = conversation.commitPendingProposalTimestamp
+                    let timestamp = conversation.commitPendingProposalDate
                 {
                     return (groupID, timestamp)
                 } else {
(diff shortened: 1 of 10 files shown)


MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
                 conversationId: groupID.bytes,
                 payload: messageBytes
             )
+
+            if let commitDelay = decryptedMessage.commitDelay {
```
typfel, line 521: If move this logic to the EventDecoder we don't have to pass the `timestamp` argument.

MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
...
+
+        let unmutableGroupsWithPendingCommits = groupsWithPendingCommits!
+        Task {
+            for (groupID, timestamp) in unmutableGroupsWithPendingCommits {
+                if timestamp.compare(Date()) != .orderedDescending {
+                    logger.info("commit scheduled in the past, committing...")
+                    try await commitPendingProposals(in: groupID)
+                } else {
+                    logger.info("commit scheduled in the future, waiting...")
+                    try await Task.sleep(nanoseconds: UInt64(timestamp.timeIntervalSinceNow * 1_000 * 1_000 * 1_000))
```
typfel, line 567: Is this a sane thing do to?

MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
...
+                }
+            }
+        }
+
+        logger.info("\(groupsWithPendingCommits.count) groups with scheduled pending proposals")
+
+        let unmutableGroupsWithPendingCommits = groupsWithPendingCommits!
+        Task {
+            for (groupID, timestamp) in unmutableGroupsWithPendingCommits {
+                if timestamp.compare(Date()) != .orderedDescending {
```
David-Henner, line 562: we could make this more readable by extending `Date` to do something like 

```
if timestamp.isInThePast {
...
}
```

MLS/MLSController.swift
David-Henner: this could also be more readable using an extension of `Task` such as

```swift
extension Task  {
    static func sleep(seconds: Double) async throws {
        let duration = UInt64(seconds * 1_000_000_000)
        try await Task.sleep(nanoseconds: duration)
    }
}
```

MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
...
+                    logger.info("commit scheduled in the past, committing...")
+                    try await commitPendingProposals(in: groupID)
+                } else {
+                    logger.info("commit scheduled in the future, waiting...")
+                    try await Task.sleep(nanoseconds: UInt64(timestamp.timeIntervalSinceNow * 1_000 * 1_000 * 1_000))
+                    try? await commitPendingProposals(in: groupID)
```
David-Henner, line 568: why `try?`

MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
...
+            try await sendMessage(commitBundle.commit, groupID: groupID)
+
+            if let welcome = commitBundle.welcome {
+                try await sendWelcomeMessage(welcome)
+            }
+        }
+
+        context?.performAndWait {
+            let converation = ZMConversation.fetch(with: groupID, in: context!)
+            converation?.commitPendingProposalTimestamp = nil
```
David-Henner, line 597:
```suggestion
            let conversation = ZMConversation.fetch(with: groupID, in: context!)
            conversation?.commitPendingProposalTimestamp = nil
```

Source/Model/Conversation/ZMConversation+MLS.swift
```diff
@@ -31,6 +31,9 @@ extension ZMConversation {
     @objc
     static let mlsStatusKey = "mlsStatus"
 
+    @objc
+    static let commitPendingProposalTimestamp = "commitPendingProposalTimestamp"
```
David-Henner, line 35:
```suggestion
    static let commitPendingProposalTimestampKey = "commitPendingProposalTimestamp"
```

MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
...
+    // MARK: - Pending proposals
+
+    public func commitPendingProposals() throws {
+        guard let context = context else {
+            return
+        }
+
+        logger.info("committing any scheduled pending proposals")
+
+        var groupsWithPendingCommits: [(MLSGroupID, Date)]!
```
johnxnguyen, line 542: If you initialize it to an empty array you don't need any force unwrapping:

```swift
var groupsWithPendingCommits = [(MLSGroupID, Date)]()
```

MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
...
             return decryptedMessage.message?.data
         } catch {
             logger.warn("failed to decrypt message for group (\(groupID)): \(String(describing: error))")
             throw MLSMessageDecryptionError.failedToDecryptMessage
         }
     }
 
+    // MARK: - Pending proposals
+
+    public func commitPendingProposals() throws {
```
johnxnguyen, line 535: From where/when are you planning to call this?

MLS/MLSController.swift
johnxnguyen, line 597: Perhaps we should unwrap the context as a first step to make sure we can finish the flow in its entirety.

MLS/MLSController.swift
johnxnguyen: - How many seconds could this sleep for?
- If the Task is interrupted (e.g app crash or termination), then we'll pick up the pending proposals again eventually, so that's good.
- What are your concerns?
- The [docs](https://developer.apple.com/documentation/swift/task/sleep(nanoseconds:)) say "This function doesn’t block the underlying thread.", which suggests that we don't risk any deadlocks.
- An alternative would to use a `Timer` but then you'll need to maintain it.
- Do we need to cancel the task in case someone else commits the proposal first? Or will we try and let Core Crypto inform us of a failure?

Source/Model/Conversation/ZMConversation+MLS.swift
johnxnguyen: I prefer the term "Date" rather than "Timestamp" here:
1. It's shorter
2. The type is `Date`
3. Date is much more natural language rather than timestamp which feels more technical.

But that's just my opinion, feel free to ignore ☺️

MLS/MLSController.swift
typfel: Mistake, I'll change it

MLS/MLSController.swift
```diff
@@ -492,13 +517,98 @@ public final class MLSController: MLSControllerProtocol {
...
+                    return (groupID, timestamp)
+                } else {
+                    return nil
+                }
+            }
+        }
+
+        logger.info("\(groupsWithPendingCommits.count) groups with scheduled pending proposals")
+
+        let unmutableGroupsWithPendingCommits = groupsWithPendingCommits!
```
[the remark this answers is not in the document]
typfel: I'm not allowed to reference the mutable `groupsWithPendingCommits` from the Task

MLS/MLSController.swift
typfel: > How many seconds could this sleep for?

The algorithm in use today (log(position) * 120) - 106 we assume no more than 100_000 members in a group will max out around ~ 500 seconds = 8 minutes.

> What are your concerns?
Nothing is particular.

> An alternative would to use a Timer but then you'll need to maintain it.

Using timers with async methods is ugly.

> Do we need to cancel the task in case someone else commits the proposal first? Or will we try and let Core Crypto inform us of a failure?

Two things can happen:
1. proposal is already committed but we haven't caught up yet. We will get an error from BE and abort.
2. proposal is already committed and we caught up. Core crypto will tell us there's nothing to commit and we do nothing except clearing the `commitPendingProposalTimestamp` flag.

MLS/MLSController.swift
typfel: It should be called:
- from the UserSession when the quick sync completes
- from the EventDecoder if we received the proposal via the websocket

MLS/MLSController.swift
johnxnguyen: Seems reasonable, but I'm wondering what happens in that theoretical maximum 8 minutes. I'm guessing that the suspended task can survive the app going into the background, but even if it couldn't we'd eventually pass the commit date and send it immediately.

MLS/MLSController.swift
johnxnguyen: @David-Henner is you make the `seconds` type `TimeInterval` then you can use static helpers like `.oneMinute` in place of a numeric value. But it might get this anyway because I believe `TimeInterval` is an alias for `Double`.

MLS/MLSController.swift
```diff
@@ -492,13 +524,105 @@ public final class MLSController: MLSControllerProtocol {
...
+                    return (groupID, timestamp)
+                } else {
+                    return nil
+                }
+            }
+        }
+
+        logger.info("\(groupsWithPendingCommits.count) groups with scheduled pending proposals")
+
+        let unmutableGroupsWithPendingCommits = groupsWithPendingCommits
```
johnxnguyen, line 569: You could factor this out into a method `groupsWithPendingCommits`, then use it directly in the Task (or even outside) with the need of `unmutableGroupsWithPendingCommits`.

MLS/MLSController.swift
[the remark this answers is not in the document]
johnxnguyen: You could factor out the code to fetch the groups with pending commits into a separate method. You could then use that directly in the task, avoiding the need for `unmutableGroupsWithPendingCommits` because the method would return an immutable array.

MLS/MLSController.swift
```diff
@@ -562,13 +602,142 @@ public final class MLSController: MLSControllerProtocol {
...
+        logger.info("committing any scheduled pending proposals")
+
+        let groupsWithPendingCommits = self.groupsWithPendingCommits()
+
+        logger.info("\(groupsWithPendingCommits.count) groups with scheduled pending proposals")
+
+        for (groupID, timestamp) in groupsWithPendingCommits {
+            if timestamp.isInThePast {
+                logger.info("commit scheduled in the past, committing...")
+                try await commitPendingProposals(in: groupID)
```
David-Henner, line 664: we could also wrap this in a Task and not block the loop

MLS/MLSController.swift
```diff
@@ -562,13 +602,142 @@ public final class MLSController: MLSControllerProtocol {
...
+        } catch {
+            logger.info("failed to commit pending proposals in \(groupID): \(String(describing: error))")
+            clearPendingPropsalCommitDate(for: groupID)
+            throw MLSCommitPendingProposalsError.failedToCommitPendingProposals
+        }
+
+        delegate?.mlsControllerDidCommitPendingProposal(groupID: groupID)
+    }
+
+    private func clearPendingPropsalCommitDate(for groupID: MLSGroupID) {
```
David-Henner, line 730:
```suggestion
    private func clearPendingProposalCommitDate(for groupID: MLSGroupID) {
```

Tests/Source/Helper/ZMBaseTest+Async.swift
```diff
@@ -0,0 +1,40 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see http://www.gnu.org/licenses/.
+//
+
+
```
David-Henner, line 19:
```suggestion

```

MLS/MLSController.swift
johnxnguyen: Waiting for it is less severe as waiting for future commits, but it's true that we may just want to fire and forget. But I'm not sure if it's a good idea to potentially flood core crypto with calls from different threads. I'd prefer to keep it as it is for now, then optimize in the future if we need to.

MLS/MLSController.swift
johnxnguyen: fixed

Tests/Source/Helper/ZMBaseTest+Async.swift
johnxnguyen: fixed

David-Henner (approved):

PR #1418, opened 2022-10-04
feat(mls): make commits atomic

johnxnguyen:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

### Issues

It's possible to interleave MLS commits, which results in error. For instance, we may try to commit pending proposals in to 2 different groups. Each action involves generating the commit via CoreCrypto, sending it to the backend, and merging the commit when the backend responds with a success. If we try to generate the second commit before we've merged the first commit, CoreCrypto will throw an error.

### Solutions

Make the flow of `generate commit -> send commit -> merge commit` atomic. From a concurrency point of view, this flow is a critical section where we need to ensure mutual exclusion. To do this, we make use a Swift's `actor` concurrency type to ensure mutual exclusion for specific methods and shared resources. The shared resource in this case would be Core Crypto, the methods are any action that trigger this flow of commits.

`MLSActionExecutor` is an actor that has a reference to Core Crypto and exposes one method for each action that involves commits: adding members, removing clients, updating key material, committing pending proposals.

When each of these methods are called, if the actor is not under contention (no thread is accessing it), then the calling thread will gain exclusive access, and only release it when the method returns (i.e after the commit is merged). If another thread is trying to access the actor that is under contention, then it will wait until it is released, thus not generating a new commit while another needs to be merged.

#### Test Coverage 

- Updated `MLSControllerTests`
- TODO: actor tests

----

----

----
##### References

### DIFF
--- a/MLS/MLSActionExecutor.swift
+++ b/MLS/MLSActionExecutor.swift
@@ -18,7 +18,16 @@
 
 import Foundation
 
-actor MLSActionExecutor {
+protocol MLSActionExecutorProtocol {
+
+    func addMembers(_ invitees: [Invitee], to groupID: MLSGroupID) async throws -> [ZMUpdateEvent]
+    func removeClients(_ clients: [ClientId], from groupID: MLSGroupID) async throws -> [ZMUpdateEvent]
+    func updateKeyMaterial(for groupID: MLSGroupID) async throws -> [ZMUpdateEvent]
+    func commitPendingProposals(in groupID: MLSGroupID) async throws -> [ZMUpdateEvent]
+
+}
+
+actor MLSActionExecutor: MLSActionExecutorProtocol {
 
     // MARK: - Types
 
--- a/MLS/MLSActionExecutor.swift
+++ b/MLS/MLSActionExecutor.swift
@@ -136,7 +136,8 @@ actor MLSActionExecutor {
             return events
 
         } catch MLSActionExecutorError.failedToSendCommit {
-            try clearPendingCommit(in: groupID)
+            // TODO: [John] implement proper error handling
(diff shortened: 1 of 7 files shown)


MLS/MLSActionExecutor.swift
```diff
@@ -0,0 +1,199 @@
...
+        self.context = context
+        self.actionsProvider = actionsProvider
+    }
+
+    // MARK: - Actions
+
+    func addMembers(_ invitees: [Invitee], to groupID: MLSGroupID) async throws -> [ZMUpdateEvent] {
+        let bundle = try commitBundle(for: .addMembers(invitees), in: groupID)
+        return try await sendCommitBundle(bundle, for: groupID)
+    }
```
johnxnguyen, line 77: This (and the other methods below) are critical sections. It's guaranteed that they can only be invoked one at a time.

David-Henner (commented):
Looks good! I'll wait for the tests to be completed to approve

Tests/MLS/MLSActionExecutorTests.swift
```diff
@@ -120,8 +129,199 @@ class MLSActionExecutorTests: ZMBaseManagedObjectTest {
...
+
     // MARK: - Update key material
 
+    func test_UpdateKeyMaterial() async throws {
+        // Given
+        let groupID = MLSGroupID(.random())
+
+        let mockCommit = Bytes.random()
+        let mockWelcome = Bytes.random()
+        let mockUpdateEvent = mockMemberJoinUpdateEvent()
```
ghuMustafa, line 209: Not sure here if we have to mock update events specifically for update key materials.

David-Henner (approved):

PR #1437, opened 2023-01-31
fix: Requests loop (possible solution)

KaterinaWire:
----

- The **PR Title**

- The **PR Description**
----

# What's new in this PR?

## Issue
It's possible to cause a request loop when sending a message.

## Cause
If the app is in the background and you receive a message from a user's new client, then the notification service extension will fetch the event and show the push notification for the new message. Since this is a new client, the NSE will create a new instance of `UserClient`, relate it to the sender `ZMUser`, and store it in the database.

When going to the main app and sending a message, the app will encrypt the message for every client of every participant in the conversation. For some reason, the user that added the new client doesn't have this `UserClient` in their `clients` property. As a result we won't encrypt the message for that new client. When posting the message, the backend will respond saying we are missing this new client (412 response). 

We then process this response by trying to fetch and establish an encryption session with this new client, then re-encrypt the message and send it again. In this flow, we are able to fetch the client from the database because it was created in the NSE earlier, but we don't establish a session because one is already established (in the NSE). So we re-encrypt the message but then get the 412 response from the backend. This is the loop.

But the real problem is not around the missing client (it exists and we have a session with it), but that we're not encrypting for this client when sending the message because it's not in the `user.clients` property. The only explanation is that the instance of the user is out of date: it was in memory (in the main app), then the client was created and related (in the NSE), but those changes aren't automatically pushed to the main app context. 

## Solution
If detect that `user.clients` doesn't contain a client we're looking for, but this client exists in the database, then refresh the `user`.

----

----

----
##### References


Source/Model/UserClient/UserClient.swift
```diff
@@ -240,6 +240,10 @@ public class UserClient: ZMManagedObject, UserClientType {
         }
 
         if let client = self.fetchExistingUserClient(with: remoteIdentifier, in: context) {
+            /// If the user data is out of date (for example, the user's client list)
+            /// due to changes in another process,
+            /// we update it.
```
johnxnguyen, line 245:
```suggestion
            // We already checked the user's client list but didn't find the client.
            // But we did find the client in the database, so the user is out of date
            // and needs to be refreshed.
```

johnxnguyen (approved):

agisilaos (approved):

netbe (approved):

KaterinaWire:
release patch

KaterinaWire:
release patch

zenkins:
Hello, @KaterinaWire! Merging PR for command **release patch** for PR #**1437**. Will create a **patch** release after merged.
Next version is 296.0.0.

